PR #172, opened 2026-04-10
feat(CC-0058): Add policy validation Job before Deployment rollout

berendt:
# GitHub Issue #159: Add policy validation Job before Deployment rollout

**Category**: feature | **Scope**: Medium

**Description**: When `spec.policyOverrides` is set, add a pre-deployment validation step that runs `oslopolicy-checker` (or `oslopolicy-validator`) against the rendered `policy.yaml` inside the Keystone container image before the Deployment is updated. Implement this as a short-lived Kubernetes Job that mounts the immutable ConfigMap, executes the policy check, and gates the Deployment rollout on successful completion. Add a new `PolicyValidReady` condition to track the validation result.

The validation Job should:
1. Use the same Keystone image as the API Deployment
2. Mount the rendered ConfigMap (containing `policy.yaml`) read-only
3. Run `oslopolicy-checker --config-file /etc/keystone/keystone.conf --policy /etc/keystone/policy.yaml` (or equivalent)
4. Succeed (exit 0) if all policy rules are syntactically valid and reference known Keystone actions
5. Fail with a descriptive error if any rule is invalid (unknown target, malformed check string, etc.)

The Deployment sub-reconciler must not proceed until `PolicyValidReady=True`. When `spec.policyOverrides` is nil, skip validation entirely and set the condition to True with reason `NotRequired`.

**Motivation**: Currently, policy overrides are rendered into the ConfigMap and mounted directly into the Deployment without any pre-flight check. An invalid policy rule (e.g., a typo in an action name or a malformed check string) is only discovered at runtime when Keystone rejects API requests with HTTP 500 or unexpected 403 errors. This can cause a service outage that is difficult to diagnose. Validating policies before rollout catches configuration errors early, before they reach running pods.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_policyvalidation.go` (new file)
- `operators/keystone/internal/controller/keystone_controller.go` (insert validation step between `reconcileConfig` and `reconcileDeployment`)
- `operators/keystone/api/v1alpha1/keystone_types.go` (new `PolicyValidReady` condition constant)
- `internal/common/policy/policy.go` (optional: helper for building the validator Job spec)

**Acceptance Criteria**:
- [ ] Validation Job runs only when `spec.policyOverrides` is set
- [ ] Validation Job uses the same image, SecurityContext, and config mounts as the API Deployment
- [ ] `PolicyValidReady` condition is False while the Job is running, True on success, False with reason on failure
- [ ] Deployment reconciliation is gated on `PolicyValidReady=True`
- [ ] When `spec.policyOverrides` is removed, the condition is set to True (reason `NotRequired`) and any existing validation Job is cleaned up
- [ ] Invalid policy rules produce a clear error message in the condition (not just "Job failed")
- [ ] Job has `backoffLimit: 2` and `ttlSecondsAfterFinished: 300`
- [ ] Unit tests cover: valid policy passes, invalid policy fails with descriptive reason, no-policy skips validation
- [ ] Integration test verifies the gating behavior (Deployment not updated until validation succeeds)

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/159


berendt:



## WARNING (6)

### W-001: Stale Documentation: Requeue interval says 30s but code uses 15s
**File**: `docs/reference/keystone-reconciler.md:156` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Cross-reference documentation examples against implementation

**Problem**: The documentation states "Requeue: 30s" in the ASCII diagram (line 156) and in the requeue behavior table (line 840), but the actual code uses RequeueValidationWait = 15 * time.Second in requeue_intervals.go:33.

**Code**:
```
│  └────────┬────────────────────────┘  Requeue: 30s                           │
```

**Action Required**: Change both "30s" occurrences to "15s" in docs/reference/keystone-reconciler.md at lines 156 and 840.

**Suggested Fix**:
```
Change "30s" to "15s" at line 156 in the ASCII diagram and at line 840 in the requeue behavior table.
```

---

### W-002: Missing E2E Tests: No Chainsaw test for policy validation gating
**File**: `tests/e2e/keystone/` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Missing E2E Tests

**Problem**: The project has Chainsaw E2E tests for every existing feature (17 directories under tests/e2e/keystone/). The existing policy-overrides/ E2E test only verifies ConfigMap content rendering (CC-0016) — it does not test the new validation Job gating behavior introduced in CC-0058. No Chainsaw E2E test validates the actual oslopolicy-checker execution and gating in a real cluster.

**Code**:
```
tests/e2e/keystone/policy-overrides/
```

**Action Required**: Add a Chainsaw E2E test under tests/e2e/keystone/ that validates the policy validation Job gating behavior end-to-end.

**Suggested Fix**:
```
Create a new Chainsaw E2E test directory (e.g. tests/e2e/keystone/policy-validation-gating/) that exercises the oslopolicy-checker Job creation, success/failure paths, and deployment gating in a real cluster.
```

---

### W-003: Job deletion without propagation policy may leave orphan Pods
**File**: `reconcile_policyvalidation.go:38-41` — **Fix**: ASK — **Confidence**: uncertain — **Pattern**: Helm Hooks and Lifecycle

**Problem**: When deleting a Job with a bare Delete() call, the default propagation policy is Background in controller-runtime, which cleans up Pods asynchronously. If the reconciler re-creates the Job in the same cycle, old Pods may briefly coexist with new ones.

**Code**:
```
existing := &batchv1.Job{}
existing.SetName(jobName)
existing.SetNamespace(keystone.Namespace)
if err := client.IgnoreNotFound(r.Delete(ctx, existing)); err != nil {
```

**Action Required**: Consider passing a PropagationPolicy of Foreground to the Delete call to ensure Pods are cleaned up before the Job object is removed, or document that Background propagation is intentional.

**Suggested Fix**:
```
Evaluate whether adding client.PropagationPolicy(metav1.DeletePropagationForeground) to the Delete call is warranted, or add a comment documenting that Background propagation is acceptable here because this delete path only fires when policyOverrides is removed entirely.
```

---

### W-004: backoffLimit: 2 diverges from sibling Jobs that use backoffLimit: 4
**File**: `reconcile_policyvalidation.go:145` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Enforce consistent patterns across sibling functions

**Problem**: Both buildDBSyncJob (reconcile_database.go:164) and buildBootstrapJob (reconcile_bootstrap.go:70) use backoffLimit: 4. This PR uses backoffLimit: 2, which is intentional per the PR description but the reasoning (deterministic syntax check, more retries won't help) is not documented in code or docs.

**Code**:
```
backoffLimit := int32(2)
```

**Action Required**: Add a brief comment explaining why backoffLimit: 2 is appropriate here, or align with the sibling value of 4 if there is no strong reason to diverge.

**Suggested Fix**:
```
Document the intentional deviation with a short comment or in the reconciler documentation explaining that policy validation is a deterministic check where additional retries are unlikely to succeed.
```

---

### W-005: driveFullReconciliationWithPolicy is dead code — defined but never called
**File**: `integration_test.go:1048` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Test fixtures must mirror real repository structure

**Problem**: The helper function driveFullReconciliationWithPolicy is defined at line 1048 but never called by any test. TestIntegration_PolicyValidation_GatesDeployment manually drives each phase instead of using this helper.

**Code**:
```
func driveFullReconciliationWithPolicy(t testing.TB, ctx context.Context, c client.Client, ksName, ns string) {
```

**Action Required**: Either use driveFullReconciliationWithPolicy in at least one test, or remove the dead code.

**Suggested Fix**:
```
Remove the unused driveFullReconciliationWithPolicy function from integration_test.go.
```

---

### W-006: Repeated Pod list call on each reconcile cycle for failing CRs
**File**: `reconcile_policyvalidation.go:56-68` — **Fix**: ASK — **Confidence**: likely — **Pattern**: Verify state mutations are not overwritten by fall-through

**Problem**: When a Job fails, extractJobFailureMessage lists Pods to extract failure messages. On each subsequent reconcile, job.RunJob returns ErrJobFailed again, triggering the same Pod list call. At scale with many failing CRs, this adds unnecessary API server load.

**Code**:
```
if err != nil {
    msg := fmt.Sprintf("Policy validation job failed: %v", err)
    if errors.Is(err, job.ErrJobFailed) {
        msg = extractJobFailureMessage(ctx, r.Client, jobName, keystone.Namespace)
    }
    conditions.SetCondition(&keystone.Status.Conditions, metav1.Condition{
        Type:               "PolicyValidReady",
        ...
    })
    return ctrl.Result{}, fmt.Errorf("running policy validation: %w", err)
}
```

**Action Required**: Consider caching the failure message in the condition message on first extraction, and on subsequent reconciles check if the condition already contains the failure details before re-listing Pods.

**Suggested Fix**:
```
Before calling extractJobFailureMessage, check if the existing PolicyValidReady condition already has reason PolicyValidationFailed with a non-empty message. If so, reuse the existing message instead of re-listing Pods.
```

---

## INFO (2)

### I-001: Condition type string "PolicyValidReady" used as inline string, not a constant
**File**: `reconcile_policyvalidation.go:44` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Extract repeated condition strings into typed constants

**Problem**: The string "PolicyValidReady" appears 4 times in reconcile_policyvalidation.go (lines 44, 62, 73, 83) and 14 times across test files. The sibling reconcile_networkpolicy.go already extracts its condition type into a constant (conditionTypeNetworkPolicyReady at line 30), but other siblings use inline strings.

**Code**:
```
"PolicyValidReady"
```

**Action Required**: Extract "PolicyValidReady" into a package-level constant like conditionTypePolicyValidReady and use it in the reconciler and tests.

**Suggested Fix**:
```
Add a constant at the top of reconcile_policyvalidation.go:

const conditionTypePolicyValidReady = "PolicyValidReady"

Then replace all occurrences of the inline string with the constant.
```

---

### I-002: oslopolicy-checker --namespace hardcodes "keystone"
**File**: `reconcile_policyvalidation.go:163-166` — **Fix**: ASK — **Confidence**: verified

**Problem**: The --namespace keystone argument is hardcoded. For a single-service operator this is correct, but if the pattern is templated for other OpenStack services, it will need parameterization.

**Code**:
```
Command: []string{
    "oslopolicy-checker",
    "--namespace", "keystone",
    "--policy", "/etc/keystone/keystone.conf.d/policy.yaml",
},
```

**Action Required**: No action required for current scope. If this pattern is reused for other services, parameterize the namespace argument.

**Suggested Fix**:
```
No change needed now. If reused for other services, replace the hardcoded "keystone" with a parameter derived from the CR or operator configuration.
```

---

## Summary

This is a well-structured PR that follows established codebase patterns closely. The reconciler, unit tests (16 test functions), integration tests (3 gating tests), Helm RBAC updates, and documentation are comprehensive and well-organized. The most important issues are a doc/code mismatch on the requeue interval (30s vs 15s), a dead test helper function, and the absence of a Chainsaw E2E test despite the project having E2E coverage for every other feature.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 6 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge after fixing the documentation requeue interval mismatch (30s → 15s) and removing or using the dead driveFullReconciliationWithPolicy helper. The missing Chainsaw E2E test and backoffLimit divergence should be discussed with the team but need not block the merge.

PR #173, opened 2026-04-10
perf: Parallelize Tempest tests and reduce default uWSGI threads

berendt:
Raise Tempest --concurrency from 1 to 4 to run the API test suite in parallel, and drop the default uWSGI threads from 2 to 1 so Keystone workers scale out via processes rather than per-worker threading.

AI-assisted: Claude Code

### DIFF
--- a/docs/reference/keystone-crd.md
+++ b/docs/reference/keystone-crd.md
@@ -121,7 +121,7 @@ status:
 | `policyOverrides` | [`*PolicySpec`](#policyspec) | No | `nil` | Custom oslo.policy rules. |
 | `autoscaling` | [`*AutoscalingSpec`](#autoscalingspec) | No | `nil` | Horizontal pod autoscaling configuration. When set, an HPA is created targeting the `{name}-api` Deployment. When removed, the HPA is deleted (CC-0038). |
 | `resources` | [`*corev1.ResourceRequirements`](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#resources) | No | See below | CPU and memory requests and limits for the Keystone API container. When unset, the defaulting webhook injects sensible defaults to ensure Burstable QoS class and enable HPA utilization calculations (CC-0042). |
-| `uwsgi` | [`*UWSGISpec`](#uwsgispec) | No | `nil` | uWSGI application server parameters. When set, the operator uses these values for the Deployment container command. When `nil`, hardcoded defaults (processes=2, threads=2, httpKeepAlive=true) are used in the reconciler, preserving backward compatibility for existing CRs (CC-0040). |
+| `uwsgi` | [`*UWSGISpec`](#uwsgispec) | No | `nil` | uWSGI application server parameters. When set, the operator uses these values for the Deployment container command. When `nil`, hardcoded defaults (processes=2, threads=1, httpKeepAlive=true) are used in the reconciler (CC-0040). |
 | `extraConfig` | `map[string]map[string]string` | No | `nil` | Free-form INI sections for additional configuration. |
 
 ### CEL Validation Rules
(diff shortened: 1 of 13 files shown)


berendt:



## BLOCKING (1)

### B-001: Breaking change to CRD default mischaracterized as backward-compatible
**File**: `operators/keystone/internal/controller/reconcile_deployment.go:291-298` — **Fix**: ASK — **Confidence**: verified

**Problem**: The +kubebuilder:default annotation on UWSGISpec.Threads changes from 2 to 1. Existing CRs with spec.uwsgi: {} (relying on schema defaults) will have their threads silently halved on the next reconcile after a CRD upgrade. The comment on line 293 still claims 'preserving backward compatibility for existing CRs', which is now false since the prior default was threads=2.

**Code**:
```
// uwsgiCommand constructs the uWSGI container command from the given spec.
// When uwsgi is nil (existing CRs without spec.uwsgi), hardcoded defaults
// (processes=2, threads=1, httpKeepAlive=true) are used to preserve backward
// compatibility.
func uwsgiCommand(uwsgi *keystonev1alpha1.UWSGISpec) []string {
	processes := int32(2)
	threads := int32(1)
```

**Action Required**: Update the comment to reflect the intentional breaking change and document the migration impact in a CHANGELOG or upgrade notes. The backward compatibility claim must be removed or corrected.

**Suggested Fix**:
```
Update the comment on lines 291-294 to:
// uwsgiCommand constructs the uWSGI container command from the given spec.
// When uwsgi is nil (existing CRs without spec.uwsgi), hardcoded defaults
// (processes=2, threads=1, httpKeepAlive=true) are used as defaults.
Also document the threads default change from 2 to 1 in upgrade/migration notes.
```

**Related**: W-001

---

## WARNING (3)

### W-001: Stale Documentation: backward compatibility claim in keystone-crd.md
**File**: `docs/reference/keystone-crd.md:124` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The documentation at line 124 and lines 211-212 still states 'preserving backward compatibility for existing CRs', which was accurate when threads defaulted to 2 but is now misleading after the change to threads=1.

**Code**:
```
| `uwsgi` | [`*UWSGISpec`](#uwsgispec) | No | `nil` | uWSGI application server parameters. When set, the operator uses these values for the Deployment container command. When `nil`, hardcoded defaults (processes=2, threads=1, httpKeepAlive=true) are used in the reconciler, preserving backward compatibility for existing CRs (CC-0040). |
```

**Action Required**: Remove or rephrase the backward compatibility claim in both locations to accurately reflect that the defaults have changed.

**Suggested Fix**:
```
| `uwsgi` | [`*UWSGISpec`](#uwsgispec) | No | `nil` | uWSGI application server parameters. When set, the operator uses these values for the Deployment container command. When `nil`, hardcoded defaults (processes=2, threads=1, httpKeepAlive=true) are used in the reconciler (CC-0040). |
```

**Related**: B-001

---

### W-002: Scope Creep: Tempest concurrency change is unrelated to uWSGI defaults
**File**: `hack/ci-run-tempest.sh:112` — **Fix**: ASK — **Confidence**: verified

**Problem**: The Tempest --concurrency change from 1 to 4 is an independent CI performance optimization with no dependency on the uWSGI thread default change. Bundling them in one commit means reverting a Tempest test flakiness issue also reverts the uWSGI default, and git blame attributes both changes to the same motivation.

**Code**:
```
TEMPEST_CMD+=" --concurrency 4 --subunit"
```

**Action Required**: Split the Tempest concurrency change into a separate commit from the uWSGI threads default change.

**Suggested Fix**:
```
Move the Tempest concurrency changes in hack/ci-run-tempest.sh and hack/run-tempest.sh into a separate commit with its own commit message explaining the CI performance motivation.
```

**Related**: W-003, I-001

---

### W-003: Hardcoded concurrency value without documentation of prerequisites
**File**: `hack/ci-run-tempest.sh:112` — **Fix**: ASK — **Confidence**: likely

**Problem**: The concurrency jump from 1 to 4 assumes Keystone can handle 4 parallel Tempest workers. With processes=2 and threads=1 (the new defaults), Keystone can serve at most 2 concurrent requests. Running 4 Tempest workers against 2 request slots will cause request queuing and potential timeouts, leading to test flakiness.

**Code**:
```
TEMPEST_CMD+=" --concurrency 4 --subunit"
```

**Action Required**: Verify that the Tempest target Keystone deployment uses enough processes (e.g., 4+) to match the concurrency. Alternatively, make the concurrency configurable via an environment variable with a documented default.

**Suggested Fix**:
```
Either increase the Keystone processes default to match the concurrency (4+), or make the concurrency configurable via an environment variable such as TEMPEST_CONCURRENCY with a default that matches available request capacity.
```

**Related**: W-002, B-001

---

## INFO (1)

### I-001: Magic Number: concurrency value hardcoded in two scripts
**File**: `hack/ci-run-tempest.sh:112` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The concurrency value 4 is hardcoded in both ci-run-tempest.sh and run-tempest.sh with no named constant or shared variable, violating the project pattern of avoiding hardcoded lists that duplicate a dynamic source.

**Code**:
```
TEMPEST_CMD+=" --concurrency 4 --subunit"
```

**Action Required**: Extract the concurrency value into a shared variable or make it configurable via an environment variable in both scripts.

**Suggested Fix**:
```
TEMPEST_CONCURRENCY=${TEMPEST_CONCURRENCY:-4}
TEMPEST_CMD+=" --concurrency ${TEMPEST_CONCURRENCY} --subunit"
```

**Related**: W-002, W-003

---

## Summary

The PR makes a clean, well-propagated default change across CRD schema, webhook, reconciler, tests, E2E tests, and documentation — all 13 files are internally consistent with the new threads=1 default. The test updates correctly distinguish between assertions on defaults vs. explicit user-supplied values. However, the backward compatibility claims in comments and docs are now stale and misleading, and the Tempest concurrency increase is unrelated to the uWSGI change and risks test flakiness given the reduced request capacity.

| Severity | Count |
|----------|-------|
| BLOCKING | 1 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Do not merge as-is. The BLOCKING finding (stale backward compatibility claim masking a breaking default change) must be addressed — either by updating documentation and comments to acknowledge the breaking change, or by providing migration guidance. The Tempest concurrency change should ideally be split into a separate commit. The potential mismatch between 4 concurrent Tempest workers and 2 available request slots should be investigated before merge.

PR #179, opened 2026-04-12
feat(CC-0063): Add graceful shutdown with preStop hook and startup probe

berendt:
# GitHub Issue #160: Add graceful shutdown with preStop hook and startup probe

**Description**: The Keystone API Deployment currently has no `preStop` lifecycle hook, no explicit `terminationGracePeriodSeconds`, and no startup probe. This means that during rolling updates or node drains, in-flight requests can be dropped because the pod is removed from the Service endpoints and terminated simultaneously. Additionally, after a cold start (e.g. following a config change or db_sync), the liveness probe with its 15-second initial delay may kill the container before Keystone has finished initializing.

Add the following to the `keystone-api` container in `buildKeystoneDeployment`:

1. **PreStop hook** with a short sleep to allow kube-proxy/iptables to converge before the process receives SIGTERM:
   ```yaml
   lifecycle:
     preStop:
       exec:
         command: ["/bin/sh", "-c", "sleep 5"]
   ```

2. **`terminationGracePeriodSeconds: 30`** on the PodSpec to give uWSGI enough time to finish in-flight requests after SIGTERM.

3. **Startup probe** with a generous failure threshold so slow first-boot scenarios (large DB, cold caches) do not trigger a liveness-kill loop:
   ```yaml
   startupProbe:
     httpGet:
       path: /v3
       port: 5000
     failureThreshold: 30
     periodSeconds: 10
   ```
   While the startup probe is active, kubelet skips both the liveness and readiness probes, preventing premature restarts.

**Motivation**: Without a preStop hook, there is a race condition between endpoint removal and process termination: kube-proxy may still route traffic to a pod that has already received SIGTERM. The 5-second sleep gives the cluster networking stack time to propagate the endpoint change. Without a startup probe, the liveness probe (15s initial delay, 20s period) can restart the container if Keystone takes longer than ~35 seconds to start — a realistic scenario after a schema migration or with cold Memcached caches. Together, these changes eliminate two classes of transient errors during rollouts.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_deployment.go` (`buildKeystoneDeployment`)

**Acceptance Criteria**:
- [ ] `keystone-api` container has a `preStop` lifecycle hook executing `sleep 5`
- [ ] PodSpec sets `terminationGracePeriodSeconds: 30`
- [ ] `keystone-api` container has a startup probe on `/v3` port 5000 with `failureThreshold: 30` and `periodSeconds: 10`
- [ ] Existing liveness and readiness probes remain unchanged
- [ ] Unit tests verify the presence and values of the lifecycle hook, termination grace period, and startup probe
- [ ] Rolling update with `replicas >= 2` completes without dropping requests (verified manually or via integration test)

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/160


berendt:



## WARNING (1)

### W-001: Missing Chainsaw E2E test for graceful-shutdown feature

**Problem**: The project has 19 Chainsaw E2E test directories under tests/e2e/keystone/ covering every existing feature (autoscaling, basic-deployment, credential-rotation, fernet-rotation, network-policy, trust-flush, uwsgi, etc.). This PR adds a new operational feature but includes no corresponding E2E test that asserts the Deployment spec fields (terminationGracePeriodSeconds, startupProbe, lifecycle.preStop) on a real cluster. The existing basic-deployment chainsaw test only asserts availableReplicas > 0 and condition statuses — it does not inspect the pod spec.

**Code**:
```
// integration_test.go:1702-1704
// TestIntegration_GracefulShutdownSpec verifies that the preStop lifecycle hook,
// terminationGracePeriodSeconds, and startup probe survive a full reconciliation
// cycle through the API server (CC-0063, REQ-006).
```

**Action Required**: Add a tests/e2e/keystone/graceful-shutdown/ directory with a chainsaw-test.yaml that applies a Keystone CR and asserts the Deployment spec contains the expected terminationGracePeriodSeconds, startupProbe, and lifecycle.preStop fields — matching the pattern used by other E2E tests in the same suite.

**Suggested Fix**:
```
Add a tests/e2e/keystone/graceful-shutdown/ directory with a chainsaw-test.yaml that applies a Keystone CR and asserts the Deployment spec contains the expected terminationGracePeriodSeconds, startupProbe, and lifecycle.preStop fields.
```

---

## INFO (1)

### I-001: Startup probe TimeoutSeconds defaults to 1s
**File**: `reconcile_deployment.go:206-215` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The startup probe does not set TimeoutSeconds explicitly. The Kubernetes default is 1 second. During cold starts — specifically the scenario this probe is designed for (large DB, cold Memcached caches) — the /v3 endpoint might take longer than 1 second to respond if it performs any backend initialization on the first request.

**Code**:
```
StartupProbe: &corev1.Probe{
    ProbeHandler: corev1.ProbeHandler{
        HTTPGet: &corev1.HTTPGetAction{
            Path: "/v3",
            Port: intstr.FromInt32(5000),
        },
    },
    FailureThreshold: 30,
    PeriodSeconds:    10,
},
```

**Action Required**: No action required now. Monitor for slow-start timeouts in cold-start scenarios. If observed, consider setting TimeoutSeconds to a higher value (e.g. 3-5s). The risk is mitigated by the generous FailureThreshold of 30 and the fact that the probe only needs an HTTP response code.

**Suggested Fix**:
```
If slow-start timeouts are observed, add TimeoutSeconds: 5 to the startup probe configuration.
```

---

## Summary

This is a clean, well-scoped PR that correctly implements graceful shutdown with a 5-second preStop sleep for kube-proxy convergence, a 30-second terminationGracePeriodSeconds for uWSGI drain time, and a startup probe with a 5-minute window to prevent liveness-kill loops during cold starts. The code follows project conventions (ptr.To usage) and includes thorough unit and integration tests. The only meaningful gap is the absence of a Chainsaw E2E test, which the project consistently provides for every other feature.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 1 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Merge after adding a Chainsaw E2E test for the graceful-shutdown feature to maintain the project's established testing standard. The INFO finding about startup probe TimeoutSeconds is optional and can be addressed later if cold-start timeouts are observed in practice.

PR #181, opened 2026-04-12
feat(CC-0064): Add schema drift detection before DatabaseReady

berendt:
# GitHub Issue #164: Add schema drift detection before setting DatabaseReady condition

**Description**: The `DatabaseReady` condition is currently set to True as soon as the `db_sync` Job completes successfully. There is no verification that the database schema actually matches what the running Keystone version expects. Add a pre-flight schema version check that queries the `alembic_version` table and compares the result against the expected migration head before setting `DatabaseReady=True`.

Implement this as a lightweight init container or a separate validation Job that runs after `db_sync` completes:

1. Connect to the database using the existing connection string from the ConfigMap
2. Execute `SELECT version_num FROM alembic_version` (single row, single column)
3. Compare against the expected Alembic head revision for the deployed Keystone version
4. If the version matches → exit 0 (schema is consistent)
5. If the version does not match → exit 1 with a descriptive error message including both expected and actual versions

Additionally, surface the detected schema version in the `DatabaseReady` condition message so operators can see it via `kubectl get keystone -o yaml` without needing database access.

The expected Alembic head can be obtained by running `keystone-manage db_version` in the Keystone image. This value should be captured during the `db_sync` Job (or a follow-up check) and compared against the live database.

**Motivation**: Schema drift can occur in several scenarios: a manual `ALTER TABLE` was run against the database, a partial migration failed silently (Alembic marked it complete but a DDL statement was rolled back by the storage engine), or the database was restored from a backup that predates the last migration. In all cases, the operator currently reports `DatabaseReady=True` based on a stale Job completion status, while Keystone encounters runtime errors (missing columns, type mismatches, integrity constraint violations). Detecting drift at the operator level enables early alerting and prevents rollout of incompatible API versions.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_database.go` (add schema verification step after db_sync completion)
- `operators/keystone/internal/controller/reconcile_database_test.go`

**Acceptance Criteria**:
- [ ] After `db_sync` Job succeeds, schema version is verified against the live database
- [ ] `DatabaseReady` condition message includes the detected Alembic version (e.g. `DatabaseSynced at revision abc123`)
- [ ] Schema mismatch sets `DatabaseReady=False` with reason `SchemaDriftDetected` and a message showing expected vs. actual version
- [ ] Schema check is idempotent and read-only (SELECT only, no DDL)
- [ ] Schema check uses the same image, SecurityContext, and config mounts as the db_sync Job
- [ ] Schema check Job has `backoffLimit: 2` and `ttlSecondsAfterFinished: 300`
- [ ] Requeue with `RequeueDatabaseWait` interval while schema check Job is running
- [ ] Unit tests cover: matching version passes, mismatched version fails with descriptive reason, missing alembic_version table fails gracefully

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/164


berendt:



## WARNING (7)

### W-001: buildSchemaCheckJob does not reuse buildDBJob helper
**File**: `reconcile_database.go:572-672` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Extract shared template logic to avoid duplication drift

**Problem**: buildSchemaCheckJob reimplements the identical volume/volumeMount/securityContext/restartPolicy structure independently instead of delegating to the shared buildDBJob helper used by buildDBSyncJob, buildExpandJob, buildMigrateJob, and buildContractJob. When buildDBJob is updated (e.g., when CC-0042 adds resource limits), buildSchemaCheckJob will silently diverge.

**Code**:
```
func buildSchemaCheckJob(keystone *keystonev1alpha1.Keystone, configMapName string) *batchv1.Job {
	backoffLimit := int32(2)
	ttl := int32(300)
	// ... 100 lines of standalone Job construction
```

**Action Required**: Refactor buildDBJob to accept optional overrides (backoffLimit, TTL) or extract the shared pod spec into a helper, then have buildSchemaCheckJob call it with the custom command, backoffLimit=2, and TTL=300.

**Suggested Fix**:
```
Refactor buildDBJob to accept optional overrides (backoffLimit, TTL) or extract the shared pod spec into a reusable helper function. Then have buildSchemaCheckJob delegate to that helper with its custom command, backoffLimit=2, and TTL=300, mirroring how buildDBSyncJob and the other Job builders work.
```

**Related**: W-002

---

### W-002: No resource limits on schema-check Job containers
**File**: `reconcile_database.go:648-651` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Kubernetes Resource Limits

**Problem**: The schema-check container has no resources.requests or resources.limits. This is consistent with the existing buildDBJob pattern which also lacks resource limits, and is tracked under CC-0042. Noting for visibility since it applies to all Job-type workloads in this operator.

**Code**:
```
Containers: []corev1.Container{{
	Name:            "schema-check",
	Image:           fmt.Sprintf("%s:%s", keystone.Spec.Image.Repository, keystone.Spec.Image.Tag),
	Command:         []string{"/bin/sh", "-eu", "-c", schemaCheckScript},
```

**Action Required**: Address as part of CC-0042 when resource limits are added to all Job-type workloads. If buildSchemaCheckJob is refactored to reuse buildDBJob (W-001), this will be resolved automatically.

**Suggested Fix**:
```
Tracked under CC-0042. If W-001 is addressed first, resource limits will be inherited from the shared helper. Otherwise, add explicit resources.requests and resources.limits to the schema-check container spec.
```

**Related**: W-001

---

### W-003: Missing E2E (Chainsaw) test for schema drift detection

**Problem**: The project has 19 Chainsaw E2E test directories under tests/e2e/keystone/ covering every existing feature (basic-deployment, upgrade-flow, brownfield-database, etc.). This PR adds a significant new feature but includes no Chainsaw E2E test for schema drift detection.

**Action Required**: Create tests/e2e/keystone/schema-drift-detection/ with a chainsaw-test.yaml that deploys a Keystone CR, waits for DatabaseReady=True, and asserts the condition message contains 'revision verified' and the schema-check Job exists.

**Suggested Fix**:
```
Create tests/e2e/keystone/schema-drift-detection/chainsaw-test.yaml covering at minimum: (1) after a normal deployment, DatabaseReady=True with message containing 'revision verified', and (2) the schema-check Job exists and completes successfully.
```

**Related**: I-002

---

### W-004: Condition reason strings are inline literals, inconsistent with NetworkPolicy pattern
**File**: `reconcile_database.go:170-181` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Extract repeated condition strings into typed constants

**Problem**: The new condition reason strings 'SchemaDriftDetected' and 'SchemaCheckInProgress' are inline string literals. reconcile_networkpolicy.go extracts condition reasons into typed constants. The new code follows the established convention in reconcile_database.go where existing reasons (DBSyncFailed, DBSyncInProgress, DatabaseSynced) are also inline literals, but the inconsistency between files is worth noting.

**Code**:
```
Reason:             "SchemaDriftDetected",
...
Reason:             "SchemaCheckInProgress",
```

**Action Required**: Consider extracting all condition reason strings in reconcile_database.go into typed constants, matching the pattern established in reconcile_networkpolicy.go. This could be done in a separate cleanup PR.

**Suggested Fix**:
```
Extract condition reason strings into typed constants at the top of reconcile_database.go, e.g. const conditionReasonSchemaDriftDetected = "SchemaDriftDetected", and use the constants in the condition updates. Consider doing this for all existing inline reason strings in the file as a separate cleanup.
```

---

### W-005: pymysql.OperationalError handler could theoretically leak credentials with non-pymysql drivers
**File**: `reconcile_database.go:599-603` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The OperationalError handler prints e.args[0] which is the MySQL error code (integer) for pymysql. This is safe for pymysql specifically. However, if the driver were changed to one whose OperationalError subclass includes the connection string in args[0], credentials could leak to pod logs.

**Code**:
```
except pymysql.OperationalError as e:
    print(f"Failed to connect to database: error {e.args[0]}", file=sys.stderr)
    sys.exit(1)
except Exception as e:
    print(f"Failed to connect to database: {type(e).__name__}", file=sys.stderr)
    sys.exit(1)
```

**Action Required**: Verify that pymysql's OperationalError.args[0] is always an integer error code across all supported driver versions. If the operator may switch drivers in the future, consider printing only a generic message or explicitly casting args[0] to int.

**Suggested Fix**:
```
Verify pymysql's error formatting guarantees. If there is any risk of driver change, replace e.args[0] with an explicit int cast or use only the error code number: print(f"Failed to connect to database: error code {int(e.args[0])}", file=sys.stderr)
```

**Related**: W-006

---

### W-006: Python script does not close cursor explicitly before conn.close()
**File**: `reconcile_database.go:612-613` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The cursor cur is never explicitly closed before conn.close(). While pymysql's Connection.close() implicitly closes open cursors and the process exits immediately after, explicit cursor close is a best practice.

**Code**:
```
rows = cur.fetchall()
conn.close()
```

**Action Required**: Add cur.close() before conn.close().

**Suggested Fix**:
```
rows = cur.fetchall()
cur.close()
conn.close()
```

**Related**: W-005

---

### W-007: keystone-manage db_version comparison assumes single-head migration model
**File**: `reconcile_database.go:618-628` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The script handles branched migrations on the actual side (sorts and joins multiple alembic_version rows). However, keystone-manage db_version output is parsed by taking only the first whitespace-delimited token (parts[0]). If Keystone's branched migration model results in db_version outputting multiple heads, only the first would be compared, causing a permanent false mismatch.

**Code**:
```
result = subprocess.run(
    ["keystone-manage", "--config-dir=/etc/keystone/keystone.conf.d/", "db_version"],
    capture_output=True, text=True)
...
expected = parts[0]
if actual == expected:
```

**Action Required**: Verify keystone-manage db_version output format for branched migrations against the actual Keystone image. If it can output multiple heads, parse and sort them the same way the actual version is handled.

**Suggested Fix**:
```
Verify keystone-manage db_version output format with the target Keystone image. If it can return multiple heads (space or newline separated), parse all of them, sort, and join with commas before comparing against the actual version string.
```

---

## INFO (2)

### I-001: Documentation added and cross-referenced correctly
**File**: `docs/reference/keystone-schema-drift-detection.md` — **Fix**: ASK — **Confidence**: verified

**Problem**: No problem — this is a positive observation. The new documentation is comprehensive and well-structured, covering data flow, Job specification, error scenarios, troubleshooting, limitations, and interactions with the upgrade flow. Existing docs were updated to reference the new feature.

**Action Required**: No action required.

**Suggested Fix**:
```
No fix needed.
```

---

### I-002: Test coverage is thorough for unit tests
**File**: `reconcile_database_test.go` — **Fix**: ASK — **Confidence**: verified

**Problem**: No problem — this is a positive observation. The PR adds 11 new test functions covering Job builder properties, reconciler states, managed mode, stale detection, and ordering. Existing tests were correctly updated to include the completed schema-check Job as a prerequisite.

**Action Required**: No action required. Consider adding E2E tests (see related finding).

**Suggested Fix**:
```
No fix needed. See the missing E2E test finding for additional test coverage.
```

**Related**: W-003

---

## Summary

The PR implements schema drift detection cleanly by inserting a verification Job between db_sync completion and DatabaseReady=True. The Python script is well-designed (read-only, avoids credential leakage, handles edge cases). Unit test coverage is thorough with 11 new tests plus updates to 6 existing tests, and documentation is comprehensive. The main concerns are the duplicated Job construction pattern (drift risk when CC-0042 lands) and the missing E2E test in a project that otherwise has Chainsaw tests for every feature.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 7 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing W-001 (refactor buildSchemaCheckJob to reuse buildDBJob) and W-003 (add Chainsaw E2E test). The remaining warnings are lower priority and can be handled in follow-up work.

Issue #186, opened 2026-04-12
Fix updateStatus to preserve the original reconcile error


berendt:
**Category**: refactoring | **Scope**: Small

**Description**: In keystone_controller.go:139-145, when r.Status().Update() fails, the original reconcileErr is discarded and only the status update error is returned. Refactor updateStatus to use errors.Join or a multi-error pattern that preserves both the original reconcile error and the status update error, ensuring the root cause is always visible in controller logs.

**Motivation**: When a reconcile fails AND the subsequent status update also fails, the root cause of the original failure is silently lost. This makes debugging production issues significantly harder because the logs only show 'failed to update status' without any indication of what actually went wrong.

**Affected Areas**:
- `operators/keystone/internal/controller/keystone_controller.go`

**Acceptance Criteria**:
- [ ] Both the original reconcile error and the status update error are returned when both fail
- [ ] Controller logs contain the root cause error, not just the status update failure
- [ ] Unit test verifies error preservation in the dual-failure case
- [ ] Single-failure cases (reconcile fails, status succeeds and vice versa) still work correctly

berendt:
Fixed with https://github.com/C5C3/forge/pull/196.

PR #196, opened 2026-04-12
feat(CC-0068): Fix updateStatus to preserve original reconcile error

berendt:
# GitHub Issue #186: Fix updateStatus to preserve the original reconcile error

**Category**: refactoring | **Scope**: Small

**Description**: In keystone_controller.go:139-145, when r.Status().Update() fails, the original reconcileErr is discarded and only the status update error is returned. Refactor updateStatus to use errors.Join or a multi-error pattern that preserves both the original reconcile error and the status update error, ensuring the root cause is always visible in controller logs.

**Motivation**: When a reconcile fails AND the subsequent status update also fails, the root cause of the original failure is silently lost. This makes debugging production issues significantly harder because the logs only show 'failed to update status' without any indication of what actually went wrong.

**Affected Areas**:

**Acceptance Criteria**:
- [ ] Both the original reconcile error and the status update error are returned when both fail
- [ ] Controller logs contain the root cause error, not just the status update failure
- [ ] Unit test verifies error preservation in the dual-failure case
- [ ] Single-failure cases (reconcile fails, status succeeds and vice versa) still work correctly

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/186


berendt:



## WARNING (1)

### W-001: Test assertion comment misleads about error wrapping
**File**: `operators/keystone/internal/controller/keystone_controller_test.go:858` — **Fix**: ASK — **Confidence**: verified

**Problem**: The test creates `statusErr := fmt.Errorf("status update failed")` and asserts `errors.Is(err, statusErr)` with the comment "errors.Is must match the original status update error". This subtly conflates the interceptor-returned error with the wrapped error inside `errors.Join`. The assertion works because `%w` wrapping preserves `errors.Is` traversal, but the naming and comment could mislead a future reader into thinking the raw `statusErr` is directly present in the join — it is actually behind a `%w` wrap.

**Code**:
```
g.Expect(errors.Is(err, statusErr)).To(BeTrue(),
    "errors.Is must match the original status update error")
```

**Action Required**: Clarify the test comment or variable naming to indicate that `statusErr` is matched through `errors.Is` unwrapping, not by direct presence in the joined error.

**Suggested Fix**:
```
Consider rewording the assertion message to: "errors.Is must unwrap through the joined error to find the original status update error" and/or renaming the variable to make the wrapping relationship clearer.
```

**Related**: I-001

---

## INFO (1)

### I-001: Documentation imprecise about errors.Join nil behavior
**File**: `docs/reference/keystone-reconciler.md:228` — **Fix**: ASK — **Confidence**: verified

**Problem**: `errors.Join(nil, wrappedErr)` returns a `*joinError` wrapping one error, not the raw error itself. The documentation states it "returns only the status update error", which is functionally accurate but pedantically imprecise — a direct equality check (`err == wrappedErr`) would return false.

**Code**:
```
errors.Join(nil, statusErr) discards the nil argument per the Go spec, returning only
the status update error.
```

**Action Required**: Consider rewording to clarify that `errors.Join(nil, wrappedErr)` returns a joined error containing only the status update error, rather than implying it returns the raw error directly.

**Suggested Fix**:
```
Consider rewording to: "errors.Join(nil, statusErr) discards the nil argument per the Go spec, returning a joined error containing only the status update error."
```

**Related**: W-001

---

## Summary

This is a clean, well-executed PR. The one-line code change from `fmt.Errorf` to `errors.Join` correctly preserves both errors in the dual-failure path while maintaining backward-compatible behavior. The test suite is thorough with 8 cases covering all truth-table rows, and documentation updates including a truth table are a strong addition. Only minor clarity issues were found — no correctness problems.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 1 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Merge as-is. The two findings are minor clarity improvements that can optionally be addressed in a follow-up.

gndrmnn (approved):

PR #198, opened 2026-04-13
fix(tempest): Split core and plugin suites into sequential stestr phases

berendt:
tempest.api.identity.v3.test_tokens.TokensV3Test.test_validate_token races against keystone_tempest_plugin ServiceProvidersTest under parallel execution. The test POSTs a token then GETs it and asserts the two response bodies are equal. Keystone re-resolves the list of enabled federation service providers on every token issue/validate and injects it into the response body -- the token itself does not cache it -- so an SP cleanup landing in the ~20 ms window between POST and GET causes the two responses to diverge on service_providers and the assertion fails.

Upstream's keystone-tempest gate sets tempest_test_regex to "keystone_tempest_plugin" and therefore never runs the core identity suite in the same invocation as the plugin, so the race is invisible upstream. We run both suites for fuller coverage.

Split include-tests.txt into two phase files at runtime (tempest.* and keystone_tempest_plugin.*) and invoke stestr twice in the same workspace. Each phase still runs at TEMPEST_CONCURRENCY internally. Subunit v2 stream-concatenation produces a single tempest.subunit that yields the same JUnit XML the CI upload step already expects. Both runners enforce the invariant that every non-comment line lands in exactly one phase. Overall exit code is max of the two phases so both always run.

AI-assisted: Claude Code

### DIFF
--- a/docs/reference/tempest-test-infrastructure.md
+++ b/docs/reference/tempest-test-infrastructure.md
@@ -216,6 +216,51 @@ For Keystone, two patterns include all identity-related tests:
 | `tempest.api.identity` | Core Tempest identity API tests |
 | `keystone_tempest_plugin.tests` | Keystone-specific plugin tests |
 
+#### Scope-split invariant
+
+Both runners (`hack/ci-run-tempest.sh` and `hack/run-tempest.sh`) split this
+include list into two phase files at runtime and run `stestr` twice in the
+same workspace:
+
+1. `phase-1-core.txt` — every non-comment line that starts with `tempest.`
+2. `phase-2-plugin.txt` — every non-comment line that starts with
+   `keystone_tempest_plugin.`
+
+The two phases run **sequentially**; within each phase `stestr` runs at
+`TEMPEST_CONCURRENCY` (default 4).
+
+Each runner enforces that every non-comment, non-empty line in
+`include-tests.txt` lands in exactly one phase. A line with any other prefix
+causes the runner to abort with a clear error — this guards against silent
+drops when new include patterns are added. Both phases must be non-empty.
+
+**Why sequential.** Keystone re-resolves the list of enabled federation
+service providers on every `POST /v3/auth/tokens` and `GET /v3/auth/tokens`
+call and injects it into the response body; the token itself does not cache
+it. `tempest.api.identity.v3.test_tokens.TokensV3Test.test_validate_token`
+issues a token via POST, validates it via GET, and asserts the two response
+bodies are equal. When `keystone_tempest_plugin.tests.api.identity.v3.
+test_service_providers.ServiceProvidersTest` runs concurrently on another
+stestr worker, its per-test `addCleanup` deletes the service provider it
+created — and if that cleanup lands in the ~20 ms window between
+`test_validate_token`'s POST and GET, the two responses diverge on the
+`service_providers` key and the assertion fails.
+
+Upstream's `openstack/keystone` `keystone-tempest` gate job sets
+`tempest_test_regex: "keystone_tempest_plugin"` and therefore only runs
+`keystone_tempest_plugin.*` tests — the core `tempest.api.identity.*` suite
+(which contains `test_validate_token`) never runs in the same Tempest
+invocation as the service-providers tests, so upstream never observes this
+race. We run both suites for fuller coverage and replicate the isolation by
+running them in two separate stestr invocations.
+
+The two phases each emit a subunit stream (`phase-1-core.subunit`,
+`phase-2-plugin.subunit`); the runner concatenates them into
+`tempest.subunit` and converts that to JUnit XML. Subunit v2 is
+stream-concatenation safe by design. The overall exit code is the maximum
+of the two phase exit codes, so both phases always run and any failure in
+either is reported.
+
 ### exclude-tests.txt
 
 **Location:** `tests/tempest/<service>/exclude-tests.txt`
@@ -312,7 +357,7 @@ SERVICE=keystone hack/run-tempest.sh
 | Pre-flight checks | Validates `SERVICE` is set, required tools (`docker`, `kubectl`, `yq`) are installed, Docker is running, service config directory exists, `test-refs.yaml` exists | Exits 1 with descriptive error |
 | Build Tempest image | Resolves versions from `test-refs.yaml`, builds with `docker build` using named build contexts pointing to GHCR base images | Exits on build failure (set -e) |
 | Extract admin password | Reads `keystone-admin` secret from the K8s cluster via `kubectl get secret` | Exits 1 if secret is not found or empty |
-| Run Tempest | Mounts `tempest.conf`, `include-tests.txt`, `exclude-tests.txt` into container; initializes Tempest workspace; runs `stestr run --subunit`; converts to JUnit XML via `subunit2junitxml` | Returns Tempest exit code |
+| Run Tempest | Mounts `tempest.conf`, `phases/phase-1-core.txt`, `phases/phase-2-plugin.txt`, `exclude-tests.txt` into container; initializes Tempest workspace; runs `stestr run --subunit` once per phase; concatenates the subunit streams; converts to JUnit XML via `subunit2junitxml` | Returns max of the two phase exit codes |
 
 **Output files:**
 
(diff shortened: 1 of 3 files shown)


berendt:



## BLOCKING (1)

### B-001: Missing JUnit XML failure guard in run-tempest.sh
**File**: `hack/run-tempest.sh:334-337` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Apply bug fixes to all instances of the same pattern

**Problem**: ci-run-tempest.sh has a guard (lines 209-216) that catches the case where stestr run --subunit exits 0 despite test failures by grepping the JUnit XML for failures or errors. The local run-tempest.sh lacks this guard entirely. The pre-existing version of run-tempest.sh had this same guard, but the refactoring removed it without replacement.

**Code**:
```
        cat /output/phase-1-core.subunit /output/phase-2-plugin.subunit > /output/tempest.subunit
        subunit2junitxml < /output/tempest.subunit > /output/tempest-results.xml 2>/dev/null || true
        exit ${overall_rc}
```

**Action Required**: Add the JUnit XML failure guard between the subunit2junitxml call and exit, matching ci-run-tempest.sh.

**Suggested Fix**:
```
        cat /output/phase-1-core.subunit /output/phase-2-plugin.subunit > /output/tempest.subunit
        subunit2junitxml < /output/tempest.subunit > /output/tempest-results.xml 2>/dev/null || true
        if [[ ${overall_rc} -eq 0 && -f /output/tempest-results.xml ]]; then
          if grep -qE 'failures="[1-9]|errors="[1-9]' /output/tempest-results.xml; then
            echo 'ERROR: Tempest reported test failures.'
            overall_rc=1
          fi
        fi
        exit ${overall_rc}
```

**Related**: W-004

---

## WARNING (5)

### W-001: Grep for TOTAL_PATTERNS counts lines with leading whitespace that phase greps won't match
**File**: `hack/ci-run-tempest.sh:141-142` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Distinguish connectivity failures from valid negative responses

**Problem**: The total-patterns regex '^[[:space:]]*[^#[:space:]]' counts lines with leading whitespace as non-comment patterns. But the phase-split greps ('^tempest.' and '^keystone_tempest_plugin.') require the pattern to start at column 1. A line like '  tempest.api.identity' would be counted in TOTAL_PATTERNS but not captured by either phase grep, causing the invariant check to fire with a misleading error message.

**Code**:
```
TOTAL_PATTERNS=$(grep -cE '^[[:space:]]*[^#[:space:]]' \
  "${CONFIG_DIR}/include-tests.txt" || true)
```

**Action Required**: Tighten the total-patterns regex to match the same anchor as the phase greps by removing the leading whitespace tolerance.

**Suggested Fix**:
```
TOTAL_PATTERNS=$(grep -cE '^[^#[:space:]]' \
  "${CONFIG_DIR}/include-tests.txt" || true)
```

---

### W-002: Grep for TOTAL_PATTERNS in run-tempest.sh has same leading whitespace mismatch
**File**: `hack/run-tempest.sh:268` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Distinguish connectivity failures from valid negative responses

**Problem**: Same issue as in ci-run-tempest.sh: the total-patterns regex tolerates leading whitespace but the phase-split greps do not, causing a mismatch in pattern counting.

**Code**:
```
TOTAL_PATTERNS=$(grep -cE '^[[:space:]]*[^#[:space:]]' \
  "${CONFIG_DIR}/include-tests.txt" || true)
```

**Action Required**: Apply the same regex tightening as in ci-run-tempest.sh.

**Suggested Fix**:
```
TOTAL_PATTERNS=$(grep -cE '^[^#[:space:]]' \
  "${CONFIG_DIR}/include-tests.txt" || true)
```

**Related**: W-001

---

### W-003: Documentation diagram shows &&-chained commands but actual implementation uses || for exit code capture
**File**: `docs/reference/tempest-test-infrastructure.md:542-545` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Cross-reference documentation examples against implementation

**Problem**: The ASCII art diagram uses && between the two stestr invocations, but the actual implementation uses '|| overall_rc=$?' to capture exit codes and always runs both phases. With &&, a phase-1 failure would skip phase-2 — the exact opposite of the documented behavior ('both phases always run').

**Code**:
```
             stestr run --include-list phases/phase-1-core.txt ... --subunit \
               | tee /output/phase-1-core.subunit &&
             stestr run --include-list phases/phase-2-plugin.txt ... --subunit \
               | tee /output/phase-2-plugin.subunit &&
```

**Action Required**: Replace && with ; in the documentation diagram to accurately reflect non-short-circuiting behavior.

**Suggested Fix**:
```
             stestr run --include-list phases/phase-1-core.txt ... --subunit \
               | tee /output/phase-1-core.subunit;
             stestr run --include-list phases/phase-2-plugin.txt ... --subunit \
               | tee /output/phase-2-plugin.subunit;
             cat phase-1-core.subunit phase-2-plugin.subunit > tempest.subunit;
```

---

### W-004: Duplicated scope-split logic across two scripts without shared implementation
**File**: `hack/ci-run-tempest.sh:125-154` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Extract shared template logic to avoid duplication drift

**Problem**: The scope-split logic (grep into phases, invariant check, empty-phase check) is copy-pasted between ci-run-tempest.sh (lines 125-154) and run-tempest.sh (lines 256-280). As demonstrated by B-001, they have already diverged: the JUnit guard was lost in one copy. The invariant check, phase splitting, and validation code are ~25 lines of identical logic in each file.

**Code**:
```
# Duplicated block in ci-run-tempest.sh lines 125-154 and run-tempest.sh lines 256-280:
# grep into phases, invariant check, empty-phase check
```

**Action Required**: Extract the scope-split logic into a shared helper script (e.g., hack/split-tempest-phases.sh) that both runners source or call to reduce the surface for future drift.

**Suggested Fix**:
```
Extract the scope-split logic into a shared helper script (e.g., hack/split-tempest-phases.sh) that both runners source or call. This eliminates the duplication that already caused B-001's regression.
```

**Related**: B-001

---

### W-005: Missing tests for the scope-split logic
**File**: `tests/ci/` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Test Coverage for New Code

**Problem**: The project has a tests/ci/verify_ci_scripts.sh that validates structural properties of CI scripts. There is no test for the scope-split invariant logic itself — no test that an include-tests.txt with an unknown prefix is rejected, that an empty phase is rejected, or that the phase files contain the expected lines.

**Code**:
```
# No test coverage found in tests/ci/ for scope-split, phase-1, or phase-2 logic
```

**Action Required**: Add a tests/ci/verify_tempest_scope_split.sh (or equivalent) that exercises the split logic against known inputs: valid 2-prefix file, unknown-prefix file, single-prefix file, empty file, file with leading whitespace.

**Suggested Fix**:
```
Add a test script tests/ci/verify_tempest_scope_split.sh that exercises the split logic against known inputs: valid 2-prefix file, unknown-prefix file, single-prefix file, empty file, file with leading whitespace.
```

**Related**: W-004

---

## INFO (1)

### I-001: stestr reinitializes the test repository for each phase in the same workspace
**File**: `hack/ci-run-tempest.sh` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: Both run_phase calls execute stestr run in the same tempest workspace. The first stestr run creates .stestr/ and a test database. The second stestr run in the same workspace appends to the same database. If stestr's internal state from phase 1 affects phase 2 discovery or filtering, it could cause subtle interactions.

**Code**:
```
# Both phases execute stestr run in the same tempest workspace,
# sharing the .stestr/ directory and test database
```

**Action Required**: Verify that stestr's --include-list filters test IDs after discovery and that shared workspace state does not cause phase interaction. Based on stestr's documentation this should be safe, but worth confirming.

**Suggested Fix**:
```
No code change needed if stestr's --include-list filtering is confirmed to be independent of prior run state. If interactions are observed, add stestr init between phases or use separate workspaces.
```

---

## Summary

The PR is well-structured with thorough documentation explaining both the problem (race condition between core and plugin Tempest suites) and the solution (sequential phase execution). The invariant enforcement that prevents silent test drops is a strong design choice. However, a regression was introduced: the JUnit XML failure guard was dropped from run-tempest.sh during the refactor (B-001), and the documentation diagram contradicts the implementation's non-short-circuiting semantics (W-002). The duplicated scope-split logic has already demonstrated drift.

| Severity | Count |
|----------|-------|
| BLOCKING | 1 |
| CRITICAL | 0 |
| WARNING  | 5 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Do not merge until B-001 (missing JUnit XML failure guard in run-tempest.sh) is resolved — this is a regression from pre-existing behavior. The WARNING-level findings (W-001 regex mismatch, W-002 documentation inaccuracy, W-003 duplication, W-004 missing tests) should ideally be addressed before merge, with W-003 and W-004 acceptable as fast-follow if the team prefers.

PR #201, opened 2026-04-13
fix(tempest): Retry failed tests once serially to absorb race flakes

berendt:
After both phase-1-core and phase-2-plugin finish, the runners now extract any failed or errored testcases from the initial JUnit report and rerun them once with stestr --concurrency 1. Tests that pass on retry are rewritten in the JUnit report as flakes (failure/error children removed, suite counters decremented, system-out note added) so only tests that still fail after retry count against the job exit code. Resolves the intermittent SystemMemberTests.test_identity_list_role_assignments failure in the 2025.2 matrix row without serializing the whole plugin suite.

AI-assisted: Claude Code

### DIFF
--- a/.github/workflows/verify-container-images.yaml
+++ b/.github/workflows/verify-container-images.yaml
@@ -14,8 +14,10 @@ name: Verify Container Images
       - releases/**
       - patches/**
       - scripts/**
+      - hack/tempest/**
       - tests/container-images/**
       - tests/scripts/**
+      - tests/tempest/test_retry_helpers.py
       - .github/workflows/build-images.yaml
       - .github/workflows/verify-container-images.yaml
   pull_request:
@@ -24,8 +26,10 @@ name: Verify Container Images
       - releases/**
       - patches/**
       - scripts/**
+      - hack/tempest/**
       - tests/container-images/**
       - tests/scripts/**
+      - tests/tempest/test_retry_helpers.py
       - .github/workflows/build-images.yaml
       - .github/workflows/verify-container-images.yaml
 
@@ -67,3 +71,11 @@ jobs:
 
       - name: Test apply-constraint-overrides
         run: bash tests/scripts/test_apply_constraint_overrides.sh
+
+      - name: Install tempest retry helper test dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install -y -qq python3-defusedxml
+
+      - name: Test tempest retry helpers
+        run: python3 tests/tempest/test_retry_helpers.py
--- a/docs/reference/tempest-test-infrastructure.md
+++ b/docs/reference/tempest-test-infrastructure.md
@@ -27,6 +27,9 @@ and `build-images.yaml` dynamically discovers releases for the Tempest image pip
 | `tests/container-images/verify_tempest.sh` | Image verification script (PASS/FAIL counters) |
 | `hack/run-tempest.sh` | Local orchestration script for running Tempest against a kind cluster |
 | `hack/ci-run-tempest.sh` | CI-specific Tempest wrapper with port-forwarding and config generation (CC-0050) |
+| `hack/tempest/extract-failed.py` | Print anchored regex patterns for failed testcases in a JUnit report (used to build the retry include-list) |
+| `hack/tempest/merge-retry-junit.py` | Merge a retry subunit stream into a JUnit report, rewriting resolved failures as flakes |
+| `hack/tempest/run-tests.sh` | Shared in-container runner invoked by both runners; holds the phase + retry + exit-code logic so it stays identical between CI and local runs |
 | `Makefile` | `tempest-test` target delegates to `hack/run-tempest.sh` |
 | `.github/workflows/ci.yaml` | `tempest` job with release matrix (CC-0051) |
 | `.github/workflows/build-images.yaml` | `build-tempest` and `merge-tempest-image` jobs (release-parameterized via `generate-matrix`) |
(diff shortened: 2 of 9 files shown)


berendt:



## CRITICAL (1)

### C-001: Missing `run-tests.sh` — both runners will crash on execution
**File**: `hack/tempest/run-tests.sh` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Cross-check all instances of a claimed fix across files

**Problem**: Both `ci-run-tempest.sh` (line 126-127) and `run-tempest.sh` (line 258-259) stage a file that does not exist. The `hack/tempest/` directory contains only `extract-failed.py` and `merge-retry-junit.py`. The `install` command will exit non-zero with "No such file or directory", and since both scripts run under `set -euo pipefail`, this aborts execution before any Tempest test runs. The PR replaces ~50 lines of inline bash with `bash /etc/tempest/run-tests.sh`, but `run-tests.sh` was never committed.

**Code**:
```
install -m 0755 "${SCRIPT_DIR}/tempest/run-tests.sh" \
  "${OUTPUT_DIR}/config/run-tests.sh"
```

**Action Required**: Create `hack/tempest/run-tests.sh` containing the shared in-container runner logic that was removed from both scripts (the `set -euo pipefail`, tempest init, phase execution, subunit merge, retry invocation, exit-code derivation). This is the entire core of the PR and it is missing.

**Suggested Fix**:
```
Create `hack/tempest/run-tests.sh` containing the shared in-container runner logic (set -euo pipefail, tempest init, phase execution, subunit merge, retry invocation, exit-code derivation) that was removed from both scripts' inline bash blocks.
```

**Related**: W-002, I-002

---

## WARNING (3)

### W-001: Parametrized test variants can mask real failures via last-wins collapse
**File**: `hack/tempest/merge-retry-junit.py:44-50` — **Fix**: ASK — **Confidence**: likely — **Pattern**: Verify semantic equivalence between redundant validation layers

**Problem**: _OutcomeCollector collapses all parametrized variants to the same bare key using last-wins. If a test method has parametrized variants and on retry one succeeds but another fails, the outcome depends on execution order. If the succeeding variant runs last, both JUnit testcases are rewritten as flakes — including the variant that still fails. This suppresses a real failure from the exit code.

**Code**:
```
def status(self, test_id=None, test_status=None, **_: object) -> None:
    if not test_id or not test_status:
        return
    if test_status in {"success", "fail", "skip", "xfail", "uxsuccess"}:
        self.outcomes[_bare(test_id)] = test_status
```

**Action Required**: Use conservative semantics: only mark the bare ID as passed if ALL variant outcomes are success or uxsuccess. Alternatively, use the variant-level (non-bare) test_id as the outcome key and match JUnit testcases by their full classname.name.

**Suggested Fix**:
```
def status(self, test_id=None, test_status=None, **_: object) -> None:
    if not test_id or not test_status:
        return
    if test_status in {"success", "fail", "skip", "xfail", "uxsuccess"}:
        key = _bare(test_id)
        if key in self.outcomes and self.outcomes[key] == "fail":
            return
        self.outcomes[key] = test_status
```

---

### W-002: Documentation references non-existent file `hack/tempest/run-tests.sh`
**File**: `docs/reference/tempest-test-infrastructure.md:32` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Cross-reference documentation examples against implementation

**Problem**: The documentation lists `hack/tempest/run-tests.sh` as a key project file with a description, but the file does not exist in the repository. This is a direct consequence of the missing `run-tests.sh` file.

**Code**:
```
| `hack/tempest/run-tests.sh` | Shared in-container runner invoked by both runners; holds the phase + retry + exit-code logic so it stays identical between CI and local runs |
```

**Action Required**: Add the file (B-001) or remove the documentation entry until the file is committed.

**Suggested Fix**:
```
Remove this documentation table row until `hack/tempest/run-tests.sh` is actually committed, or commit the missing file first.
```

**Related**: C-001

---

### W-003: Missing unit tests for `extract-failed.py` and `merge-retry-junit.py`
**File**: `hack/tempest/merge-retry-junit.py:1` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Missing Tests

**Problem**: Both Python scripts contain non-trivial logic (XML parsing with regex generation, subunit stream parsing with JUnit counter manipulation) but have no tests. The project has an established test convention with existing verification scripts and shared helpers.

**Code**:
```
extract-failed.py: correct anchor generation, regex metacharacter escaping, parametrized variant stripping
merge-retry-junit.py: counter decrement correctness, handling of testcases with both <failure> and <error>
```

**Action Required**: Add tests covering at minimum the happy path, edge cases (empty failures, parametrized variants, mixed failure+error), and error cases (malformed XML, missing file) for both scripts.

**Suggested Fix**:
```
Add `tests/tempest/test_extract_failed.py` and `tests/tempest/test_merge_retry_junit.py` (or a combined bash verification script following the existing project pattern) covering happy path, edge cases (empty failures, parametrized variants, mixed failure+error), and error cases (malformed XML, missing file).
```

**Related**: W-001

---

## INFO (2)

### I-001: `_dec` does not guard against non-numeric attribute values
**File**: `hack/tempest/merge-retry-junit.py:61-64` — **Fix**: AUTO-FIX — **Confidence**: uncertain

**Problem**: `int()` will raise `ValueError` if the attribute is present but non-numeric (e.g., `failures="NaN"` from a buggy reporter). Although the runner's fallback error handling mitigates this, a defensive check would be more robust.

**Code**:
```
def _dec(elem: ET.Element, attr: str) -> None:
    value = int(elem.attrib.get(attr, "0"))
    if value > 0:
        elem.attrib[attr] = str(value - 1)
```

**Action Required**: Wrap in a try/except ValueError or validate before parsing.

**Suggested Fix**:
```
def _dec(elem: ET.Element, attr: str) -> None:
    try:
        value = int(elem.attrib.get(attr, "0"))
    except ValueError:
        return
    if value > 0:
        elem.attrib[attr] = str(value - 1)
```

---

### I-002: Stale documentation — diagram shows inline logic, not `run-tests.sh` invocation
**File**: `docs/reference/tempest-test-infrastructure.md:570-585` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The ASCII flow diagram shows the retry logic as individual inline bash commands, but the actual scripts now invoke `bash /etc/tempest/run-tests.sh` as a single command. The diagram does not match the actual docker run invocation.

**Code**:
```
             python3 /etc/tempest/extract-failed.py tempest-results.xml > retry-list.txt;
             stestr run --include-list retry-list.txt --concurrency 1 --subunit \
               | tee /output/retry.subunit;
```

**Action Required**: Add a comment in the diagram clarifying this is the internal logic of `run-tests.sh`, not the literal docker run command.

**Suggested Fix**:
```
Add a note above or below the relevant diagram section: "Note: The following shows the internal logic of run-tests.sh. Both runners invoke this via `bash /etc/tempest/run-tests.sh`."
```

**Related**: C-001, W-002

---

## Summary

The PR's design is sound — extracting inline bash into a shared runner and adding Python helpers for flake retry/rewrite is a good architectural approach. The Python scripts are well-structured with proper use of defusedxml for XML parsing and careful handling of JUnit semantics. However, the critical `hack/tempest/run-tests.sh` file was not committed, meaning both runner scripts will crash immediately on execution. The parametrized variant last-wins bug could silently mask real test failures.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 1 |
| WARNING  | 3 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Do not merge. The missing `hack/tempest/run-tests.sh` file makes the PR completely non-functional — both CI and local runners will abort on the `install` command. Once that file is committed, the parametrized variant collapse issue (B-002) should be addressed before merge to prevent incorrect flake classification. Unit tests for the new Python scripts are strongly recommended given their non-trivial logic.

Issue #206, opened 2026-04-14
Stop triggering a full Deployment rollout on every fernet/credential key rotation


berendt:
**Category**: improvement | **Scope**: Small

**Description**: `reconcileDeployment` (operators/keystone/internal/controller/reconcile_deployment.go:67-74) reads the `fernet-keys` and `credential-keys` Secrets, computes a SHA-256 of their `Data` map, and injects the digests as pod template annotations `keystone.c5c3.io/fernet-keys-hash` and `keystone.c5c3.io/credential-keys-hash` (reconcile_deployment.go:170-173). Because the rotation CronJobs (CC-0013, CC-0036) PATCH those Secrets on every run, the annotation changes, the pod template hash changes, and the entire Keystone Deployment is rolled. Remove the hash plumbing so that key rotation no longer triggers a rollout: drop `keysHash`/`fernetKeysHash`/`credentialKeysHash` (lines 33-61), drop the two annotations from `buildKeystoneDeployment`, and let the kubelet's Secret projection deliver the rotated keys to running pods through their existing volume mounts (`/etc/keystone/fernet-keys/`, `/etc/keystone/credential-keys/`).

**Motivation**: Keystone's fernet provider reads keys from the on-disk keys repository on each token operation, and Kubernetes Secret volume projections are updated in-place within roughly one kubelet sync period after the Secret changes. With `maxActiveKeys` defaulting to 3, the upstream multi-key overlap window already covers any brief skew between pods. The current design throws away those guarantees: every weekly fernet rotation and every credential rotation forces a full rolling restart of all Keystone replicas (3 by default), which (a) causes a capacity dip on what is the gating dependency for every other OpenStack service, (b) breaks long-running uWSGI connections and warm Memcached token caches, (c) multiplies the blast radius of any unrelated bug surfaced during a rollout, and (d) competes for PDB budget with legitimate voluntary disruptions like node drains. Rotation should be a transparent on-disk update, not a fleet-wide restart.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_deployment.go`
- `operators/keystone/internal/controller/reconcile_deployment_test.go`
- `tests/e2e/keystone/fernet-rotation/`
- `tests/e2e/keystone/credential-rotation/`

**Acceptance Criteria**:
- [ ] `keysHash`, `fernetKeysHash`, `credentialKeysHash` and their callers in `reconcileDeployment` are removed
[...]

PR #213, opened 2026-04-14
feat(CC-0074): Avoid full rollout on fernet/credential key rotation

berendt:
# GitHub Issue #206: Stop triggering a full Deployment rollout on every fernet/credential key rotation

**Description**: `reconcileDeployment` (operators/keystone/internal/controller/reconcile_deployment.go:67-74) reads the `fernet-keys` and `credential-keys` Secrets, computes a SHA-256 of their `Data` map, and injects the digests as pod template annotations `keystone.c5c3.io/fernet-keys-hash` and `keystone.c5c3.io/credential-keys-hash` (reconcile_deployment.go:170-173). Because the rotation CronJobs (CC-0013, CC-0036) PATCH those Secrets on every run, the annotation changes, the pod template hash changes, and the entire Keystone Deployment is rolled. Remove the hash plumbing so that key rotation no longer triggers a rollout: drop `keysHash`/`fernetKeysHash`/`credentialKeysHash` (lines 33-61), drop the two annotations from `buildKeystoneDeployment`, and let the kubelet's Secret projection deliver the rotated keys to running pods through their existing volume mounts (`/etc/keystone/fernet-keys/`, `/etc/keystone/credential-keys/`).

**Motivation**: Keystone's fernet provider reads keys from the on-disk keys repository on each token operation, and Kubernetes Secret volume projections are updated in-place within roughly one kubelet sync period after the Secret changes. With `maxActiveKeys` defaulting to 3, the upstream multi-key overlap window already covers any brief skew between pods. The current design throws away those guarantees: every weekly fernet rotation and every credential rotation forces a full rolling restart of all Keystone replicas (3 by default), which (a) causes a capacity dip on what is the gating dependency for every other OpenStack service, (b) breaks long-running uWSGI connections and warm Memcached token caches, (c) multiplies the blast radius of any unrelated bug surfaced during a rollout, and (d) competes for PDB budget with legitimate voluntary disruptions like node drains. Rotation should be a transparent on-disk update, not a fleet-wide restart.

**Affected Areas**:
- `tests/e2e/keystone/fernet-rotation/`
- `tests/e2e/keystone/credential-rotation/`

**Acceptance Criteria**:
- [ ] `keysHash`, `fernetKeysHash`, `credentialKeysHash` and their callers in `reconcileDeployment` are removed
- [ ] `buildKeystoneDeployment` no longer emits the `keystone.c5c3.io/fernet-keys-hash` or `keystone.c5c3.io/credential-keys-hash` pod annotations
- [ ] The fernet and credential key Secrets remain mounted as volumes; no behavioural change to in-pod key paths
- [ ] Existing `TestBuildKeystoneDeployment_*HashAnnotation`, `TestFernetKeysHash_*`, `TestCredentialKeysHash_*` and `TestReconcileDeployment_*KeysHashFromSecret` tests are removed or replaced with assertions that the annotations are absent
- [ ] A new unit test asserts that `buildKeystoneDeployment` produces a stable pod template hash across two invocations with different Secret contents
- [ ] The `fernet-rotation` and `credential-rotation` Chainsaw E2E tests are extended to assert that pod UIDs are unchanged after a rotation cycle (no rollout occurred)
- [ ] An E2E assertion proves a token issued before rotation still validates after rotation, demonstrating the multi-key overlap window covers the kubelet projection delay
- [ ] `architecture/docs/09-implementation/04-keystone-reconciler.md` is updated to document that key rotation is in-place and does not roll the Deployment

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/206


berendt:



## WARNING (4)

### W-001: Scope creep: CC-0075 planning commit in wrong PR
**File**: `.planwerk/` — **Fix**: ASK — **Confidence**: verified

**Problem**: Commit b223879 (plan for CC-0075: topologySpreadConstraints and PriorityClass support) belongs to a different feature and pollutes this branch's commit history.

**Code**:
```
b223879 plan(CC-0075): Add topologySpreadConstraints and PriorityClass support
```

**Action Required**: Remove the CC-0075 commit from this branch via interactive rebase or cherry-pick the relevant commits onto a clean branch.

**Suggested Fix**:
```
Rebase this branch to exclude commit b223879, or move it to the CC-0075 feature branch.
```

---

### W-002: Misleading no-op kubelet projection wait loop (fernet-rotation)
**File**: `tests/e2e/keystone/fernet-rotation/chainsaw-test.yaml:123-131` — **Fix**: ASK — **Confidence**: verified

**Problem**: The loop re-reads the Secret API object and compares it to AFTER_HASH, which was already read from the same API. Since nothing else mutates the Secret between the two reads, CURRENT_HASH will always equal AFTER_HASH on the first iteration, making the loop a no-op. The comment claims it waits for kubelet projection, but kubelet projection happens at the pod-level filesystem, not the Secret API object.

**Code**:
```
          # Allow time for kubelet to project new keys (up to 30s, avoids flakiness on slow clusters).
          for i in $(seq 1 15); do
            CURRENT_HASH=$(kubectl get secret keystone-fernet-fernet-keys -n openstack \
              -o jsonpath='{.data}' | sha256sum | cut -d' ' -f1)
            if [ "$CURRENT_HASH" = "$AFTER_HASH" ]; then
              break
            fi
            sleep 2
          done
```

**Action Required**: Either remove the loop entirely (the token validation covers the semantic guarantee) or replace it with a real kubelet projection check that exec-s into the pod and checks the projected files.

**Suggested Fix**:
```
Either remove the loop entirely, or replace with a real kubelet projection check:
          # Allow time for kubelet to project new keys (up to 60s).
          for i in $(seq 1 30); do
            IN_POD_KEYS=$(kubectl exec -n openstack "$POD" -- ls /etc/keystone/fernet-keys/ 2>/dev/null | sort)
            if [ -n "$IN_POD_KEYS" ]; then
              break
            fi
            sleep 2
          done
```

**Related**: W-003

---

### W-003: Misleading no-op kubelet projection wait loop (credential-rotation)
**File**: `tests/e2e/keystone/credential-rotation/chainsaw-test.yaml:106-114` — **Fix**: ASK — **Confidence**: verified

**Problem**: Identical issue to the fernet-rotation wait loop. The loop re-reads the Secret API object and compares it to AFTER_HASH which was already set from the same source, so CURRENT_HASH always equals AFTER_HASH on the first iteration. The loop is a no-op that gives a false sense of synchronization.

**Code**:
```
          # Allow time for kubelet to project new keys (up to 30s, avoids flakiness on slow clusters).
          for i in $(seq 1 15); do
            CURRENT_HASH=$(kubectl get secret keystone-credential-credential-keys -n openstack \
              -o jsonpath='{.data}' | sha256sum | cut -d' ' -f1)
            if [ "$CURRENT_HASH" = "$AFTER_HASH" ]; then
              break
            fi
            sleep 2
          done
```

**Action Required**: Apply the same fix as the fernet-rotation loop: either remove entirely or replace with a real kubelet projection check.

**Suggested Fix**:
```
Either remove the loop entirely, or replace with a real kubelet projection check analogous to the fernet-rotation fix.
```

**Related**: W-002

---

### W-004: Brittle index-based container assertion in VolumesMaintained test
**File**: `operators/keystone/internal/controller/reconcile_deployment_test.go:469` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Avoid brittle index-based container assertions in tests

**Problem**: The test hard-codes Containers[0] to get volume mounts. If a sidecar is added in the future, this test would either break or silently validate the wrong container. The volumes portion of the same test already uses a name-based map lookup pattern.

**Code**:
```
	mounts := deploy.Spec.Template.Spec.Containers[0].VolumeMounts
```

**Action Required**: Replace the index-based container access with a name-based lookup for the keystone-api container.

**Suggested Fix**:
```
	var mounts []corev1.VolumeMount
	for _, c := range deploy.Spec.Template.Spec.Containers {
		if c.Name == "keystone-api" {
			mounts = c.VolumeMounts
			break
		}
	}
	g.Expect(mounts).NotTo(BeEmpty(), "keystone-api container must exist with volume mounts")
```

---

## INFO (2)

### I-001: StablePodTemplate test doc comment accurately reflects limitation
**File**: `operators/keystone/internal/controller/reconcile_deployment_test.go:427-434` — **Fix**: ASK — **Confidence**: verified

**Problem**: The acceptance criteria mentions testing with different Secret contents, but since buildKeystoneDeployment no longer takes hash parameters, the function is inherently stable. The doc comment correctly acknowledges this limitation and defers to E2E tests for full coverage.

**Code**:
```
// TestBuildKeystoneDeployment_StablePodTemplate verifies that two calls to
// buildKeystoneDeployment with the same Keystone CR and configMapName return
// Deployments with deeply equal Spec.Template fields. ...
// It does not exercise scenarios with differing Secret contents as described
// in REQ-002; those must be covered by higher-level reconciliation tests.
```

**Action Required**: No action needed. The doc comment is accurate, and E2E pod-UID-stability assertions provide the real coverage.

**Suggested Fix**:
```
No change required.
```

---

### I-002: Credential-rotation E2E intentionally lacks token validation
**File**: `tests/e2e/keystone/credential-rotation/chainsaw-test.yaml` — **Fix**: ASK — **Confidence**: verified

**Problem**: The credential-rotation test has no token validation equivalent to the fernet-rotation test. This is correct because credential keys encrypt long-lived credentials in the database rather than signing auth tokens, so token validation is not applicable.

**Code**:
```
(no specific code snippet — absence of token validation step in credential-rotation test)
```

**Action Required**: No action needed. The asymmetry between the two tests is intentional and correct.

**Suggested Fix**:
```
No change required.
```

---

## Summary

A clean, well-scoped PR that correctly removes hash-triggered rollouts and replaces them with kubelet Secret projection. Production code changes are straightforward and complete — all hash functions, annotations, and Secret reads removed while volume mounts are preserved. The main issues are two misleading no-op wait loops in E2E tests that re-read the Secret API object instead of checking in-pod filesystems, and one brittle index-based container assertion. A stray CC-0075 planning commit should be excluded from this branch.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 4 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing the three WARNING findings: remove or fix the two no-op wait loops in E2E tests, fix the brittle container index assertion, and exclude the CC-0075 commit from this branch. The INFO findings require no action.

operators/keystone/internal/controller/reconcile_deployment_test.go
```diff
@@ -413,226 +409,110 @@ func TestReconcileDeployment_ServiceCreatedAlongsideDeployment(t *testing.T) {
...
-			"1": []byte("credential-key-1"),
-		},
+	annotations := deploy.Spec.Template.Annotations
+	if annotations != nil {
+		g.Expect(annotations).NotTo(HaveKey("keystone.c5c3.io/fernet-keys-hash"),
+			"pod template must not contain fernet-keys-hash annotation (CC-0074)")
+		g.Expect(annotations).NotTo(HaveKey("keystone.c5c3.io/credential-keys-hash"),
+			"pod template must not contain credential-keys-hash annotation (CC-0074)")
 	}
 }
```
gndrmnn: This `TestBuildKeystoneDeployment_NoHashAnnotations` test is pointless. We created the `fernet-keys-hash` and `credential-keys-hash`  key mappings ourselves. We control what happens to the mapping. Now that we remove them again, there is no need to add a unit test which tests for their absence.

Remove the test.

operators/keystone/internal/controller/reconcile_deployment_test.go
```diff
@@ -413,226 +409,110 @@ func TestReconcileDeployment_ServiceCreatedAlongsideDeployment(t *testing.T) {
...
+	}
+	g.Expect(mountPaths["config"]).To(Equal("/etc/keystone/keystone.conf.d/"))
+	g.Expect(mountPaths["fernet-keys"]).To(Equal("/etc/keystone/fernet-keys/"))
+	g.Expect(mountPaths["credential-keys"]).To(Equal("/etc/keystone/credential-keys/"))
 }
 
-func TestFernetKeysHash_SecretNotFound(t *testing.T) {
+// TestReconcileDeployment_NoSecretReadRequired verifies that reconcileDeployment
+// succeeds and creates a Deployment even when fernet-keys and credential-keys
+// Secrets do not exist, confirming the hash computation path is removed (CC-0074, REQ-001).
```
gndrmnn: The phrasing `confirming the hash computation path is removed` is superfluous and irritating for future readers. Remove this part of the comment.

gndrmnn (approved):

Issue #214, opened 2026-04-14
Fix race condition in memcached-pod-kill chaos test that causes flaky failures


berendt:
**Category**: bug | **Scope**: Small

**Description**: The `keystone-memcached-pod-kill` chaos test in `tests/e2e-chaos/memcached-pod-kill/chainsaw-test.yaml` races with Chaos Mesh pod deletion and fails non-deterministically. In step-4 the test uses `wait` with a label selector to confirm the killed pod becomes `Ready=false` before asserting Keystone stays `Ready=True`. chainsaw/kubectl `wait` resolves the label selector once up front, grabs the doomed pod (e.g. `openstack-memcached-75845cbcd5-ht4h9`), and watches that specific object. When Chaos Mesh deletes the pod mid-wait, the watch errors with `Error from server (NotFound): pods "openstack-memcached-75845cbcd5-ht4h9" not found` instead of re-resolving the selector onto the replacement pod the ReplicaSet immediately creates. Observed in run https://github.com/C5C3/forge/actions/runs/24412091069/job/71314792352: the replacement pod `openstack-memcached-75845cbcd5-ptznn` was already `Scheduled`/`Created`/`Started` within the same second (visible in the event dump), the memcached-operator Deployment was healthy, and the Keystone CR ended the test with `Ready=True/AllReady` — confirming this is a test-only race, not an operator regression. The `NOTE` block at lines 50-56 of `chainsaw-test.yaml` already anticipates this race, but the documented mitigation ("pod-kill with gracePeriod:0 plus controller restart takes multiple seconds") does not hold in practice — the Pending→Scheduled→Created transition for the replacement pod happens at the same timestamp as the delete.

**Motivation**: The `keystone-mariadb-pod-kill` and `keystone-openbao-pod-kill` tests in the same suite pass reliably because they assert on Keystone CR sub-conditions (e.g. `DatabaseReady=False`), which chainsaw retries internally until they match. The memcached test cannot use that pattern by design — cache outages are treated as performance degradation only and must not flip any Keystone condition (CC-0047, SC-CHAOS-002). The current workaround (pod-level `Ready=false` wait) introduces a structural race that produces false negatives in CI, blocks the chaos test suite (issue #165), and hides real regressions in the noise.

**Affected Areas**:
- `tests/e2e-chaos/memcached-pod-kill/chainsaw-test.yaml`

**Acceptance Criteria**:
[...]

Issue #165, opened 2026-04-09
Add chaos E2E tests to CI pipeline


berendt:
## Summary

The chaos test framework introduced in CC-0047 (12a8c92) provides Chainsaw-based chaos E2E tests using Chaos Mesh `PodChaos` CRDs (MariaDB, Memcached, OpenBao pod-kill scenarios). These tests are currently only runnable locally via `make e2e-chaos` but are **not integrated into the CI pipeline**.

As more chaos test scenarios are planned, CI integration is needed to ensure operator resilience is continuously validated.

## Current state

- Three chaos scenarios exist under `tests/e2e-chaos/`:
  - `mariadb-pod-kill` (SC-CHAOS-001)
  - `memcached-pod-kill` (SC-CHAOS-002)
  - `openbao-pod-kill` (SC-CHAOS-003)
- A dedicated Chainsaw config exists at `tests/e2e-chaos/chainsaw-config.yaml` with serial execution and extended timeouts (5m assert, 2m cleanup)
- A `make e2e-chaos` target exists in the Makefile
- No reference to chaos tests exists in `.github/workflows/ci.yaml`

## Requirements

- [ ] Add a `e2e-chaos` job to the CI workflow (`.github/workflows/ci.yaml`)
- [ ] The job needs a kind cluster with Chaos Mesh installed (in addition to the standard E2E infrastructure)
- [ ] Wire up change detection so chaos tests run when `tests/e2e-chaos/**` or operator/shared code changes
- [ ] Chaos tests must run **after** the regular E2E tests pass (they depend on a working baseline)
- [ ] Collect JUnit reports and diagnostics artifacts on failure (as done for other E2E jobs)
- [ ] Consider whether chaos tests should be required to pass for merge or run as informational/non-blocking initially

berendt:
Fixed with https://github.com/C5C3/forge/pull/170.

PR #216, opened 2026-04-14
feat(CC-0076): Fix race condition in memcached-pod-kill chaos test

berendt:
# GitHub Issue #214: Fix race condition in memcached-pod-kill chaos test that causes flaky failures

**Description**: The `keystone-memcached-pod-kill` chaos test in `tests/e2e-chaos/memcached-pod-kill/chainsaw-test.yaml` races with Chaos Mesh pod deletion and fails non-deterministically. In step-4 the test uses `wait` with a label selector to confirm the killed pod becomes `Ready=false` before asserting Keystone stays `Ready=True`. chainsaw/kubectl `wait` resolves the label selector once up front, grabs the doomed pod (e.g. `openstack-memcached-75845cbcd5-ht4h9`), and watches that specific object. When Chaos Mesh deletes the pod mid-wait, the watch errors with `Error from server (NotFound): pods "openstack-memcached-75845cbcd5-ht4h9" not found` instead of re-resolving the selector onto the replacement pod the ReplicaSet immediately creates. Observed in run https://github.com/C5C3/forge/actions/runs/24412091069/job/71314792352: the replacement pod `openstack-memcached-75845cbcd5-ptznn` was already `Scheduled`/`Created`/`Started` within the same second (visible in the event dump), the memcached-operator Deployment was healthy, and the Keystone CR ended the test with `Ready=True/AllReady` — confirming this is a test-only race, not an operator regression. The `NOTE` block at lines 50-56 of `chainsaw-test.yaml` already anticipates this race, but the documented mitigation ("pod-kill with gracePeriod:0 plus controller restart takes multiple seconds") does not hold in practice — the Pending→Scheduled→Created transition for the replacement pod happens at the same timestamp as the delete.

**Motivation**: The `keystone-mariadb-pod-kill` and `keystone-openbao-pod-kill` tests in the same suite pass reliably because they assert on Keystone CR sub-conditions (e.g. `DatabaseReady=False`), which chainsaw retries internally until they match. The memcached test cannot use that pattern by design — cache outages are treated as performance degradation only and must not flip any Keystone condition (CC-0047, SC-CHAOS-002). The current workaround (pod-level `Ready=false` wait) introduces a structural race that produces false negatives in CI, blocks the chaos test suite (issue #165), and hides real regressions in the noise.

**Affected Areas**:
- `tests/e2e-chaos/memcached-pod-kill/chainsaw-test.yaml`

**Acceptance Criteria**:
- [ ] `keystone-memcached-pod-kill` no longer relies on `wait` with `condition: Ready=false` against a label selector that matches the pod being deleted
- [ ] The chaos effect is verified via a mechanism that is resilient to pod replacement, e.g. asserting on `Deployment.status.unavailableReplicas > 0` or on a `Pod` UID/`resourceVersion` change, both of which chainsaw retries
- [ ] Alternatively, the `Ready=false` wait is dropped entirely and the test relies only on the subsequent `Ready=true` wait plus the Keystone `AllReady` assertion — which is sufficient to validate the test's actual goal (memcached failures do not degrade Keystone)
- [ ] The `NOTE` block at lines 50-56 is removed or rewritten to reflect the new approach
- [ ] The test passes on three consecutive CI runs of `.github/workflows/ci.yaml` (e2e-chaos job)
- [ ] No change is required in the memcached-operator repository — this is a pure test fix

**Labels:** bug

---
Source: https://github.com/C5C3/forge/issues/214


berendt:



## INFO (2)

### I-001: Sibling tests hardcode replica counts (pre-existing)
**File**: `chainsaw-test.yaml` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Derive expected values from source of truth instead of hardcoding

**Problem**: The memcached test correctly reads the desired replica count from .spec.replicas, but sibling tests operator-pod-crash and api-pod-kill-pdb still hardcode their replica counts (2 and 3 respectively). This PR improves on the existing pattern but the sibling tests remain inconsistent.

**Code**:
```
N/A — pre-existing issue in sibling test files, not in this PR's diff
```

**Action Required**: Consider backporting the dynamic replica count approach to the sibling tests operator-pod-crash and api-pod-kill-pdb in a follow-up PR.

**Suggested Fix**:
```
In sibling tests, replace hardcoded replica counts with dynamic reads from .spec.replicas, following the pattern established in this PR.
```

**Related**: I-002

---

### I-002: Sibling test timeout budgets are tighter than this test (pre-existing)
**File**: `chainsaw-test.yaml` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Test script timeout must exceed worst-case loop duration

**Problem**: Sibling tests have insufficient timeout margins: operator-pod-crash has a 150s timeout for a 240s worst case, and api-pod-kill-pdb has a 120s timeout for a 120s worst case (zero margin). This PR correctly sizes its timeout at 150s for a 120s worst case (30s margin), setting a good precedent.

**Code**:
```
N/A — pre-existing issue in sibling test files, not in this PR's diff
```

**Action Required**: Consider reviewing and adjusting timeout budgets in sibling tests operator-pod-crash and api-pod-kill-pdb to ensure adequate margin above worst-case loop duration.

**Suggested Fix**:
```
Adjust sibling test timeouts to include at least a 25% margin above worst-case script duration, following the precedent set by this PR.
```

**Related**: I-001

---

## Summary

This is a well-executed bug fix. The PR correctly replaces a fragile Pod-level kubectl wait with Deployment-level readyReplicas polling that is resilient to pod replacement, dynamically reads the desired replica count from the Deployment spec, properly calculates timeout budgets with margin, and updates documentation to match. No issues were found against this PR — the two INFO findings are pre-existing issues in sibling tests noted for future improvement.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 0 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge without changes. The PR is clean, well-structured, and follows established project patterns. The two INFO findings are pre-existing issues in sibling tests and do not affect this PR.

PR #218, opened 2026-04-15
feat(CC-0077): Prune stale immutable ConfigMaps after deployment

berendt:
# GitHub Issue #134: Prune stale immutable ConfigMaps after Deployment rollout

**Description**: reconcileConfig creates new immutable ConfigMaps with content-hash suffixes on every config change but never deletes old ones. Add a pruning step that lists ConfigMaps matching the naming pattern, identifies those no longer referenced by any active Pod or ReplicaSet, and deletes them. Retain the last N (e.g., 3) for rollback safety.

**Motivation**: Over the operator's lifetime, orphaned ConfigMaps accumulate indefinitely. This wastes etcd storage and clutters namespace listings, making debugging harder.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_config.go`
- `internal/common/config/`

**Acceptance Criteria**:
- [ ] After a successful Deployment rollout, unreferenced ConfigMaps older than the current one are deleted
- [ ] The last 3 ConfigMaps are retained for rollback
- [ ] Pruning is idempotent and safe against concurrent reconcile loops
- [ ] Unit tests validate pruning logic

---
Source: https://github.com/C5C3/forge/issues/134


berendt:



## WARNING (3)

### W-001: Pre-existing ConfigMaps invisible to pruning due to missing labels
**File**: `config.go:229` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Verify blast radius of changes to shared internal code

**Problem**: PruneImmutableConfigMaps uses client.MatchingLabels{ConfigBaseLabelKey: baseName} as a server-side selector, but CreateImmutableConfigMap on main does not set this label — it was added in this same PR. ConfigMaps created before this PR is deployed will lack the forge.c5c3.io/config-base label, so the label-based List will never find them. These pre-existing unlabeled ConfigMaps will accumulate forever and never be pruned.

**Code**:
```
if err := c.List(ctx, &allConfigMaps, client.InNamespace(namespace), client.MatchingLabels{ConfigBaseLabelKey: baseName}); err != nil {
```

**Action Required**: Either (a) document this as a known limitation — only ConfigMaps created after upgrade are prunable — and note that old ones will be GC'd when the owner CR is deleted, or (b) fall back to listing all ConfigMaps and filtering by prefix when the label-based list returns zero candidates, or (c) add a migration step that labels existing ConfigMaps on first reconcile.

**Suggested Fix**:
```
Document as a known limitation or implement a migration strategy — requires team decision on which approach to take.
```

---

### W-002: Missing E2E test for ConfigMap pruning

**Problem**: The project has 25 Chainsaw E2E test directories under tests/e2e/keystone/ covering every existing feature. This PR adds a new operational behavior (pruning) but includes no tests/e2e/keystone/config-pruning/ directory. Unit and integration tests are thorough, but an E2E test that deploys a Keystone CR, triggers multiple config changes, and asserts the ConfigMap count after pruning would verify the end-to-end wiring on a real cluster.

**Code**:
```
tests/e2e/keystone/ (25 existing Chainsaw test directories — no config-pruning directory present)
```

**Action Required**: Create tests/e2e/keystone/config-pruning/chainsaw-test.yaml that: (1) deploys a Keystone CR, (2) patches the spec to trigger config changes multiple times, (3) waits for DeploymentReady, (4) asserts the number of ConfigMaps matching the prefix is at most retain + 1.

**Suggested Fix**:
```
Create tests/e2e/keystone/config-pruning/chainsaw-test.yaml covering deploy, multiple config changes, and post-pruning ConfigMap count assertion.
```

**Related**: W-003

---

### W-003: Integration test does not verify which ConfigMaps were retained
**File**: `integration_test.go:179-180` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Assert absence of old state in mutation tests

**Problem**: TestIntegration_PruneImmutableConfigMaps asserts remaining has length 4 and contains currentName, but does not assert that the oldest ConfigMap was deleted. Since CreateImmutableConfigMap is called in a loop with hash-based names, the test cannot distinguish 'retained the 3 newest' from 'retained 3 random ones.'

**Code**:
```
g.Expect(remaining).To(HaveLen(4), "current + 3 retained historical ConfigMaps expected")
g.Expect(remaining).To(ContainElement(currentName), "current ConfigMap must survive pruning")
```

**Action Required**: Record the first ConfigMap name created in the loop and add an assertion that it is absent from the remaining list to verify the oldest was actually pruned.

**Suggested Fix**:
```
g.Expect(remaining).To(HaveLen(4), "current + 3 retained historical ConfigMaps expected")
g.Expect(remaining).To(ContainElement(currentName), "current ConfigMap must survive pruning")
g.Expect(remaining).NotTo(ContainElement(firstCreatedName), "oldest ConfigMap should have been pruned")
```

**Related**: W-002

---

## INFO (1)

### I-001: Delete call missing PropagationPolicy diverges from codebase convention
**File**: `config.go:260` — **Fix**: ASK — **Confidence**: likely — **Pattern**: Check resource deletion for orphan cleanup policy

**Problem**: The codebase has an established pattern of using DeletePropagationBackground for delete operations (in job.go:82 and reconcile_policyvalidation.go:204). The pruning delete at config.go:260 uses a bare c.Delete(ctx, &cm) without specifying propagation policy, diverging from the established convention.

**Code**:
```
if err := client.IgnoreNotFound(c.Delete(ctx, &cm)); err != nil {
```

**Action Required**: No change required — ConfigMaps have no owned children so propagation policy is irrelevant. Only change if project style mandates uniform propagation on all deletes.

**Suggested Fix**:
```
No change needed. ConfigMaps have no owned children, so propagation policy is functionally irrelevant here.
```

---

## Summary

This is a well-implemented PR with clean architecture: pruning logic in the shared library with thorough unit tests, a thin controller wrapper, and comprehensive documentation. Test coverage is excellent with 12 unit tests, 2 integration tests, and 3 controller-level tests. The most significant concern is that pre-existing ConfigMaps created before this PR will never be pruned because they lack the new label used as a server-side selector. The absence of an E2E/Chainsaw test is notable given the project's otherwise complete E2E coverage for every feature.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing the label-based selector limitation (CRITICAL-1) — either document it as a known limitation or implement a migration strategy. The missing E2E test (WARNING-1) and integration test assertion gap (WARNING-2) should ideally be addressed before merge but are not blocking given the strong unit test coverage.

gndrmnn (approved):

Issue #190, opened 2026-04-12
Restrict rotation CronJob ServiceAccount to read-only and let operator handle Secret writes


berendt:
**Category**: security | **Scope**: Medium

**Description**: Refactor the fernet/credential key rotation architecture so that the CronJob ServiceAccount only has read access to the key Secrets. The CronJob should compute the new key set and output it (e.g., to a temporary ConfigMap or annotation), and the operator's reconcile loop should detect the output and perform the actual Secret update using its own privileged ServiceAccount.

**Motivation**: The current rotation CronJob's Role grants get and patch on fernet-keys or credential-keys Secrets. If the CronJob pod is compromised via a supply chain attack on the Keystone image, the attacker can write arbitrary fernet keys enabling token forgery. Separating compute (CronJob) from write (operator) follows least-privilege principles.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_fernet.go`
- `operators/keystone/internal/controller/reconcile_credential.go`

**Acceptance Criteria**:
- [ ] CronJob ServiceAccount has only read access to key Secrets
- [ ] Operator reconciler detects rotation output and performs the Secret update
- [ ] Rotation workflow functions identically from an end-user perspective
- [ ] Unit test verifies CronJob RBAC does not include patch/update verbs on Secrets
- [ ] E2E test verifies key rotation completes successfully with the new architecture

PR #244, opened 2026-04-18
feat(CC-0081): Restrict rotation CronJob ServiceAccount to read-only

berendt:
# GitHub Issue #190: Restrict rotation CronJob ServiceAccount to read-only and let operator handle Secret writes

**Category**: security | **Scope**: Medium

**Description**: Refactor the fernet/credential key rotation architecture so that the CronJob ServiceAccount only has read access to the key Secrets. The CronJob should compute the new key set and output it (e.g., to a temporary ConfigMap or annotation), and the operator's reconcile loop should detect the output and perform the actual Secret update using its own privileged ServiceAccount.

**Motivation**: The current rotation CronJob's Role grants get and patch on fernet-keys or credential-keys Secrets. If the CronJob pod is compromised via a supply chain attack on the Keystone image, the attacker can write arbitrary fernet keys enabling token forgery. Separating compute (CronJob) from write (operator) follows least-privilege principles.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_fernet.go`
- `operators/keystone/internal/controller/reconcile_credential.go`

**Acceptance Criteria**:
- [ ] CronJob ServiceAccount has only read access to key Secrets
- [ ] Operator reconciler detects rotation output and performs the Secret update
- [ ] Rotation workflow functions identically from an end-user perspective
- [ ] Unit test verifies CronJob RBAC does not include patch/update verbs on Secrets
- [ ] E2E test verifies key rotation completes successfully with the new architecture

---
Source: https://github.com/C5C3/forge/issues/190


berendt:
> Reviewed by planwerk-review 342b9d8 with Claude Code



## WARNING (2)

### W-001: CredentialKeysReady/FernetKeysReady condition message stale on apply path
**File**: `operators/keystone/internal/controller/reconcile_credential.go:134` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: On the apply-success early return, the condition is set with Reason 'CredentialKeysAvailable' and a message copy-pasted from the steady-state path at step 8. It does not reflect that a rotation was just applied, and the reconciler short-circuits before re-running steps 4-7. The same pattern exists in reconcile_fernet.go:109. Condition is technically correct but a distinct Reason/message would make kubectl describe clearly reflect the transition.

**Code**:
```
conditions.SetCondition(&keystone.Status.Conditions, metav1.Condition{
    Type:    "CredentialKeysReady",
    Status:  metav1.ConditionTrue,
    Reason:  "CredentialKeysAvailable",
    Message: "Credential keys Secret exists and rotation CronJob is configured",
})
return ctrl.Result{Requeue: true}, nil
```

**Action Required**: Use Reason: 'CredentialKeysRotated' / 'FernetKeysRotated' on the apply-success branch to match the emitted event, and update the message to indicate the rotation was just applied (e.g. 'rotation applied; staging secret cleared').

**Suggested Fix**:
```
conditions.SetCondition(&keystone.Status.Conditions, metav1.Condition{
    Type:    "CredentialKeysReady",
    Status:  metav1.ConditionTrue,
    Reason:  "CredentialKeysRotated",
    Message: "rotation applied; staging secret cleared",
})
return ctrl.Result{Requeue: true}, nil
```

**Related**: I-002

---

### W-002: Documentation overstates credential-migrate ordering guarantee
**File**: `docs/how-to/keystone-key-rotation.md:266-267` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: The docs claim every credential row is re-encrypted under the new primary by the time the operator commits the Secret swap, implying no data loss. This glosses over the ~60s window between credential_migrate completion and the kubelet's in-place Secret projection refresh (CC-0074). During that window, running Keystone pods still have the old credential keyset mounted — DB rows are encrypted under the new primary but pods cannot decrypt them yet. This is a pre-existing pre-CC-0081 property, slightly lengthened by the added reconcile hop, and the docs should call it out explicitly. Same overstatement exists in docs/reference/keystone-reconciler.md:444-447.

**Code**:
```
Because the CronJob already ran `credential_migrate` before PATCHing staging, every credential row in the database is re-encrypted with the new primary by the time the operator commits the Secret swap — no data loss when old keys age out (CC-0036).
```

**Action Required**: Add a sentence noting the ~60s key-rollover window between credential_migrate completion and the Secret projection refresh, and explicitly flag this as a pre-existing property (not a CC-0081 regression). Apply the same clarification to docs/reference/keystone-reconciler.md:444-447.

**Suggested Fix**:
```
Add a follow-up paragraph: 'Note: there is a ~60s window between credential_migrate completion and the kubelet refreshing the in-place Secret projection (CC-0074). During this window, running Keystone pods still have the old keyset mounted and cannot decrypt rows already re-encrypted under the new primary. This is a pre-existing property of the rotation flow, not introduced by CC-0081.'
```

---

## INFO (5)

### I-001: maxKeys = normalized + 1 bound is looser than necessary
**File**: `operators/keystone/internal/controller/reconcile_fernet.go:91` — **Fix**: ASK — **Confidence**: likely

**Problem**: The +1 upper bound accounts for the extra incoming primary key produced by keystone-manage fernet_rotate/credential_rotate. In practice, these commands prune to max_active_keys within a single invocation, so the staging payload should have exactly max_active_keys, not +1. For spec.fernet.maxActiveKeys=3, the validator accepts [3,4] where [3,3] would be tighter. The current bound is harmless but overly permissive. Same pattern exists in reconcile_credential.go:96.

**Code**:
```
The upper bound on keys is normalized max + 1 to account for the extra incoming primary key produced by `keystone-manage credential_rotate`.
```

**Action Required**: Consider tightening the upper bound to exact max_active_keys (instead of +1) to produce a more precise regression guard. Needs discussion because relaxing/tightening rotation validator bounds can mask or expose edge cases depending on how fernet_rotate/credential_rotate behave across Keystone versions.

**Suggested Fix**:
```
Replace the +1 upper bound with the normalized max_active_keys exactly, and update the comment to reflect that fernet_rotate/credential_rotate prune within a single invocation. Verify with the Keystone team that no supported version leaves a transient extra key.
```

---

### I-002: Event message count phrasing drifts from docs
**File**: `operators/keystone/internal/controller/rotation_staging.go:192` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Code emits 'rotation applied from staging secret %s (%d active keys)' while docs/how-to/keystone-key-rotation.md:141 show the example as '(3 keys)'. The phrasing 'active keys' vs 'keys' is inconsistent between code and doc example.

**Code**:
```
"rotation applied from staging secret %s (%d active keys)"
```

**Action Required**: Update the doc example to match the code's 'active keys' phrasing.

**Suggested Fix**:
```
rotation applied from staging secret <ks>-fernet-keys-rotation (3 active keys)
```

**Related**: W-001

---

### I-003: ensureStagingSecret 409-requeue behavior undocumented
**File**: `operators/keystone/internal/controller/rotation_staging.go:69` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: controllerutil.CreateOrUpdate does Get -> mutate -> Update. Between Get and Update, if the CronJob PATCHes staging, the Update's embedded ResourceVersion is stale and the API server returns 409 Conflict. The reconciler requeues, the next Get sees the CronJob's Data, and the mutator preserves it — so the CronJob's payload is never lost. However, this flow produces transient error-requeue log lines during a rotation window. Behavior is correct but not documented.

**Code**:
```
// ensureStagingSecret ... (existing doc comment at rotation_staging.go:69)
```

**Action Required**: Add a sentence to the ensureStagingSecret doc comment explicitly calling out the 409-requeue behavior during rotation windows and noting that the CronJob's payload is preserved by the mutator on the retry.

**Suggested Fix**:
```
Extend the ensureStagingSecret doc comment with a note such as: 'Note: if the CronJob PATCHes the staging Secret between this function's Get and Update, the Update returns 409 Conflict and the reconciler requeues. The next invocation observes the CronJob's Data and the mutator preserves it, so the CronJob payload is never lost — but transient error-requeue log lines are expected during rotation windows.'
```

---

### I-004: Repeated setup across rotation envtest tests
**File**: `operators/keystone/internal/controller/integration_test.go:545` — **Fix**: AUTO-FIX — **Confidence**: uncertain

**Problem**: TestRotationApplyEndToEnd_EnvTest (line 545), TestRotationApplyRejectsMalformedKeys_EnvTest (line 643), and TestRotationApplyReplacesDisjointIndices_EnvTest (line 744) share ~30 lines of setup (namespace, prerequisites, Keystone object, driveFullReconciliation, re-fetch). Duplication is ~90 lines total.

**Code**:
```
TestRotationApplyEndToEnd_EnvTest, TestRotationApplyRejectsMalformedKeys_EnvTest, and TestRotationApplyReplacesDisjointIndices_EnvTest each repeat the namespace/prerequisites/Keystone/driveFullReconciliation/re-fetch block.
```

**Action Required**: Extract a setupRotationEnvTest helper that returns the namespace, Keystone object, and driven-to-steady-state client handle. Replace the duplicated setup in all three tests with a call to the helper.

**Suggested Fix**:
```
Add a setupRotationEnvTest(t) helper in integration_test.go that performs namespace creation, prerequisite objects, Keystone creation, driveFullReconciliation, and the re-fetch, returning the resulting Keystone and namespace. Replace the three repeated blocks with calls to this helper.
```

---

### I-005: RotationCompletedAnnotation lacks //nolint:gosec annotation
**File**: `operators/keystone/internal/controller/rotation_staging.go:33` — **Fix**: AUTO-FIX — **Confidence**: uncertain

**Problem**: StagingSecretLabelKey has '//nolint:gosec // label key, not a credential' to suppress G101. RotationCompletedAnnotation lacks the annotation and does not currently need one (the string 'rotation-completed-at' does not match G101's pattern). However, the asymmetry means a future string change could trigger gosec without the reviewer noticing.

**Code**:
```
RotationCompletedAnnotation = "rotation-completed-at"
```

**Action Required**: For consistency, consider adding '//nolint:gosec // annotation key, not a credential' on RotationCompletedAnnotation so both exported constants have matching guard comments.

**Suggested Fix**:
```
RotationCompletedAnnotation = "rotation-completed-at" //nolint:gosec // annotation key, not a credential
```

---

## Summary

Solid, security-focused PR implementing a split compute/write rotation architecture for Keystone Fernet and credential-encryption keys, eliminating the token-forgery primitive on the rotation CronJob ServiceAccount. RBAC narrowing is correctly verified at unit and E2E levels, full-replacement Update semantics are well-documented and regression-tested, and the operator-side validator provides meaningful defense-in-depth. Findings are limited to a stale condition message on the apply path (M1), a documentation overstatement about the credential-migrate key-rollover window (M2), and five low-priority polish items.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 2 |
| INFO     | 5 |

> [!IMPORTANT]
> **Recommendation**: Recommend approval after addressing M1 (condition Reason/message drift on apply path). M2 (documenting the ~60s kubelet key-rollover window) is strongly suggested but optional. Remaining INFO findings are polish and can be deferred to a follow-up.

operators/keystone/internal/controller/reconcile_credential_test.go
```diff
@@ -632,10 +634,49 @@ func TestCredentialRotateScript_EmbeddedContent(t *testing.T) {
...
+		"sync note must reference CC-0081 so the staging-Secret contract stays mirrored across rotation scripts")
+
+	// Verify the rotation-completed-at annotation is written atomically with the
+	// data patch — it is the single-shot commit marker the operator watches (CC-0081, REQ-008).
+	g.Expect(credentialRotateScript).To(ContainSubstring("forge.c5c3.io/rotation-completed-at"),
+		"script must stamp the rotation-completed-at annotation on the staging Secret")
+	g.Expect(credentialRotateScript).To(ContainSubstring("datetime.datetime.now(datetime.timezone.utc)"),
+		"annotation timestamp must be UTC so the operator can parse it as RFC3339 Zulu")
+	g.Expect(credentialRotateScript).To(ContainSubstring(`isoformat(timespec="seconds")`),
+		"annotation timestamp must use seconds precision to match the RFC3339 contract")
```
gndrmnn: That's too implementation specific. You can remove these checks

operators/keystone/internal/controller/reconcile_credential_test.go
```diff
@@ -632,10 +634,49 @@ func TestCredentialRotateScript_EmbeddedContent(t *testing.T) {
...
+	patchIdx := strings.Index(credentialRotateScript, "python3 << 'PYTHON'")
+
+	g.Expect(rotateIdx).To(BeNumerically(">=", 0), "credential_rotate invocation must be present")
+	g.Expect(migrateIdx).To(BeNumerically(">=", 0), "credential_migrate invocation must be present")
+	g.Expect(patchIdx).To(BeNumerically(">=", 0), "python3 PATCH heredoc must be present")
+
+	g.Expect(rotateIdx).To(BeNumerically("<", migrateIdx),
+		"credential_rotate must run before credential_migrate so the active keyset is rotated first")
+	g.Expect(migrateIdx).To(BeNumerically("<", patchIdx),
+		"credential_migrate must run before the K8s API PATCH — stored credentials must be re-encrypted before keys are published")
```
gndrmnn: Too implementation specific. You can remove these

operators/keystone/internal/controller/reconcile_fernet_test.go
```diff
@@ -625,6 +627,19 @@ func TestFernetRotateScript_EmbeddedContent(t *testing.T) {
...
+		"sync note must reference CC-0081 so the staging-Secret contract stays mirrored across rotation scripts")
+
+	// Verify the rotation-completed-at annotation is written atomically with the
+	// data patch — it is the single-shot commit marker the operator watches (CC-0081, REQ-007).
+	g.Expect(fernetRotateScript).To(ContainSubstring("forge.c5c3.io/rotation-completed-at"),
+		"script must stamp the rotation-completed-at annotation on the staging Secret")
+	g.Expect(fernetRotateScript).To(ContainSubstring("datetime.datetime.now(datetime.timezone.utc)"),
+		"annotation timestamp must be UTC so the operator can parse it as RFC3339 Zulu")
+	g.Expect(fernetRotateScript).To(ContainSubstring(`isoformat(timespec="seconds")`),
+		"annotation timestamp must use seconds precision to match the RFC3339 contract")
```
gndrmnn: Too implementation specific. Remove these

operators/keystone/internal/controller/reconcile_fernet_test.go
```diff
@@ -622,6 +627,17 @@ func TestFernetRotateScript_EmbeddedContent(t *testing.T) {
...
+	g.Expect(fernetRotateScript).To(ContainSubstring("CC-0081"),
+		"sync note must reference CC-0081 so the staging-Secret contract stays mirrored across rotation scripts")
+
+	// Verify the rotation-completed-at annotation is written atomically with the
+	// data patch — it is the single-shot commit marker the operator watches (CC-0081, REQ-007).
+	g.Expect(fernetRotateScript).To(ContainSubstring("forge.c5c3.io/rotation-completed-at"),
+		"script must stamp the rotation-completed-at annotation on the staging Secret")
+	g.Expect(fernetRotateScript).To(ContainSubstring("datetime.datetime.now(datetime.timezone.utc)"),
+		"annotation timestamp must be UTC so the operator can parse it as RFC3339 Zulu")
+
```
gndrmnn: The added tests in reconcile_fernet_test.go Line 630-640 are STILL too implementation specific. We are not going to add unit tests that check for certain substrings in the Python source code. Remove the tests that check for the substring `CC-0081`, remove the check for the substring `forge.c5c3.io/rotation-completed-at` as well as `datetime.datetime` etc.

The same applies for reconcile_credential_test.go Lines 637-647

Checking the source code for the occurrence of substrings is very bad practice and very brittle. You can keep the check for `python3 << 'PYTHON'` as that will probably not change soon and validates that our script has been successfully deployed. But the others must go.

gndrmnn (changes requested):

gndrmnn (approved):

Issue #249, opened 2026-04-20
Grant ESO write capabilities on Keystone KV paths so fernet/credential PushSecrets succeed


berendt:
**Category**: bug | **Scope**: Small

**Description**: The Keystone operator creates `PushSecret` CRs to back up the `*-fernet-keys` and `*-credential-keys` Secrets to OpenBao under `kv-v2/openstack/keystone/fernet-keys` and `.../credential-keys` (`operators/keystone/internal/controller/reconcile_fernet.go:332-358`, `operators/keystone/internal/controller/reconcile_credential.go:323-349`). Both `PushSecret` CRs reference `ClusterSecretStore openbao-cluster-store` (`deploy/eso/clustersecretstore.yaml:17`), which authenticates via the `eso-management` Kubernetes auth role (`deploy/openbao/bootstrap/setup-auth.sh:85-91`, `token_policies=eso-management`). The bound policy `deploy/openbao/policies/eso-management.hcl:27-29` grants ONLY `read` on `kv-v2/data/openstack/keystone/*`, so every push fails with a 403 and the KV paths for fernet/credential keys remain empty. The only Keystone-related material actually present in OpenBao is what the bootstrap script writes directly with the root token (`deploy/openbao/bootstrap/write-bootstrap-secrets.sh:84-92`): `bootstrap/keystone-admin`, `infrastructure/mariadb`, `openstack/keystone/db`. Introduce a dedicated push-scope policy (mirroring `deploy/openbao/policies/push-app-credentials.hcl` and `push-ceph-keys.hcl`, which use `capabilities = [\"create\", \"update\", \"read\"]`) and bind it additionally to the `eso-management` role so the read-vs-write separation of concerns is preserved.

**Motivation**: The `PushSecret` backup is the only mechanism that makes rotated fernet and credential keys recoverable from OpenBao. Without it, key material lives exclusively in the in-cluster Kubernetes Secret — a cluster rebuild, operator replay, or accidental Secret deletion permanently loses every fernet/credential key ever generated or rotated, invalidating all issued Fernet tokens and rendering every existing application credential in the DB undecryptable. The operator already builds the correct `PushSecret` CRs and the target KV paths are the right ones; the gap is purely on the OpenBao authorization side. Keeping the read-only `eso-management` policy untouched and adding a separate `push-keystone-keys` policy mirrors the pattern already established for Ceph keys and application credentials, and keeps least-privilege intact (the push role cannot read unrelated paths, the management role cannot write).

[...]

Issue #193, opened 2026-04-12
Add finalizer for OpenBao PushSecret cleanup on CR deletion


berendt:
**Category**: improvement | **Scope**: Medium

**Description**: Implement a finalizer on the Keystone CR that cleans up PushSecret data written to OpenBao KV-v2 when the CR is deleted. While owned Kubernetes resources (Deployments, Services, Jobs) are garbage-collected via ownerReferences, OpenBao secrets persist indefinitely after CR deletion. The finalizer should delete the relevant KV-v2 paths and then remove itself.

**Motivation**: Deleting a Keystone CR leaves stale secrets in OpenBao including database credentials, fernet keys, and credential keys. These orphaned secrets are a security risk (stale credentials that should no longer exist) and an operational burden (manual cleanup required). This is the standard Kubernetes pattern for external resource lifecycle management.

**Affected Areas**:
- `operators/keystone/internal/controller/keystone_controller.go`
- `operators/keystone/internal/controller/reconcile_secrets.go`

**Acceptance Criteria**:
- [ ] Finalizer added to Keystone CR on first reconcile
- [ ] On CR deletion, PushSecret-managed paths in OpenBao are cleaned up
- [ ] Finalizer is removed after successful cleanup
- [ ] If OpenBao is unavailable during deletion, the finalizer blocks deletion with a clear status message
- [ ] E2E test verifies OpenBao paths are empty after CR deletion

PR #250, opened 2026-04-20
feat(CC-0083): Grant ESO write access on Keystone KV paths

berendt:
# GitHub Issue #249: Grant ESO write capabilities on Keystone KV paths so fernet/credential PushSecrets succeed

**Description**: The Keystone operator creates `PushSecret` CRs to back up the `*-fernet-keys` and `*-credential-keys` Secrets to OpenBao under `kv-v2/openstack/keystone/fernet-keys` and `.../credential-keys` (`operators/keystone/internal/controller/reconcile_fernet.go:332-358`, `operators/keystone/internal/controller/reconcile_credential.go:323-349`). Both `PushSecret` CRs reference `ClusterSecretStore openbao-cluster-store` (`deploy/eso/clustersecretstore.yaml:17`), which authenticates via the `eso-management` Kubernetes auth role (`deploy/openbao/bootstrap/setup-auth.sh:85-91`, `token_policies=eso-management`). The bound policy `deploy/openbao/policies/eso-management.hcl:27-29` grants ONLY `read` on `kv-v2/data/openstack/keystone/*`, so every push fails with a 403 and the KV paths for fernet/credential keys remain empty. The only Keystone-related material actually present in OpenBao is what the bootstrap script writes directly with the root token (`deploy/openbao/bootstrap/write-bootstrap-secrets.sh:84-92`): `bootstrap/keystone-admin`, `infrastructure/mariadb`, `openstack/keystone/db`. Introduce a dedicated push-scope policy (mirroring `deploy/openbao/policies/push-app-credentials.hcl` and `push-ceph-keys.hcl`, which use `capabilities = [\"create\", \"update\", \"read\"]`) and bind it additionally to the `eso-management` role so the read-vs-write separation of concerns is preserved.

**Motivation**: The `PushSecret` backup is the only mechanism that makes rotated fernet and credential keys recoverable from OpenBao. Without it, key material lives exclusively in the in-cluster Kubernetes Secret — a cluster rebuild, operator replay, or accidental Secret deletion permanently loses every fernet/credential key ever generated or rotated, invalidating all issued Fernet tokens and rendering every existing application credential in the DB undecryptable. The operator already builds the correct `PushSecret` CRs and the target KV paths are the right ones; the gap is purely on the OpenBao authorization side. Keeping the read-only `eso-management` policy untouched and adding a separate `push-keystone-keys` policy mirrors the pattern already established for Ceph keys and application credentials, and keeps least-privilege intact (the push role cannot read unrelated paths, the management role cannot write).

**Affected Areas**:
- `deploy/openbao/policies/push-keystone-keys.hcl` (new file — `create`, `update`, `read` on `kv-v2/data/openstack/keystone/fernet-keys` and `.../credential-keys`)
- `deploy/openbao/bootstrap/setup-auth.sh` (bind additional policy to the `eso-management` role: `token_policies=eso-management,push-keystone-keys`)
- `architecture/docs/09-implementation/09-openbao-deployment.md` (document the new policy alongside the existing `eso-management` / `push-*` policies)

**Acceptance Criteria**:
- [ ] New policy file `deploy/openbao/policies/push-keystone-keys.hcl` exists with two path blocks (`kv-v2/data/openstack/keystone/fernet-keys`, `kv-v2/data/openstack/keystone/credential-keys`) each granting `capabilities = [\"create\", \"update\", \"read\"]`; no wildcard path, no `delete`, no `list`
- [ ] `deploy/openbao/bootstrap/setup-auth.sh` binds `token_policies=eso-management,push-keystone-keys` on the management cluster's `eso-management` role; other cluster roles (`control-plane`, `hypervisor`, `storage`) are unchanged
- [ ] `deploy/openbao/bootstrap/setup-policies.sh` picks up the new `.hcl` automatically (no change needed; verified via `ls deploy/openbao/policies/*.hcl`)
- [ ] `deploy/openbao/policies/eso-management.hcl` stays read-only — the existing DEVIATION comment (lines 19-26) is updated to point at the new push policy for write access
- [ ] On a fresh kind cluster, after `make deploy-infra` plus a Keystone CR reconcile, `bao kv get kv-v2/openstack/keystone/fernet-keys` and `.../credential-keys` both return the current key set (not \"no value found\")
- [ ] `kubectl -n openstack describe pushsecret <keystone>-fernet-keys-backup` reports `SecretSynced=True`; the same for `<keystone>-credential-keys-backup`
- [ ] New Chainsaw e2e step (extending `tests/e2e/keystone/basic-deployment/` or a dedicated `openbao-pushsecret-sync` test) asserts both `PushSecret` CRs reach `Ready=True` within the existing timeout budget

**Non-Goals**:
- Changing the `PushSecret` target paths or the `ClusterSecretStore` wiring. The push targets (`kv-v2/data/openstack/keystone/{fernet,credential}-keys`) are correct and already match the naming used by the operator.
- Widening the `eso-management` policy to include write verbs. The read/push separation is intentional and matches the pattern of `push-app-credentials` vs `eso-*` policies.
- Cleaning up orphaned OpenBao paths after CR deletion — that is already tracked by #193 (finalizer for OpenBao PushSecret cleanup).
- Re-scoping what the rotation CronJob is allowed to write — covered by #190 (restrict rotation CronJob SA to read-only).

**Labels:** bug

**Labels:** bug

---
Source: https://github.com/C5C3/forge/issues/249


berendt:



## WARNING (1)

### W-001: Diagnostic catch block omits OpenBao audit/server logs
**File**: `tests/e2e/keystone/basic-deployment/chainsaw-test.yaml:197-203` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: A 403 on PushSecret is most authoritatively diagnosed from OpenBao's audit log (which records the denied path + policy) rather than the ESO controller log (which only reports the HTTP status). Without OpenBao logs in the catch block, operators need a second CI run with more logging to diagnose failures.

**Code**:
```
catch:
  - script:
      content: |
        echo "=== PushSecret status ==="
        kubectl -n keystone get pushsecrets -o yaml 2>&1 || true
```

**Action Required**: Add a kubectl logs command for OpenBao to the chainsaw catch script so 403 failures can be diagnosed in a single CI run.

**Suggested Fix**:
```
echo "=== openbao logs (last 100 lines) ==="
kubectl -n openbao logs -l app.kubernetes.io/name=openbao --tail=100 2>&1 || true
```

---

## INFO (7)

### I-001: Redundant `read` capability in push-keystone-keys policy
**File**: `deploy/openbao/policies/push-keystone-keys.hcl:23-27` — **Fix**: ASK — **Confidence**: likely

**Problem**: The `read` capability in push-keystone-keys.hcl is already granted to the same role via the wildcard `kv-v2/data/openstack/keystone/*` in eso-management.hcl:34-36 (both policies are bound together on the management role at setup-auth.sh:89). The `read` capability adds nothing on the management cluster. However, it matches the convention in push-app-credentials.hcl and keeps the policy self-contained/portable.

**Code**:
```
path "kv-v2/data/openstack/keystone/fernet-keys" {
  capabilities = ["create", "update", "read"]
}

path "kv-v2/data/openstack/keystone/credential-keys" {
  capabilities = ["create", "update", "read"]
}
```

**Action Required**: Recommend leaving as-is but adding a note to the header comment explaining that `read` is retained for policy-portability rather than being load-bearing at the current binding.

**Suggested Fix**:
```
Add a comment in the policy header clarifying that `read` is retained for policy-portability and is already covered by eso-management's wildcard read at the current binding.
```

---

### I-002: Loop-scoped variable reuse without explicit reset
**File**: `deploy/openbao/bootstrap/setup-auth.sh:88-92` — **Fix**: AUTO-FIX — **Confidence**: uncertain

**Problem**: `token_policies` is assigned inside the loop but not explicitly reset each iteration. With `set -u` and four clusters the current code is safe because every branch of the `if` assigns. However, the pattern is fragile: a future developer adding a third branch that forgets to assign would inherit the prior iteration's value.

**Code**:
```
if [[ "${cluster}" == "management" ]]; then
  token_policies="eso-${cluster},push-keystone-keys"
else
  token_policies="eso-${cluster}"
fi
```

**Action Required**: Refactor to initialize token_policies with a default value at the top of each iteration, then append cluster-specific extras conditionally.

**Suggested Fix**:
```
local token_policies="eso-${cluster}"
if [[ "${cluster}" == "management" ]]; then
  token_policies+=",push-keystone-keys"
fi
```

**Related**: I-003

---

### I-003: Scaling the per-cluster policy fan-out
**File**: `deploy/openbao/bootstrap/setup-auth.sh:88-92` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: If a future cluster needs its own extra write policy (e.g. a push-cinder-backups on storage), the if/elif chain grows. Before the pattern ossifies across more clusters, a declarative map at file scope would keep the loop body pure.

**Code**:
```
if [[ "${cluster}" == "management" ]]; then
  token_policies="eso-${cluster},push-keystone-keys"
else
  token_policies="eso-${cluster}"
fi
```

**Action Required**: Consider introducing a declarative associative array mapping cluster names to extra policies, consumed in the loop — not needed now, but worth keeping in mind before the pattern spreads.

**Suggested Fix**:
```
Introduce a top-level `declare -A EXTRA_POLICIES=([management]="push-keystone-keys")` and in the loop use `extra="${EXTRA_POLICIES[${cluster}]:-}"; token_policies="eso-${cluster}${extra:+,${extra}}"` to avoid the if/elif chain growing with new clusters.
```

**Related**: I-002

---

### I-004: Policy paths correctly match operator-emitted RemoteKey
**File**: `deploy/openbao/policies/push-keystone-keys.hcl:23-27` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Positive observation: the operator sets RemoteKey to `kv-v2/data/openstack/keystone/fernet-keys` (literal, includes `data/`) at reconcile_fernet.go:352 and reconcile_credential.go:343. The new policy paths match exactly and are consistent with the existing convention in push-app-credentials.hcl and push-ceph-keys.hcl.

**Code**:
```
path "kv-v2/data/openstack/keystone/fernet-keys" {
  capabilities = ["create", "update", "read"]
}
path "kv-v2/data/openstack/keystone/credential-keys" {
  capabilities = ["create", "update", "read"]
}
```

**Action Required**: No action required — internally consistent with how the repo treats the KV mount.

**Suggested Fix**:
```
No action required.
```

---

### I-005: Bidirectional documentation links between policy files
**File**: `deploy/openbao/policies/eso-management.hcl:28-33` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Positive observation: eso-management.hcl points forward to push-keystone-keys.hcl, and the new file points back with a note about being bound additionally alongside eso-management. This ensures future readers grepping for 'keystone write access' will not miss either half.

**Code**:
```
# eso-management.hcl references push-keystone-keys.hcl, and push-keystone-keys.hcl references eso-management.hcl
```

**Action Required**: No action required — documentation cross-linking is good.

**Suggested Fix**:
```
No action required.
```

---

### I-006: Step renumbering in chainsaw test is complete
**File**: `tests/e2e/keystone/basic-deployment/chainsaw-test.yaml:204-209` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Positive observation: both comment headers (Step 6, Step 7) were updated consistently; no stale 'Step 5' references remain after the new assertion step was inserted.

**Code**:
```
# Step 6: ...
# Step 7: ...
```

**Action Required**: No action required.

**Suggested Fix**:
```
No action required.
```

---

### I-007: Consolidated PushSecret assertion with shared catch is appropriate
**File**: `tests/e2e/keystone/basic-deployment/chainsaw-test.yaml:172-175` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Positive observation: the two PushSecrets are driven by the same policy, so any regression affects both — one diagnostic block is the economical choice, and the inline comment explains this reasoning.

**Code**:
```
# inline comment explaining consolidated assertion
```

**Action Required**: No action required.

**Suggested Fix**:
```
No action required.
```

**Related**: W-001

---

## Summary

Tight, well-scoped fix that grants ESO write access on two literal Keystone KV paths while preserving the read-only invariant on eso-management. Security reasoning is strong: literal paths avoid the wildcard-scoping pitfall flagged in prior CC-0009 reviews, and the threat-model comment about MariaDB credentials under the shared openstack/keystone/ prefix is embedded directly in the policy. The e2e assertion targets the observable outcome (Ready=True on both PushSecrets) at the correct level of abstraction, and documentation is cross-linked between policy files. Two minor nice-to-haves: adding OpenBao logs to the catch block for faster 403 diagnosis, and hardening the per-cluster policy loop against future branch-addition mistakes.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 1 |
| INFO     | 7 |

> [!IMPORTANT]
> **Recommendation**: Approve. No blockers; only two small nice-to-haves (add OpenBao logs to the chainsaw catch block; harden the setup-auth.sh loop against future branch-addition mistakes).

Issue #247, opened 2026-04-20
Harden graceful pod termination during Keystone rolling updates


berendt:
**Category**: improvement | **Scope**: Medium

**Description**: `buildKeystoneDeployment` (operators/keystone/internal/controller/reconcile_deployment.go:138-186) relies on a hardcoded `terminationGracePeriodSeconds: 30` paired with a 5 s `preStop` sleep (`/bin/sh -c "sleep 5"`) and uWSGI's `--master` signal handling to drain in-flight requests during rolling updates (e.g. `kubectl patch keystone ... spec.image.tag`). This baseline works for the common case but leaves four production gaps: (1) the 30 s grace period is not user-configurable, leaving only ~25 s for uWSGI to finish in-flight requests — insufficient for workloads with slow LDAP/DB token validation; (2) `uwsgiCommand` (reconcile_deployment.go:278-306) does not set `--harakiri`, so a single hung request can block graceful shutdown until SIGKILL; (3) `--http-keepalive` defaults to `true` (reconcile_deployment.go:281-295) with no keep-alive timeout, letting clients reuse a connection to a pod that has already been removed from the Service endpoints; (4) `Deployment.Spec.Strategy` is never set, so the Kubernetes default `RollingUpdate` with `maxUnavailable: 25%` / `maxSurge: 25%` applies — combined with the `minAvailable: 1` PDB (reconcile_deployment.go:248-271) this permits brief capacity dips rather than zero-downtime rollouts. Expose `spec.terminationGracePeriodSeconds`, `spec.preStopSleepSeconds`, `spec.uwsgi.harakiri`, and `spec.uwsgi.httpKeepAliveTimeout` as optional CRD fields; set an explicit `RollingUpdate` strategy with `maxUnavailable: 0`, `maxSurge: 1` as the new default and make it overridable via `spec.strategy`.

[...]

PR #252, opened 2026-04-21
feat(CC-0084): Harden graceful pod termination in Keystone rolling updates

berendt:
# GitHub Issue #247: Harden graceful pod termination during Keystone rolling updates

**Description**: `buildKeystoneDeployment` (operators/keystone/internal/controller/reconcile_deployment.go:138-186) relies on a hardcoded `terminationGracePeriodSeconds: 30` paired with a 5 s `preStop` sleep (`/bin/sh -c "sleep 5"`) and uWSGI's `--master` signal handling to drain in-flight requests during rolling updates (e.g. `kubectl patch keystone ... spec.image.tag`). This baseline works for the common case but leaves four production gaps: (1) the 30 s grace period is not user-configurable, leaving only ~25 s for uWSGI to finish in-flight requests — insufficient for workloads with slow LDAP/DB token validation; (2) `uwsgiCommand` (reconcile_deployment.go:278-306) does not set `--harakiri`, so a single hung request can block graceful shutdown until SIGKILL; (3) `--http-keepalive` defaults to `true` (reconcile_deployment.go:281-295) with no keep-alive timeout, letting clients reuse a connection to a pod that has already been removed from the Service endpoints; (4) `Deployment.Spec.Strategy` is never set, so the Kubernetes default `RollingUpdate` with `maxUnavailable: 25%` / `maxSurge: 25%` applies — combined with the `minAvailable: 1` PDB (reconcile_deployment.go:248-271) this permits brief capacity dips rather than zero-downtime rollouts. Expose `spec.terminationGracePeriodSeconds`, `spec.preStopSleepSeconds`, `spec.uwsgi.harakiri`, and `spec.uwsgi.httpKeepAliveTimeout` as optional CRD fields; set an explicit `RollingUpdate` strategy with `maxUnavailable: 0`, `maxSurge: 1` as the new default and make it overridable via `spec.strategy`.

**Motivation**: A `kubectl patch keystone ... spec.image.tag` triggers a pod rotation that must deliver both (a) no new requests routed to a terminating pod and (b) in-flight requests completing cleanly. The current implementation mitigates (a) via the 5 s preStop sleep that covers EndpointSlice propagation, but keep-alive connections bypass this protection because clients keep reusing an already-stale socket. (b) is limited by a fixed 25 s drain window without per-worker hang detection — a single stuck token validation blocks shutdown until SIGKILL truncates every other in-flight request on the same pod. Operators cannot tune any of these values today without forking the operator. At the same time, the default `maxUnavailable: 25%` reduces Keystone capacity during rollout — Keystone is the authentication gate for the entire control plane, so any capacity dip amplifies user-visible latency across all OpenStack services.

**Affected Areas**:
- `operators/keystone/api/v1alpha1/keystone_types.go`
- `operators/keystone/api/v1alpha1/keystone_webhook.go`
- `operators/keystone/helm/keystone-operator/crds/keystone.openstack.c5c3.io_keystones.yaml`
- `operators/keystone/config/crd/bases/keystone.openstack.c5c3.io_keystones.yaml`
- `architecture/docs/04-architecture/04-high-availability.md`

**Acceptance Criteria**:
- [ ] `KeystoneSpec` gains optional `terminationGracePeriodSeconds *int64` (default 30, min 10) and `preStopSleepSeconds *int64` (default 5, min 0)
- [ ] `UWSGISpec` gains optional `harakiri *int32` (worker request timeout in seconds, default unset) and `httpKeepAliveTimeout *int32` (idle keep-alive timeout in seconds, default unset)
- [ ] `buildKeystoneDeployment` sets `.Spec.Strategy` to `RollingUpdate` with `maxUnavailable: 0` and `maxSurge: 1` by default; new optional `spec.strategy *appsv1.DeploymentStrategy` field allows full override
- [ ] `preStop` sleep duration is derived from `spec.preStopSleepSeconds` instead of the hardcoded `sleep 5`
- [ ] `uwsgiCommand` appends `--harakiri <n>` when set and `--http-keepalive-timeout <n>` when set
- [ ] Validating webhook rejects `preStopSleepSeconds >= terminationGracePeriodSeconds` and rejects `harakiri >= (terminationGracePeriodSeconds - preStopSleepSeconds)` so a worker timeout cannot exceed the remaining drain window
- [ ] Unit tests in `reconcile_deployment_test.go` cover: defaults (grace=30, preStop=5, maxUnavailable=0, maxSurge=1, no harakiri, no keep-alive timeout), all four fields overridden, strategy override replaces defaults, webhook validation paths (preStop >= grace rejected, harakiri too large rejected)
- [ ] E2E test verifies a rolling image-tag patch produces zero failed requests under a steady load generator running against the Service
- [ ] CRD manifests in both `helm/.../crds/` and `config/crd/bases/` are regenerated and committed
- [ ] `architecture/docs/04-architecture/04-high-availability.md` documents the rollout flow (EndpointSlice removal ‖ preStop sleep → SIGTERM → uWSGI graceful shutdown with harakiri fallback → SIGKILL) and the new tunables, including recommended production values and the keep-alive-timeout interaction with the preStop sleep

**Labels:** enhancement

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/247


berendt:



## WARNING (3)

### W-001: Missing aggregate-validation test coverage for CC-0084 paths
**File**: `operators/keystone/api/v1alpha1/keystone_webhook_test.go:1109-1188` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Update aggregate validation tests when adding new validation paths

**Problem**: The PR adds 8 new validation paths (terminationGracePeriod minimum, preStopSleep minimum, preStop ≥ grace cross-field, harakiri minimum, harakiri ≥ drain-window, keepAliveTimeout minimum, keepAliveTimeout requires keepAlive, strategy Recreate+rollingUpdate) but none of them are exercised in TestValidateCreate_RunsAllValidations, which is explicitly designed to prove every validation rule fires simultaneously (no short-circuit). A future regression that short-circuits the webhook on, say, the first replica error will silently skip all the new CC-0084 checks and no test catches it.

**Code**:
```
func TestValidateCreate_RunsAllValidations(t *testing.T) {
    ...
    // REQ-005 (CC-0075): Break TSC — wrong label selectors.
    k.Spec.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{...}

    _, err := w.ValidateCreate(context.Background(), k)
    g.Expect(err).To(HaveOccurred())
    errMsg := err.Error()
    g.Expect(errMsg).To(ContainSubstring("replicas"))
    ...
    g.Expect(errMsg).To(ContainSubstring("topologySpreadConstraints"))
}
```

**Action Required**: Extend TestValidateCreate_RunsAllValidations to break all four new axes (graceful termination, uWSGI termination, keep-alive, strategy) and add substring assertions for terminationGracePeriodSeconds, preStopSleepSeconds, harakiri, httpKeepAliveTimeout, and strategy.

**Suggested Fix**:
```
    // REQ-001/REQ-002/REQ-007 (CC-0084): Break graceful termination cross-field.
    k.Spec.TerminationGracePeriodSeconds = ptrInt64(5)
    k.Spec.PreStopSleepSeconds = ptrInt64(30)
    // REQ-003/REQ-008/REQ-012 (CC-0084): Break uWSGI termination fields.
    k.Spec.UWSGI.Harakiri = ptrInt32(0)
    k.Spec.UWSGI.HTTPKeepAliveTimeout = ptrInt32(5)
    k.Spec.UWSGI.HTTPKeepAlive = false
    // REQ-006 (CC-0084): Break strategy — Recreate with rollingUpdate block.
    mu := intstr.FromInt32(0); ms := intstr.FromInt32(1)
    k.Spec.Strategy = &appsv1.DeploymentStrategy{
        Type: appsv1.RecreateDeploymentStrategyType,
        RollingUpdate: &appsv1.RollingUpdateDeployment{MaxUnavailable: &mu, MaxSurge: &ms},
    }

    g.Expect(errMsg).To(ContainSubstring("terminationGracePeriodSeconds"))
    g.Expect(errMsg).To(ContainSubstring("preStopSleepSeconds"))
    g.Expect(errMsg).To(ContainSubstring("harakiri"))
    g.Expect(errMsg).To(ContainSubstring("httpKeepAliveTimeout"))
    g.Expect(errMsg).To(ContainSubstring("strategy"))
```

**Related**: W-002

---

### W-002: Local ptrInt32/ptrInt64 helpers duplicate k8s.io/utils/ptr.To
**File**: `operators/keystone/api/v1alpha1/keystone_webhook_test.go:1666-1669` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Two new helpers ptrInt64 and ptrInt32 are introduced and called 27 times across the new CC-0084 tests. The project already uses ptr.To from k8s.io/utils/ptr for exactly these pointers — the new integration test at operators/keystone/internal/controller/integration_test.go:1998 uses ptr.To(int64(60)) for TerminationGracePeriodSeconds. This duplicates existing functionality and matches prior review feedback from CC-0013 to use k8s.io/utils/ptr instead of reinventing pointer helpers.

**Code**:
```
func ptrInt64(v int64) *int64 { return &v }

func ptrInt32(v int32) *int32 { return &v }
```

**Action Required**: Delete the local ptrInt64/ptrInt32 helpers, import k8s.io/utils/ptr, and replace the 27 call sites with ptr.To(int64(v)) and ptr.To(int32(v)).

**Related**: W-001

---

### W-003: Docstring on deploymentStrategy misstates aliasing behavior
**File**: `operators/keystone/internal/controller/reconcile_deployment.go:410-427` — **Fix**: ASK — **Confidence**: verified

**Problem**: The docstring claims 'A fresh value is constructed on every call (including the *IntOrString pointers inside RollingUpdate) so the result is safe to reuse across reconciles without aliasing.' This is only true for the default branch. The override path returns *keystone.Spec.Strategy, which is a shallow struct copy — the returned value's RollingUpdate *RollingUpdateDeployment is the same pointer as the CR's, and so are the MaxUnavailable/MaxSurge pointers nested inside. Code written against this comment that mutates deploy.Spec.Strategy.RollingUpdate.MaxSurge would also mutate the CR. Today's downstream EnsureDeployment flow doesn't mutate through the pointer, so this is a latent hazard rather than an active bug.

**Code**:
```
// A fresh value is constructed on every call (including the *IntOrString
// pointers inside RollingUpdate) so the result is safe to reuse across
// reconciles without aliasing.
func deploymentStrategy(keystone *keystonev1alpha1.Keystone) appsv1.DeploymentStrategy {
    if keystone.Spec.Strategy != nil {
        return *keystone.Spec.Strategy
    }
    ...
}
```

**Action Required**: Either correct the docstring to accurately describe the override path's shallow copy behavior, or DeepCopy the spec.Strategy on the override path so the comment's guarantee actually holds. Decision between the two depends on whether the team values the stronger invariant.

**Suggested Fix**:
```
// deploymentStrategy returns the effective Deployment rollout strategy for the
// Keystone API. If spec.Strategy is non-nil, a deep copy is returned so the
// Deployment spec never shares pointer chains with the CR. Otherwise, the
// hardened default is returned: RollingUpdate with maxUnavailable=0 and
// maxSurge=1, which guarantees no capacity loss during rollouts
// (CC-0084, REQ-005, REQ-006).
func deploymentStrategy(keystone *keystonev1alpha1.Keystone) appsv1.DeploymentStrategy {
	if keystone.Spec.Strategy != nil {
		return *keystone.Spec.Strategy.DeepCopy()
	}
	maxUnavailable := intstr.FromInt32(0)
	maxSurge := intstr.FromInt32(1)
	return appsv1.DeploymentStrategy{
		Type: appsv1.RollingUpdateDeploymentStrategyType,
		RollingUpdate: &appsv1.RollingUpdateDeployment{
			MaxUnavailable: &maxUnavailable,
			MaxSurge:       &maxSurge,
		},
	}
}
```

---

## INFO (2)

### I-001: Load-generator Job name collision on rapid re-runs
**File**: `tests/e2e/keystone/rolling-update-zero-downtime/chainsaw-test.yaml:119-196` — **Fix**: AUTO-FIX — **Confidence**: likely — **Pattern**: Cleanup Chainsaw-scripted Jobs with --ignore-not-found

**Problem**: The chainsaw test applies a Job with a fixed name via kubectl apply -f - inside a script: step. Because the Job is created by a script (not a chainsaw declarative resource), chainsaw will not track it for cleanup. ttlSecondsAfterFinished: 300 handles cleanup in ≤5 minutes, but rapid CI retries (or parallel runs that land on the same deadline) can race with the TTL and cause kubectl apply to fail because Job spec fields are immutable after creation. This matches the project pattern surfaced in CC-0048 for cleaning up chainsaw-scripted Jobs.

**Code**:
```
          kubectl apply -n openstack -f - <<'JOB'
```

**Action Required**: Add a pre-create 'kubectl delete job ... --ignore-not-found --wait=true' before the kubectl apply to make the script idempotent on re-entry.

**Suggested Fix**:
```
          kubectl delete job keystone-rolling-update-load-gen -n openstack --ignore-not-found --wait=true

          kubectl apply -n openstack -f - <<'JOB'
```

---

### I-002: Submodule content for 04-high-availability.md not verifiable in this review
**File**: `architecture` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The architecture submodule pointer moves from 7beebd6 to 7cfe7d2. The PR's acceptance criteria require documenting the rollout flow and new tunables in architecture/docs/04-architecture/04-high-availability.md, but the submodule is not checked out in the review worktree so the content cannot be verified. Additionally, commit a3a88bb (fix(CC-0084): repair docs build) deliberately removed a cross-repo link to this file from docs/reference/keystone-crd.md because VitePress couldn't resolve it — users now have no on-site pointer from the CRD reference to the HA architecture doc; the only discoverability path is the submodule's own site.

**Code**:
```
Subproject commit 7beebd6...7cfe7d2 (architecture submodule)
```

**Action Required**: A reviewer with the architecture repo cloned should verify commit 7cfe7d2 actually adds the required rollout-flow and tunable-recommendations sections. Separately, consider how users will discover the HA doc from the CRD reference now that the cross-repo link was removed.

**Suggested Fix**:
```
Verify the submodule commit content externally and restore a discoverability path (e.g., a VitePress-resolvable link or an inline summary in docs/reference/keystone-crd.md) from the CRD reference to the HA architecture documentation.
```

---

## Summary

The implementation is well-structured: CRD fields, webhook validation, reconciler wiring, unit/integration/E2E tests, regenerated CRDs in both locations, and documentation all ship together and track the acceptance criteria line-by-line. The validation layer is thoughtful — defaults resolve consistently between webhook and reconciler (30/5), cross-field math treats preStopSleep ≥ grace and harakiri ≥ drainWindow symmetrically, and the E2E Chainsaw test runs a real in-cluster load generator asserting 0 × 5xx/resets and availableReplicas ≥ 3 throughout. The main gap is the aggregate-validation test not being extended for the 8 new CC-0084 paths, plus smaller items around pointer-helper duplication, a misleading docstring, and a Job-name race on rapid CI retries.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing Findings 1 and 2 (quick mechanical fixes that align with established project patterns). Findings 3 and 4 are latent hazards worth hardening before merge but not blocking. Finding 5 requires external verification of the architecture submodule content by a reviewer with that repo cloned.

PR #260, opened 2026-04-22
feat(CC-0079): Add finalizer for OpenBao PushSecret cleanup

berendt:
# GitHub Issue #193: Add finalizer for OpenBao PushSecret cleanup on CR deletion

**Description**: Implement a finalizer on the Keystone CR that cleans up PushSecret data written to OpenBao KV-v2 when the CR is deleted. While owned Kubernetes resources (Deployments, Services, Jobs) are garbage-collected via ownerReferences, OpenBao secrets persist indefinitely after CR deletion. The finalizer should delete the relevant KV-v2 paths and then remove itself.

**Motivation**: Deleting a Keystone CR leaves stale secrets in OpenBao including database credentials, fernet keys, and credential keys. These orphaned secrets are a security risk (stale credentials that should no longer exist) and an operational burden (manual cleanup required). This is the standard Kubernetes pattern for external resource lifecycle management.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_secrets.go`

**Acceptance Criteria**:
- [ ] Finalizer added to Keystone CR on first reconcile
- [ ] On CR deletion, PushSecret-managed paths in OpenBao are cleaned up
- [ ] Finalizer is removed after successful cleanup
- [ ] If OpenBao is unavailable during deletion, the finalizer blocks deletion with a clear status message
- [ ] E2E test verifies OpenBao paths are empty after CR deletion

---
Source: https://github.com/C5C3/forge/issues/193


berendt:



## WARNING (3)

### W-001: Doc/Code mismatch: finalizeOpenBaoSecrets return signature
**File**: `docs/reference/keystone-reconciler.md:820-873` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Cross-reference documentation examples against implementation

**Problem**: The documentation describes a three-value return shape for finalizeOpenBaoSecrets that does not exist in the code. The actual signature is (done bool, err error) only; the stuck object's name is recorded as a side effect on keystone.Status.Conditions via setOpenBaoFinalizerBlockedCondition, not returned as a third value. The idempotency paragraph's (true, "", nil) tuple does not match the real signature (true, nil).

**Code**:
```
3. **Cleanup.** `finalizeOpenBaoSecrets` issues `Delete` on each backup
   PushSecret then immediately re-`Get`s it. `done` is `true` only when
   **both** PushSecrets return `NotFound` on the follow-up Get. A
   still-present PushSecret (typically Terminating behind ESO's cleanup
   finalizer) sets `done=false` and records the stuck object's name in
   `stuckName`.

...

This tolerance makes `finalizeOpenBaoSecrets` **idempotent**: calling it
twice in a row with no PushSecrets present returns `(true, "", nil)` both
times...
```

**Action Required**: Rewrite both passages to match the actual two-value signature. Replace the 'records the stuck object's name in stuckName' sentence with a description of the side-effect on keystone.Status.Conditions, and replace the (true, "", nil) tuple with (true, nil).

**Suggested Fix**:
```
3. **Cleanup.** `finalizeOpenBaoSecrets` issues `Delete` on each backup
   PushSecret then immediately re-`Get`s it. `done` is `true` only when
   **both** PushSecrets return `NotFound` on the follow-up Get. A
   still-present PushSecret (typically Terminating behind ESO's cleanup
   finalizer) sets `done=false`, and the first stuck object's name is
   recorded on `keystone.Status.Conditions` via
   `setOpenBaoFinalizerBlockedCondition` so the blocked state surfaces on
   `kubectl get keystone`.

...

This tolerance makes `finalizeOpenBaoSecrets` **idempotent**: calling it
twice in a row with no PushSecrets present returns `(true, nil)` both
times...
```

---

### W-002: RBAC over-privilege: delete verb granted on externalsecrets
**File**: `operators/keystone/helm/keystone-operator/templates/_helpers.tpl:187-199` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Enforce least-privilege on wildcard access paths

**Problem**: The combined RBAC rule grants 'delete' on both externalsecrets and pushsecrets, but the only caller of r.Delete on external-secrets.io resources is finalizeOpenBaoSecrets which operates exclusively on PushSecret objects. Granting delete on externalsecrets is unnecessary and broader than least-privilege requires; a compromised operator token could delete any ExternalSecret in the cluster despite the controller never doing so by design. The kubebuilder marker in keystone_controller.go:115 has the same shape, and Helm test assertions in clusterrole_test.yaml:174-193 and role_test.yaml:267-291 mirror it.

**Code**:
```
- apiGroups:
    - external-secrets.io
  resources:
    - externalsecrets
    - pushsecrets
  verbs:
    - get
    - list
    - watch
    - create
    - update
    - patch
    - delete
```

**Action Required**: Split the RBAC rule so externalsecrets keeps the pre-existing verb set (get/list/watch/create/update/patch) and only pushsecrets gains delete. Update the kubebuilder marker at keystone_controller.go:115 to match by splitting into two lines. Update Helm test assertions in clusterrole_test.yaml:174-193 and role_test.yaml:267-291, and the Chainsaw test in tests/e2e/keystone/deletion-cleanup/ if it asserts the ClusterRole verb list.

**Suggested Fix**:
```
Split the rule into two: one for externalsecrets with verbs get/list/watch/create/update/patch (no delete), and one for pushsecrets with verbs get/list/watch/create/update/patch/delete. Mirror the change in the kubebuilder marker at keystone_controller.go:115 (two +kubebuilder:rbac lines), and update Helm test assertions in clusterrole_test.yaml:174-193 and role_test.yaml:267-291 plus any Chainsaw E2E assertion of the ClusterRole verb list.
```

---

### W-003: Serialized PushSecret cleanup doubles worst-case deletion window
**File**: `operators/keystone/internal/controller/reconcile_secrets.go:182-210` — **Fix**: ASK — **Confidence**: likely — **Pattern**: Avoid unconditional updates in reconcile loops

**Problem**: The cleanup loop does Delete → immediate Get → return false on the first stuck PushSecret. If both PushSecrets are held Terminating by ESO, the second one never has Delete issued until the first is fully garbage-collected. With RequeueSecretPolling=15s and realistic ESO purge latency, the serial path is roughly 2x slower than issuing both Deletes up front. Not a correctness issue — deletion still completes — but it stretches a wall-clock-critical operation over extra requeue intervals for no functional benefit. The integration test at integration_test.go:2545-2550 openly relies on this ordering and would need relaxation.

**Code**:
```
for _, name := range openBaoBackupPushSecretNames(keystone) {
    key := client.ObjectKey{Namespace: keystone.Namespace, Name: name}
    ps := &esov1alpha1.PushSecret{
        ObjectMeta: metav1.ObjectMeta{Name: key.Name, Namespace: key.Namespace},
    }
    if delErr := r.Delete(ctx, ps); delErr != nil {
        ...
    }

    getErr := r.Get(ctx, key, &esov1alpha1.PushSecret{})
    if apierrors.IsNotFound(getErr) {
        continue
    }
    ...
    setOpenBaoFinalizerBlockedCondition(keystone, key.Name)
    ...
    return false, nil
}
```

**Action Required**: Split the loop into two passes: pass 1 issues Delete on every backup PushSecret so all ESO cleanup finalizers fire in parallel; pass 2 reports blocked on the first PushSecret still present. Relax the integration test comment and the 'fernet transitions first' assertion to either 'at least one transitions' or to observe both PushSecrets going Terminating without ordering constraints.

**Suggested Fix**:
```
Refactor finalizeOpenBaoSecrets into two sequential passes over openBaoBackupPushSecretNames(keystone): pass 1 calls r.Delete on every name (tolerating NotFound), pass 2 calls r.Get on every name and on the first one still present calls setOpenBaoFinalizerBlockedCondition and returns (false, nil). If all Gets return NotFound return (true, nil). Relax integration_test.go:2545-2550 to not depend on fernet-keys-backup transitioning before credential-keys-backup.
```

---

## INFO (2)

### I-001: No Warning event for prolonged blocked state
**File**: `operators/keystone/internal/controller/reconcile_secrets.go:203-209` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: When a PushSecret stays stuck Terminating (OpenBao unreachable, ESO crashed, auth revoked), the only signals are the SecretsReady=False/OpenBaoFinalizerBlocked condition and a V(1) log line per pass. No Kubernetes event — neither Normal nor Warning — is emitted. Standard Kubernetes alerting tooling (kube-events exporters, OpenShift alerts) watches for Warning events, so a prolonged finalizer stall would silently not fire any alert.

**Code**:
```
setOpenBaoFinalizerBlockedCondition(keystone, key.Name)
logger.V(1).Info("openbao finalizer blocked on PushSecret garbage collection",
    "pushsecret", key.Name)
return false, nil
```

**Action Required**: Consider emitting a Warning event once per blocked transition (e.g., first pass after FinalizingOpenBaoSecrets, or after N requeue intervals). Not required by the task spec, but matches the CC-0078 event vocabulary and improves on-call experience. Leave the condition and log as the primary observability channels.

**Suggested Fix**:
```
If the team agrees on scope expansion, emit a Warning event via r.Recorder.Event(keystone, corev1.EventTypeWarning, "OpenBaoFinalizerBlocked", ...) alongside the condition update — ideally throttled so it fires once per transition into the blocked state rather than every requeue.
```

**Related**: I-002

---

### I-002: Event suppression can hide cleanup trigger after operator restart
**File**: `operators/keystone/internal/controller/keystone_controller.go:360-367` — **Fix**: ASK — **Confidence**: likely

**Problem**: hasLiveOpenBaoBackupPushSecrets treats a PushSecret with DeletionTimestamp != 0 as 'not live' (intentional, to suppress repeat events across requeues). But if the operator pod restarts while the PushSecret is already Terminating (e.g., ESO stuck), the first reconcile after restart sees hasLive=false, skips the FinalizingOpenBaoSecrets event, and the Keystone CR appears to enter blocked state silently without any operator-visible announcement. The OpenBaoFinalizerBlocked condition still surfaces the state, so it is not a correctness issue — an observability gap acceptable since restarts are rare.

**Code**:
```
hasLiveCleanupWork, err := r.hasLiveOpenBaoBackupPushSecrets(ctx, keystone)
if err != nil {
    return ctrl.Result{}, err
}
if hasLiveCleanupWork {
    r.Recorder.Event(keystone, corev1.EventTypeNormal, "FinalizingOpenBaoSecrets",
        "Cleaning up OpenBao backup PushSecrets before removing Keystone")
}
```

**Action Required**: Consider whether the restart-silence case warrants special handling (e.g., emitting the FinalizingOpenBaoSecrets event when hasLive=false but a Terminating PushSecret is observed on the first reconcile after a restart). Accept as-is if the blocked condition is deemed sufficient.

**Suggested Fix**:
```
If accepted as in-scope, widen hasLiveOpenBaoBackupPushSecrets (or add a sibling check) to also report 'live' when a Terminating PushSecret is seen and no prior FinalizingOpenBaoSecrets event has been recorded for this generation, so operator restarts still announce the cleanup once.
```

**Related**: I-001

---

## Summary

The PR correctly implements the OpenBao PushSecret finalizer with appropriate defense in depth (Keystone finalizer holds the CR, PushSecret DeletionPolicy=Delete drives the remote KV-v2 purge via ESO, owner references as fallback). Test coverage is comprehensive across unit, reconcile, builder, integration, and E2E layers, and the documentation addition is thorough. Three findings of note: a doc passage in keystone-reconciler.md describes a non-existent third return value of finalizeOpenBaoSecrets, the new RBAC rule grants delete on externalsecrets even though only pushsecrets are ever deleted, and the cleanup loop's serial Delete-then-Get structure doubles the worst-case deletion window when both backups are stuck. None are blockers.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Merge-ready once the doc mismatch and RBAC split are applied; both are cheap mechanical cleanups worth landing with this PR. The serialized-cleanup change requires a small test relaxation and can be landed in this PR or a follow-up. The two INFO observability items are optional and should be triaged separately.

gndrmnn (approved):

Issue #194, opened 2026-04-12
Add field indexer for Secret-to-Keystone mapper to reduce API server load


berendt:
**Category**: performance | **Scope**: Small

**Description**: Replace the current secretToKeystoneMapper implementation (keystone_controller.go:213) which lists all Keystone CRs on every Secret event with a field indexer registered via mgr.GetFieldIndexer(). Index Keystone CRs by their referenced Secret names so the mapper can perform an O(1) indexed lookup instead of an O(n) list operation.

**Motivation**: In namespaces with many Secrets (common with ExternalSecretOperator), every Secret create/update/delete triggers a full list of all Keystone CRs against the API server. This generates unnecessary API server load that scales linearly with Secret count. Field indexers are the standard controller-runtime pattern for efficient reverse lookups.

**Affected Areas**:
- `operators/keystone/internal/controller/keystone_controller.go`

**Acceptance Criteria**:
- [ ] Field indexer registered for Secret name references on Keystone CRs
- [ ] secretToKeystoneMapper uses indexed lookup instead of List
- [ ] API server request count for Keystone CR listing reduced to zero during Secret-only events
- [ ] Unit test verifies mapper returns correct reconcile requests via index
- [ ] No behavioral change in reconcile triggering

berendt:
Fixed by https://github.com/C5C3/forge/pull/262.

PR #262, opened 2026-04-22
feat(CC-0087): Add field indexer for Secret-to-Keystone mapper to reduce AP

berendt:
# GitHub Issue #194: Add field indexer for Secret-to-Keystone mapper to reduce API server load

**Category**: performance | **Scope**: Small

**Description**: Replace the current secretToKeystoneMapper implementation (keystone_controller.go:213) which lists all Keystone CRs on every Secret event with a field indexer registered via mgr.GetFieldIndexer(). Index Keystone CRs by their referenced Secret names so the mapper can perform an O(1) indexed lookup instead of an O(n) list operation.

**Motivation**: In namespaces with many Secrets (common with ExternalSecretOperator), every Secret create/update/delete triggers a full list of all Keystone CRs against the API server. This generates unnecessary API server load that scales linearly with Secret count. Field indexers are the standard controller-runtime pattern for efficient reverse lookups.

**Affected Areas**:

**Acceptance Criteria**:
- [ ] Field indexer registered for Secret name references on Keystone CRs
- [ ] secretToKeystoneMapper uses indexed lookup instead of List
- [ ] API server request count for Keystone CR listing reduced to zero during Secret-only events
- [ ] Unit test verifies mapper returns correct reconcile requests via index
- [ ] No behavioral change in reconcile triggering

---
Source: https://github.com/C5C3/forge/issues/194


berendt:



## INFO (3)

### I-001: Non-NotFound Get errors silently swallowed without logging
**File**: `operators/keystone/internal/controller/keystone_controller.go:623-627` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The owner-ref Get path intentionally falls through to enqueue on non-NotFound errors (so transient cache blips do not swallow a legitimate event), but unlike the indexed-List branch at line 603 which logs on error, this path produces zero operational signal. If the informer cache is unhealthy (mid-sync, disconnected, GVK not registered), every owner-ref will flow through unlogged while the mapper silently assumes the target exists.

**Code**:
```
var ks keystonev1alpha1.Keystone
if err := c.Get(ctx, key, &ks); apierrors.IsNotFound(err) {
    continue
}
seen[key] = struct{}{}
```

**Action Required**: Log at V(1) when the Get error is non-nil and not-NotFound, including the secret and ownerRef keys plus the error, before continuing to enqueue.

**Suggested Fix**:
```
var ks keystonev1alpha1.Keystone
if err := c.Get(ctx, key, &ks); err != nil {
    if apierrors.IsNotFound(err) {
        continue
    }
    log.FromContext(ctx).V(1).Info("owner-ref Get returned non-NotFound error; enqueueing anyway",
        "secret", client.ObjectKeyFromObject(obj), "ownerRef", key, "error", err)
}
seen[key] = struct{}{}
```

---

### I-002: Flakiness risk in TestIntegration_UnrelatedSecretDoesNotTriggerReconcile
**File**: `operators/keystone/internal/controller/integration_test.go:187-193` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The test uses a 1s Consistently window to prove the unrelated Secret did not drive a reconcile. This negative test's reliability depends on the controller's periodic requeue cadence being longer than 1s. If a future change reduces the periodic requeue interval, this test will silently become vacuous (still pass) or flaky. A stronger assertion would capture controller metrics (e.g., reconcile counter) across the window and assert no increment.

**Code**:
```
g.Consistently(func(ig Gomega) {
    got := &keystonev1alpha1.Keystone{}
    ig.Expect(c.Get(ctx, key, got)).To(Succeed())
    ig.Expect(got.ResourceVersion).To(Equal(stableRV), ...)
}, 1*time.Second, pollInterval/2).Should(Succeed(), ...)
```

**Action Required**: Track as a follow-up if CI flake materializes. Current form is acceptable per the author's documented rationale. Consider switching to a metrics-based assertion (reconcile counter delta) if flakes appear.

**Suggested Fix**:
```
Track as a follow-up: if CI flake appears, replace the ResourceVersion-stability assertion with a controller reconcile-counter metric delta assertion over the same window, which is independent of the periodic requeue interval.
```

---

### I-003: Hardcoded "Keystone" Kind literal in owner-ref match
**File**: `operators/keystone/internal/controller/keystone_controller.go:612` — **Fix**: ASK — **Confidence**: verified

**Problem**: The Kind string "Keystone" is hardcoded in the owner-ref comparison. If the type is ever renamed, this comparison silently stops matching and rotation-staging Secrets stop triggering reconciles. The project does not have a KindKeystone constant elsewhere, so this matches existing convention, but a compile-time-derived form or an exported const on the API package would be safer.

**Code**:
```
for _, ref := range obj.GetOwnerReferences() {
    if ref.Kind != "Keystone" {
        continue
    }
```

**Action Required**: Optional — extract const KeystoneKind = "Keystone" in the API package (alongside GroupVersion) and reference it here, or derive via reflect.TypeOf(&keystonev1alpha1.Keystone{}).Elem().Name().

**Suggested Fix**:
```
Add an exported constant (e.g., KeystoneKind = "Keystone") in the keystonev1alpha1 API package alongside GroupVersion, and replace the "Keystone" literal here with keystonev1alpha1.KeystoneKind. Discuss with the team whether to adopt this convention repo-wide.
```

---

## Summary

Carefully implemented PR that adds a field indexer for the Secret-to-Keystone mapper, preserving owner-ref-based backward compatibility for rotation-staging Secrets while eliminating API server amplification for ESO-managed Secrets. Sourcery-ai review #1 was resolved with substantive code and test changes (group-only owner matching, cached existence check), not token fixes. Test coverage (13 unit + 3 integration) and documentation are thorough. Only three minor INFO findings: a logging gap on transient Get errors, an author-acknowledged flakiness window in one integration test, and a hardcoded Kind literal.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 0 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Merge. All three findings are INFO-level and non-blocking; the logging-gap fix is a trivial auto-fix that could land in this PR or a follow-up, and the other two are optional/defer-until-needed.

operators/keystone/internal/controller/keystone_controller.go
```diff
@@ -545,27 +567,51 @@ func (r *KeystoneReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 // secretToKeystoneMapper returns a MapFunc that maps Secret events to reconcile
-// requests for Keystone CRs that reference the Secret by name or own it.
+// requests for Keystone CRs that either reference the Secret by name
+// (resolved via the KeystoneSecretNameIndexKey field indexer) or own it via
+// an OwnerReference with Kind=Keystone and APIVersion matching
+// keystonev1alpha1.GroupVersion (e.g. rotation staging Secrets).
+//
+// The index path replaces the prior unfiltered namespace-scoped List so that a
+// high-frequency Secret event no longer forces the API server to return every
+// Keystone CR in the namespace (CC-0087, REQ-001, REQ-002). The owner-ref path
```
gndrmnn: Do not reference prior implementation details in code comments. Nobody evaluating the new code will know or care about how it was once implemented. Only document current behavior and behavior that could look odd for someone looking at the code for the first time.

operators/keystone/internal/controller/keystone_controller_test.go
```diff
@@ -2480,3 +2495,326 @@ func TestKeystoneSecretIndexer_SkipsEmptyNames(t *testing.T) {
...
+// is the exact string that registerSecretNameIndex passes to IndexField.
+// A future refactor that inlines the literal at either site will break this
+// test (CC-0087, REQ-006).
+func TestKeystoneSecretIndexerKey_IsDeclaredAsConst(t *testing.T) {
+	g := NewGomegaWithT(t)
+
+	// The const value is a stable contract surface: any drift requires a
+	// matched update at every consumer (registration and lookup sites).
+	g.Expect(KeystoneSecretNameIndexKey).To(Equal("spec.secretRefs.name"),
+		"const value drift breaks the field-indexer contract with the API server")
```
gndrmnn: Do not test for constant value literals

operators/keystone/internal/controller/keystone_controller_test.go
```diff
@@ -2747,3 +2762,556 @@ func TestSecretToKeystoneMapper_DerivedDBConnectionSecret(t *testing.T) {
...
```
gndrmnn: Remove lines 3287-3290. We have no interest in testing the value of constants we declare ourselves. There is no good reason to test if `KeystoneSecretNameIndexKey` has the value we assign to it.

gndrmnn (approved):

PR #263, opened 2026-04-22
feat(CC-0084): Harden graceful pod termination in Keystone rolling updates

berendt:
## Scope notes (post review #1, CC-0084)

- **Out of scope — follow-up PR against `architecture/` submodule**: the HA architecture rollout-flow subsection called out in the original acceptance criteria (`architecture/docs/04-architecture/04-high-availability.md`) is deferred to a separate PR against the `architecture` submodule because this worktree is a worktree-level fork that cannot modify submodule content from inside itself (see worktree rule #12 in `CLAUDE.md` and the blocker documented in commit `607b423`; dead cross-links were removed in `6b38719`). A follow-up ticket will track the submodule subsection covering: EndpointSlice removal ‖ preStop sleep → SIGTERM → uWSGI graceful shutdown with harakiri fallback → SIGKILL, plus recommended production values for `terminationGracePeriodSeconds`, `preStopSleepSeconds`, `harakiri`, and `httpKeepAliveTimeout`.

---

# GitHub Issue #247: Harden graceful pod termination during Keystone rolling updates

**Description**: `buildKeystoneDeployment` (operators/keystone/internal/controller/reconcile_deployment.go:138-186) relies on a hardcoded `terminationGracePeriodSeconds: 30` paired with a 5 s `preStop` sleep (`/bin/sh -c "sleep 5"`) and uWSGI's `--master` signal handling to drain in-flight requests during rolling updates (e.g. `kubectl patch keystone ... spec.image.tag`). This baseline works for the common case but leaves four production gaps: (1) the 30 s grace period is not user-configurable, leaving only ~25 s for uWSGI to finish in-flight requests — insufficient for workloads with slow LDAP/DB token validation; (2) `uwsgiCommand` (reconcile_deployment.go:278-306) does not set `--harakiri`, so a single hung request can block graceful shutdown until SIGKILL; (3) `--http-keepalive` defaults to `true` (reconcile_deployment.go:281-295) with no keep-alive timeout, letting clients reuse a connection to a pod that has already been removed from the Service endpoints; (4) `Deployment.Spec.Strategy` is never set, so the Kubernetes default `RollingUpdate` with `maxUnavailable: 25%` / `maxSurge: 25%` applies — combined with the `minAvailable: 1` PDB (reconcile_deployment.go:248-271) this permits brief capacity dips rather than zero-downtime rollouts. Expose `spec.terminationGracePeriodSeconds`, `spec.preStopSleepSeconds`, `spec.uwsgi.harakiri`, and `spec.uwsgi.httpKeepAliveTimeout` as optional CRD fields; set an explicit `RollingUpdate` strategy with `maxUnavailable: 0`, `maxSurge: 1` as the new default and make it overridable via `spec.strategy`.

**Motivation**: A `kubectl patch keystone ... spec.image.tag` triggers a pod rotation that must deliver both (a) no new requests routed to a terminating pod and (b) in-flight requests completing cleanly. The current implementation mitigates (a) via the 5 s preStop sleep that covers EndpointSlice propagation, but keep-alive connections bypass this protection because clients keep reusing an already-stale socket. (b) is limited by a fixed 25 s drain window without per-worker hang detection — a single stuck token validation blocks shutdown until SIGKILL truncates every other in-flight request on the same pod. Operators cannot tune any of these values today without forking the operator. At the same time, the default `maxUnavailable: 25%` reduces Keystone capacity during rollout — Keystone is the authentication gate for the entire control plane, so any capacity dip amplifies user-visible latency across all OpenStack services.

**Affected Areas**:
- `operators/keystone/api/v1alpha1/keystone_types.go`
- `operators/keystone/api/v1alpha1/keystone_webhook.go`
- `operators/keystone/helm/keystone-operator/crds/keystone.openstack.c5c3.io_keystones.yaml`
- `operators/keystone/config/crd/bases/keystone.openstack.c5c3.io_keystones.yaml`
- `architecture/docs/04-architecture/04-high-availability.md`

**Acceptance Criteria**:
- [ ] `KeystoneSpec` gains optional `terminationGracePeriodSeconds *int64` (default 30, min 10) and `preStopSleepSeconds *int64` (default 5, min 0)
- [ ] `UWSGISpec` gains optional `harakiri *int32` (worker request timeout in seconds, default unset) and `httpKeepAliveTimeout *int32` (idle keep-alive timeout in seconds, default unset)
- [ ] `buildKeystoneDeployment` sets `.Spec.Strategy` to `RollingUpdate` with `maxUnavailable: 0` and `maxSurge: 1` by default; new optional `spec.strategy *appsv1.DeploymentStrategy` field allows full override
- [ ] `preStop` sleep duration is derived from `spec.preStopSleepSeconds` instead of the hardcoded `sleep 5`
- [ ] `uwsgiCommand` appends `--harakiri <n>` when set and `--http-keepalive-timeout <n>` when set
- [ ] Validating webhook rejects `preStopSleepSeconds >= terminationGracePeriodSeconds` and rejects `harakiri >= (terminationGracePeriodSeconds - preStopSleepSeconds)` so a worker timeout cannot exceed the remaining drain window
- [ ] Unit tests in `reconcile_deployment_test.go` cover: defaults (grace=30, preStop=5, maxUnavailable=0, maxSurge=1, no harakiri, no keep-alive timeout), all four fields overridden, strategy override replaces defaults, webhook validation paths (preStop >= grace rejected, harakiri too large rejected)
- [ ] E2E test verifies a rolling image-tag patch produces zero failed requests under a steady load generator running against the Service
- [ ] CRD manifests in both `helm/.../crds/` and `config/crd/bases/` are regenerated and committed

**Labels:** enhancement

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/247


berendt:



## WARNING (3)

### W-001: Aggregate validation test not updated for CC-0084 fields
**File**: `operators/keystone/api/v1alpha1/keystone_webhook_test.go:1109-1198` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Update aggregate validation tests when adding new validation paths

**Problem**: This PR adds seven new webhook validations (REQ-001/002/003/004/006/007/008/012) but TestValidateCreate_RunsAllValidations — the sentinel test whose stated purpose is simultaneous exercise of every validation rule to prove error accumulation with no short-circuiting — was not updated. A future regression that short-circuits before reaching the new rules will go undetected for graceful-termination fields. The project has been flagged for exactly this omission before (CC-0075, where PriorityClassName and TopologySpreadConstraints were left out of the same test until a reviewer caught it).

**Code**:
```
g.Expect(errMsg).To(ContainSubstring("replicas"))
g.Expect(errMsg).To(ContainSubstring("maxActiveKeys"))
// … 12 more substrings, none for CC-0084 fields
```

**Action Required**: Add the new broken fields (TerminationGracePeriodSeconds, PreStopSleepSeconds, Harakiri, HTTPKeepAliveTimeout, HTTPKeepAlive, Strategy=Recreate with RollingUpdate block) to the test setup and add matching substring assertions for terminationGracePeriodSeconds, preStopSleepSeconds, harakiri, httpKeepAliveTimeout, and strategy.

**Suggested Fix**:
```
	// REQ-001,002,007,008 (CC-0084): Break graceful-termination envelope.
	grace := int64(10)
	preStop := int64(30)
	k.Spec.TerminationGracePeriodSeconds = &grace
	k.Spec.PreStopSleepSeconds = &preStop
	harakiri := int32(50)
	// REQ-006 (CC-0084): Break strategy — Recreate with RollingUpdate block.
	k.Spec.Strategy = &appsv1.DeploymentStrategy{
		Type: appsv1.RecreateDeploymentStrategyType,
		RollingUpdate: &appsv1.RollingUpdateDeployment{},
	}
```

---

### W-002: E2E test hardcodes replica count instead of reading from .spec.replicas
**File**: `tests/e2e/keystone/rolling-update-zero-downtime/chainsaw-test.yaml:214` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Derive expected values from source of truth instead of hardcoding

**Problem**: The availability-polling script hardcodes EXPECTED_REPLICAS=3 while 00-keystone-cr.yaml declares replicas: 3. Two sibling tests already follow the correct pattern — tests/e2e-chaos/memcached-pod-kill/chainsaw-test.yaml:64 and tests/e2e-chaos/operator-pod-kill/chainsaw-test.yaml:74 — so the precedent is established in this repository. If the CR's .spec.replicas is ever changed (e.g. to exercise a different surge/drain cadence), the test will silently validate the wrong invariant.

**Code**:
```
          set -eu
          EXPECTED_REPLICAS=3
          DEADLINE=$(( $(date +%s) + 75 ))
          MIN=999
```

**Action Required**: Replace the hardcoded EXPECTED_REPLICAS=3 with a kubectl jsonpath read from the deployment's .spec.replicas, with a failure path if the read returns empty.

**Suggested Fix**:
```
          set -eu
          EXPECTED_REPLICAS=$(kubectl -n openstack get deploy keystone-rolling-update-api \
            -o jsonpath='{.spec.replicas}')
          if [ -z "$EXPECTED_REPLICAS" ]; then
            echo "FAIL: could not read .spec.replicas from deployment"
            exit 1
          fi
          DEADLINE=$(( $(date +%s) + 75 ))
          MIN=999
```

**Related**: E2E `assert` steps use `>= 3` where exact equality holds

---

### W-003: E2E assert steps use loose `>= 3` where exact equality holds
**File**: `tests/e2e/keystone/rolling-update-zero-downtime/chainsaw-test.yaml:65-336` — **Fix**: AUTO-FIX — **Confidence**: likely — **Pattern**: Use exact-match assertions instead of loose comparisons

**Problem**: Step 2 (assert-ready, pre-rollout, line 65) and Step 7 (assert-rollout-complete, post-rollout, lines 322 and 336) both assert (availableReplicas >= 3): true. At both of those checkpoints the system is in steady state — availableReplicas == 3 holds deterministically. CC-0066's remediation replaced availableReplicas >= 2 with exact-match in the existing e2e test suite; this regresses that pattern for two new assert sites. The polling loop inside Step 5 correctly uses exact less-than comparison (if [ "$AVAIL" -lt "$EXPECTED_REPLICAS" ]), so only the declarative asserts need tightening.

**Code**:
```
          status:
            (availableReplicas >= `3`): true
```

**Action Required**: Replace `(availableReplicas >= `3`): true` with `availableReplicas: 3` at lines 65, 322, and the closing assert. Line 336 ((updatedReplicas == replicas): true) is correct as-is.

**Suggested Fix**:
```
          status:
            availableReplicas: 3
```

**Related**: W-002

---

## INFO (6)

### I-001: Leftover "Reviewer: please verify" prompt in E2E fixture
**File**: `tests/e2e/keystone/rolling-update-zero-downtime/01-patch-image.yaml:12-13` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Resolve ambiguous commented-out alternatives before merge

**Problem**: The review-1 (commit a98b432) removed an identical "Reviewer: please verify this matches intent." phrase from keystone_types.go:154, but the same pattern in the new test fixture was missed. The 2025.2-upgraded tag IS pre-loaded by CI (.github/workflows/ci.yaml:611-614), so the prompt has been resolved and the comment can be removed.

**Code**:
```
# Reviewer: please verify the '2025.2-upgraded' tag is pre-loaded into kind
# for this test lane.
```

**Action Required**: Replace the reviewer-prompt comment with a factual statement of how the tag is made available in CI.

**Suggested Fix**:
```
# The '2025.2-upgraded' tag is pre-loaded into kind by ci.yaml:611-614
# alongside 2025.2, enabling the image-tag patch without an external pull.
```

---

### I-002: Timing-budget comment's arithmetic does not match its stated inequality
**File**: `tests/e2e/keystone/rolling-update-zero-downtime/chainsaw-test.yaml:200-209` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The BUDGET NOTE claims "Three surge-then-drain cycles at ~35 s each fit comfortably under 75 s." Three 35-second cycles are 105 seconds, not ≤ 75. A future maintainer raising the grace period who relies on this comment's arithmetic will under-budget the new poll window. The actual test behavior is probably fine (uWSGI drains well before the 30s grace maximum, cycles complete closer to 20s each), but the comment's stated rationale does not support its numbers.

**Code**:
```
  # DefaultPreStopSleepSeconds). Three surge-then-drain cycles at ~35 s each fit
  # comfortably under 75 s. If either default is raised in a future change (or
  # this test's CR overrides them), revisit this budget — a grace of 60 s would
  # push three cycles to ~195 s and make this loop silently flaky.
```

**Action Required**: Rephrase the comment to describe the empirical expectation (~20-25s per cycle) rather than assert an arithmetic inequality that does not hold.

**Suggested Fix**:
```
  # DefaultPreStopSleepSeconds). In practice each pod replacement cycle takes
  # ~20-25 s (readiness probe + early uWSGI graceful exit well before the 30 s
  # grace cap), so three sequential cycles complete in 60-75 s. If either
  # default is raised, or the test's CR overrides them, recompute: with a grace
  # of 60 s the per-cycle worst case rises to ~65 s and this poll budget must
  # be widened accordingly.
```

---

### I-003: Docs claim `Default: 30` but CRD emits no kubebuilder default
**File**: `docs/reference/keystone-crd.md:157-158` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The KeystoneSpec fields table lists terminationGracePeriodSeconds with Default: 30 and preStopSleepSeconds with Default: 5. The CRD schema emits no default: for either field (by design — the DECISION comment explains this preserves the nil-pointer upgrade contract). The effective behavior is indeed 30/5 via the reconciler fallback, but a user reading this row will not find "30" in kubectl get keystone -o yaml when they omit the field. The ambiguity is downstream: does "Default: 30" mean "the API server materializes 30 on admission" or "the reconciler treats nil as 30"? Other pointer fields in the same table (priorityClassName, strategy) are more careful.

**Code**:
```
| `terminationGracePeriodSeconds` | `*int64` | No | `30` | Grace period (seconds) …
| `preStopSleepSeconds` | `*int64` | No | `5` | Sleep duration (seconds) …
```

**Action Required**: Align the style with strategy/priorityClassName — put the resolved behavior in the Effect column and document the default as nil with an explanation of reconciler-applied value.

**Suggested Fix**:
```
| `terminationGracePeriodSeconds` | `*int64` | No | `nil` | Grace period (seconds) granted to Keystone API pods between SIGTERM and SIGKILL during rolling updates. Minimum: `10`. When `nil`, the reconciler applies `30` to the PodSpec (Kubernetes default). See [Graceful-termination fields](#graceful-termination-fields-cc-0084) (CC-0084). |
| `preStopSleepSeconds` | `*int64` | No | `nil` | Sleep duration (seconds) of the preStop lifecycle hook. Minimum: `0`. Must be strictly less than `terminationGracePeriodSeconds`. When `nil`, the reconciler applies `5`. See [Graceful-termination fields](#graceful-termination-fields-cc-0084) (CC-0084). |
```

**Related**: I-004

---

### I-004: CRD `description` contains an internal DECISION block
**File**: `operators/keystone/config/crd/bases/keystone.openstack.c5c3.io_keystones.yaml:814-823` — **Fix**: ASK — **Confidence**: verified

**Problem**: The godoc on TerminationGracePeriodSeconds was kept intentionally detailed, but the DECISION: block explains internal rationale ("Task 1.1 title mentions 'default=30' but REQ-001's scenario explicitly requires…") that controller-gen now materializes into the CRD's user-facing description. kubectl explain keystone.spec.terminationGracePeriodSeconds will spill this to every operator. The DECISION rationale belongs in a non-exported comment (e.g., a Go file comment outside the struct godoc) or in the architecture docs, not in the CRD shown to end-users.

**Code**:
```
  DECISION: Task 1.1 title mentions "default=30" but REQ-001's scenario
  explicitly requires "webhook Default() leaves the pointer nil so an
  upgrade does not mutate existing CRs". A +kubebuilder:default=30 marker
  on a pointer field would cause the API server to materialize the value
  at admission, mutating pre-existing CRs on operator upgrade — exactly
  what REQ-001 forbids. The marker is therefore omitted and the effective
  "default 30" is applied by the reconciler (task 3.x) when the pointer
  is nil, mirroring the existing AutoscalingSpec.MinReplicas pattern.
```

**Action Required**: Move the DECISION text out of the godoc on KeystoneSpec.TerminationGracePeriodSeconds (operators/keystone/api/v1alpha1/keystone_types.go:141-152) into a non-godoc comment block above the field — e.g., a // Note: line that controller-gen does not pick up — then re-run make manifests && make sync-crds. The user-facing description should explain what the field does and its bounds, not why the kubebuilder marker was omitted.

**Suggested Fix**:
```
Move the DECISION block from the godoc comment on KeystoneSpec.TerminationGracePeriodSeconds in operators/keystone/api/v1alpha1/keystone_types.go:141-152 to a separate non-godoc comment (blank line between godoc and the Note:) or to architecture documentation. Regenerate the CRD with `make manifests && make sync-crds` so the user-facing description only documents the field's purpose, bounds, and nil behavior.
```

**Related**: I-003

---

### I-005: Type tests mostly verify Go struct/DeepCopy language semantics
**File**: `operators/keystone/api/v1alpha1/keystone_types_test.go:93-214` — **Fix**: ASK — **Confidence**: likely — **Pattern**: Remove tests that only verify language semantics

**Problem**: TestKeystoneSpecTerminationGracePeriodSecondsField, TestKeystoneSpecStrategyField, and TestUWSGISpecOptionalTimeoutFields set fields on a struct and then assert the fields have the values just set, plus that DeepCopy() allocates new memory. These tests have been flagged in past reviews as low-value (CC-0013, CC-0078: "we certainly do not need to test if a struct we just created has values we assign to it"). They do partially protect against stale zz_generated.deepcopy.go (the pointer-identity assertions would fail if controller-gen was not run and **out = **in was used for the new pointer fields), but the reconciler and envtest tests already exercise the deep-copied values end-to-end, so the narrow coverage here is redundant.

**Code**:
```
TestKeystoneSpecTerminationGracePeriodSecondsField, TestKeystoneSpecStrategyField, TestUWSGISpecOptionalTimeoutFields — set a field, assert it has that value, assert DeepCopy() yields a distinct pointer.
```

**Action Required**: Decide whether to keep these for belt-and-braces coverage of the generated file or delete them per the CC-0013/CC-0078 convention. The repo's convention on this is ambiguous, so a specific fix is not proposed.

**Suggested Fix**:
```
Either delete TestKeystoneSpecTerminationGracePeriodSecondsField, TestKeystoneSpecStrategyField, and TestUWSGISpecOptionalTimeoutFields (aligning with CC-0013/CC-0078 feedback) or keep them as stale-deepcopy tripwires. Reach team consensus first.
```

---

### I-006: HA architecture doc subsection deferred to follow-up PR
**File**: `architecture/docs/04-architecture/04-high-availability.md` — **Fix**: ASK — **Confidence**: verified

**Problem**: PR description lists this as an affected area, and task 6.1 in the planwerk progress file explicitly mentions a "HA rollout-flow subsection" as part of the acceptance criteria. Commit 607b423 documents the blocker (submodule rule), and commit 6b38719 removes the now-dead #keystone-rolling-update-cc-0084 cross-links so the docs build unblocks. This is transparent and explicitly deferred to a separate PR against the submodule — but it means the acceptance criterion listed in the PR description is NOT satisfied by this PR.

**Code**:
```
PR description lists architecture/docs/04-architecture/04-high-availability.md as an affected area; commits 607b423 and 6b38719 document the submodule blocker and remove the dead anchor cross-links.
```

**Action Required**: Update the PR description to mark the HA doc entry as out-of-scope and link the follow-up ticket that will add the submodule subsection.

**Suggested Fix**:
```
Edit the PR description so the HA architecture doc entry is marked out-of-scope for this PR, with an explicit link to the follow-up ticket that will land the submodule subsection.
```

---

## Summary

The implementation is thorough and well-architected: shared Default constants give the webhook and reconciler a single source of truth, TestReconcileAndWebhookDefaultsAgree pins the invariant against drift, the webhook accumulates validation errors correctly, and the E2E test's steady-load-then-patch approach exercises the zero-downtime contract on a real cluster. The pointer-typed optional fields correctly preserve brownfield CRs on upgrade. Two issues warrant action before merge: the aggregate validation test is stale (same hole caught post-merge on CC-0075) and the E2E availability-poll hardcodes a replica count the repo already teaches us to read from .spec.replicas. Remaining findings are documentation quality and minor cleanup.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 6 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing the two WARNINGs (aggregate validation test coverage for the new CC-0084 fields, and deriving EXPECTED_REPLICAS from .spec.replicas). INFO-level findings are worth addressing but do not block merge; the HA architecture doc is explicitly deferred and should be called out in the PR description as out-of-scope.

tests/e2e/keystone/rolling-update-zero-downtime/01-patch-image.yaml
```diff
@@ -9,8 +9,8 @@
 # image-upgrade test (see tests/e2e/keystone/image-upgrade/01-patch-image.yaml).
 # This keeps a single pre-load responsibility in the CI pipeline rather than
 # introducing a second alias.
-# Reviewer: please verify the '2025.2-upgraded' tag is pre-loaded into kind
-# for this test lane.
+# The '2025.2-upgraded' tag is pre-loaded into kind by ci.yaml:611-614
+# alongside 2025.2, enabling the image-tag patch without an external pull.
```
gndrmnn: Remove the specific line reference `611-614` as this will likely be stale in the future

Issue #265, opened 2026-04-22
Expose the Keystone API via Gateway API in the Quick Start using Envoy Gateway and nip.io


berendt:
**Category**: enhancement | **Scope**: Medium

**Description**: Take the optional `spec.gateway` HTTPRoute sub-reconciler landed in CC-0065 (#238, commit `0b2c7d9`) all the way through the Quick Start, so that a fresh kind cluster ends with the Keystone API reachable at a fixed, externally-looking hostname — `https://keystone.127-0-0-1.nip.io/v3` — instead of the current `kubectl port-forward svc/keystone-api -n openstack 5000:5000` workaround (`docs/quick-start.md:498-546`). Using the public `nip.io` wildcard resolver means any `*.127-0-0-1.nip.io` label resolves to `127.0.0.1` over regular DNS with no `/etc/hosts` editing, no MetalLB, and no CoreDNS customisation — the hostname is stable, identical on every developer's machine, and round-trips cleanly through the Keystone CRD's `status.endpoint` derivation (CC-0065: *"Derive status.endpoint from spec.gateway.hostname as https://{hostname}/v3 when gateway is set"*). Today `hack/deploy-infra.sh:707-729` installs the Gateway API `standard-install.yaml` CRDs but no controller, no `GatewayClass`, and no `Gateway`, so `spec.gateway` is effectively a prod-only code path the Quick Start never exercises. The HTTPRoute E2E (`tests/e2e/keystone/httproute/`) compensates by applying the stub CRD from `00-httproute-crd.yaml` and patching `Accepted=True` onto the parent status manually — a fresh onboarding user has no such shortcut.

[...]

Issue #257, opened 2026-04-21
Enable the flux-operator Web UI in the kind overlay and document access in the Quick Start


berendt:
**Category**: enhancement | **Scope**: Small

**Description**: Enable the flux-operator's embedded Flux Web UI in the kind overlay as an additional demo surface alongside Headlamp (`deploy/kind/base/headlamp.yaml`) and the OpenBao UI (`deploy/kind/base/kustomization.yaml:42-57`), and add a Quick Start "Step 4c" section that covers port-forward + access. The Web UI is shipped by ControlPlane as part of the same `flux-operator` chart that CC-0085 already bootstraps (`hack/deploy-infra.sh:58,701` — `FLUX_OPERATOR_VERSION=\"v0.47.0\"`, installed via `kubectl apply -f .../install.yaml`), and since the operator is installed from `install.yaml` rather than Helm, its `web.enabled` Helm value stays at the default (off). That matches the documented prerequisite for the standalone install path (\"Flux Operator configured with web UI disabled (`web.enabled: false`)\" — [fluxoperator.dev/docs/web-ui/standalone-install](https://fluxoperator.dev/docs/web-ui/standalone-install/)), which deploys the UI as a separate `HelmRelease` driven by a `ResourceSet` that pulls the `flux-operator` chart from `oci://ghcr.io/controlplaneio-fluxcd/charts/flux-operator` with `web.serverOnly: true` and `installCRDs: false`.

[...]

Issue #246, opened 2026-04-20
Enable the OpenBao UI in the kind overlay and document access in the Quick Start


berendt:
**Category**: enhancement | **Scope**: Small

**Description**: Flip `ui = true` in the standalone server config produced by the kind overlay patch (`deploy/kind/base/kustomization.yaml:42`) and add a Quick Start step analogous to "Step 4 — Open the Headlamp UI" (`docs/quick-start.md:150`) that walks through port-forwarding to the OpenBao service and logging in with the root token already stashed in the `openbao-init-keys` Secret. The listener is already TLS-terminated via cert-manager on `:8200` (`deploy/kind/base/kustomization.yaml:44-50`), so the UI is reachable at `https://localhost:8200/ui/` after `kubectl port-forward -n openbao-system svc/openbao 8200:8200`. Token retrieval is a one-liner that is already documented in `docs/reference/infrastructure/openbao-bootstrap.md:170-171`: `kubectl get secret openbao-init-keys -n openbao-system -o jsonpath='{.data.init-output}' | base64 -d | jq -r '.root_token'`. The production HelmRelease (`deploy/flux-system/releases/openbao.yaml:43`) keeps `ui = false` unchanged — the flip is intentionally scoped to the kind overlay only, mirroring the kind-only posture of Headlamp (see the header comment in `deploy/kind/base/headlamp.yaml:12-15`).

Upstream docs for the toggle: https://openbao.org/docs/configuration/ui/. Requirements are minimal — a single `ui = true` top-level key plus at least one listener (which we already have).

[...]

PR #266, opened 2026-04-22
feat(CC-0088): Expose Keystone API via Gateway API with Envoy and nip.io

berendt:
# GitHub Issue #265: Expose the Keystone API via Gateway API in the Quick Start using Envoy Gateway and nip.io

**Description**: Take the optional `spec.gateway` HTTPRoute sub-reconciler landed in CC-0065 (#238, commit `0b2c7d9`) all the way through the Quick Start, so that a fresh kind cluster ends with the Keystone API reachable at a fixed, externally-looking hostname — `https://keystone.127-0-0-1.nip.io/v3` — instead of the current `kubectl port-forward svc/keystone-api -n openstack 5000:5000` workaround (`docs/quick-start.md:498-546`). Using the public `nip.io` wildcard resolver means any `*.127-0-0-1.nip.io` label resolves to `127.0.0.1` over regular DNS with no `/etc/hosts` editing, no MetalLB, and no CoreDNS customisation — the hostname is stable, identical on every developer's machine, and round-trips cleanly through the Keystone CRD's `status.endpoint` derivation (CC-0065: *"Derive status.endpoint from spec.gateway.hostname as https://{hostname}/v3 when gateway is set"*). Today `hack/deploy-infra.sh:707-729` installs the Gateway API `standard-install.yaml` CRDs but no controller, no `GatewayClass`, and no `Gateway`, so `spec.gateway` is effectively a prod-only code path the Quick Start never exercises. The HTTPRoute E2E (`tests/e2e/keystone/httproute/`) compensates by applying the stub CRD from `00-httproute-crd.yaml` and patching `Accepted=True` onto the parent status manually — a fresh onboarding user has no such shortcut.

Concrete steps: (1) add a kind-only Envoy Gateway install — matches the `envoy-gateway-system` namespace already referenced by `operators/keystone/api/v1alpha1/keystone_webhook_test.go:1083,1537`, the network-policy E2E (`tests/e2e/keystone/network-policy/00-keystone-cr.yaml:38`), and the CRD reference example at `docs/reference/keystone-crd.md:476`. Deliver it as a `HelmRepository` + `HelmRelease` under `deploy/kind/base/` (new `envoy-gateway.yaml`) so Flux reconciles it the same way it reconciles cert-manager / MariaDB-operator / memcached-operator (`deploy/flux-system/releases/*.yaml`); keep the production overlay `deploy/flux-system/` untouched — same kind-only posture as Headlamp (`deploy/kind/base/headlamp.yaml`) and the OpenBao UI flip (CC-0082). (2) Ship a `GatewayClass/envoy` + `Gateway/openstack-gw` (namespace `openstack`, not `envoy-gateway-system`, so Keystone's `spec.gateway.parentRef` does not need a `ReferenceGrant` — the operator explicitly does not manage `ReferenceGrant`, see `keystone_types.go:348-354`) with a single HTTPS listener on `:443`, hostname `keystone.127-0-0-1.nip.io`, TLS terminated with a self-signed certificate issued by the existing `selfsigned-cluster-issuer` (already set up by the OpenBao chain). Use an `EnvoyProxy` CR to switch the proxy Service to `NodePort` with a fixed nodePort so `kind` can route it without `MetalLB`. Name the Gateway `openstack-gw` to line up with the E2E fixture at `tests/e2e/keystone/httproute/01-keystone-cr.yaml:40`. (3) Extend `hack/kind-config.yaml` with `extraPortMappings` mapping host `443` → the fixed Envoy NodePort so `https://keystone.127-0-0-1.nip.io/v3` (which resolves to `127.0.0.1` via the public nip.io wildcard) reaches the Envoy proxy on the kind container. (4) Add a `Step 2b` to `hack/deploy-infra.sh` — after the Gateway API CRDs install — that waits for the new Envoy Gateway `HelmRelease` and the `Gateway/openstack-gw` resource to report `Programmed=True`; extend the existing Step 4 HelmRelease wait list to include `envoy-gateway`. (5) Rewrite Quick Start Step 7 (`docs/quick-start.md:374-414`) so the sample `Keystone` CR includes `spec.gateway` with `parentRef.name: openstack-gw`, `hostname: keystone.127-0-0-1.nip.io`, `path: /`. (6) Rewrite the `## Access Keystone from your local machine` section (`docs/quick-start.md:498-546`) to drop the port-forward and document: export `OS_AUTH_URL=https://keystone.127-0-0-1.nip.io/v3`, decide whether to trust the self-signed CA or set `OS_INSECURE=true`, then `openstack token issue` works without any terminal-blocking port-forward and without any local DNS / hosts-file editing. (7) Add an `## Accept the self-signed certificate` subsection explaining how to extract the self-signed CA from the `cert-manager` `selfsigned-cluster-issuer` and add it to the local trust store, for users who do not want `-k` / `OS_INSECURE`. (8) Update the `docs/reference/infrastructure/e2e-deployment.md:104-108` diagram to insert the Envoy Gateway + `Gateway/openstack-gw` install between `Step 2a` (CRDs) and `Step 3` (base overlay). (9) Add a new E2E suite under `tests/e2e/keystone/gateway-quick-start/` (or extend `httproute/`) that deploys against a real Envoy Gateway on kind and asserts that `HTTPRoute.status.parents[0].conditions[type=Accepted]=True` arrives from the real controller — not the simulated patch step — so the Quick Start path is covered by CI.

**Motivation**: CC-0065 (#238) made `spec.gateway` a first-class feature but kept the Quick Start on the port-forward path. That leaves three concrete gaps for the on-ramp: (a) contributors who read the CRD reference (`docs/reference/keystone-crd.md:590-615`) and decide to try `spec.gateway` in kind will set it, watch `HTTPRouteReady=GatewayAPINotInstalled` stick, and have no documented way to install a Gateway controller on the kind cluster — the operator reports `GatewayAPINotInstalled` as a terminal state if the `HTTPRoute` CRD is present but no Gateway controller has claimed the parent, exactly the kind-cluster state today; (b) the `status.endpoint` field, which the operator derives as `https://{hostname}/v3` when `spec.gateway` is set, is a hollow promise in the Quick Start — it prints a URL that nothing resolves; (c) the current port-forward flow forces a second terminal window and then silently breaks the moment `openstack` CLI tries to resolve catalog endpoints, because the catalog contains cluster-internal URLs (`docs/quick-start.md:541-545`) — a real Gateway listener fixes both problems at once for `identity`-scope commands and sets up the story for follow-on OpenStack services landing under the same Gateway. Envoy Gateway is the natural choice because the project already assumes that namespace (`envoy-gateway-system`) in tests and reference docs, and because the upstream Envoy Gateway Helm chart supports `EnvoyProxy`-based NodePort exposure out of the box — no `MetalLB` dependency. `nip.io` is chosen over `/etc/hosts` entries because it needs zero local configuration: every developer hits the exact same hostname, CI gets the same hostname, the Quick Start has no "add this line to your hosts file as root" caveat, and the Keystone CR fixtures are byte-for-byte reproducible across machines. Shape and scope match CC-0082 (OpenBao UI, kind-only Flux release + Quick Start section) and CC-0086/#257 (flux-operator Web UI, kind-only Flux release + Quick Start section): one kind-only manifest, one Quick Start rewrite, one E2E suite extension; production overlay untouched.

**Affected Areas**:
- `deploy/kind/base/envoy-gateway.yaml` (new — `HelmRepository` + `HelmRelease` for the upstream `envoy-gateway` chart, pinned; `EnvoyProxy` CR switching the proxy Service to `NodePort` with a fixed nodePort so `extraPortMappings` can target it)
- `deploy/kind/base/openstack-gateway.yaml` (new — `GatewayClass/envoy` (controllerName `gateway.envoyproxy.io/gatewayclass-controller`), `Gateway/openstack-gw` in namespace `openstack` with an HTTPS listener on `:443` for hostname `keystone.127-0-0-1.nip.io`, `tls.mode: Terminate`, `certificateRefs` pointing at a `Certificate` issued by `selfsigned-cluster-issuer`)
- `deploy/kind/base/kustomization.yaml` (add the two new manifests to `resources` alongside `headlamp.yaml`; production overlay `deploy/flux-system/kustomization.yaml` stays unchanged)
- `deploy/flux-system/sources/` (new `envoy-gateway-charts.yaml` `HelmRepository` if the chart is not already reachable via `c5c3-charts`)
- `hack/kind-config.yaml` (add `extraPortMappings` for host `443` → container NodePort; the current file has an empty `nodes:` list so this is a single additive block)
- `hack/deploy-infra.sh` (new `Step 2b` that waits for `HelmRelease/envoy-gateway` Ready and `Gateway/openstack-gw` `Programmed=True`; extend the existing Step 4 HelmRelease wait list to include `envoy-gateway`)
- `docs/quick-start.md` — multiple edits: (a) Step 3 "What happens" table (`:103-116`) gets a `Step 2b` row for Envoy Gateway + Gateway install; (b) Step 7 sample CR (`:380-410`) gains `spec.gateway`; (c) the `## Access Keystone from your local machine` section (`:498-546`) is rewritten to use `https://keystone.127-0-0-1.nip.io/v3` directly — no `/etc/hosts` step, no port-forward; (d) a new subsection documents how to trust / bypass the self-signed TLS cert; (e) a short one-liner explains how `nip.io` works (public wildcard DNS that resolves `*.127-0-0-1.nip.io` to `127.0.0.1`, so no local DNS changes are required — the hostname is the same everywhere); (f) the architecture snapshot (`:141-155`) lists `envoy-gateway-system   envoy-gateway-*   Ready` alongside the existing controllers
- `docs/reference/infrastructure/e2e-deployment.md` (insert a Gateway-install block between the existing `Step 2a` and `Step 3` at `:104-108`)
- `docs/reference/keystone-crd.md:590-615` (add a kind-specific note pointing at the new Quick Start section and clarifying that `status.endpoint = https://{hostname}/v3` now actually resolves on a Quick Start cluster)
- `tests/e2e/keystone/gateway-quick-start/` (new Chainsaw suite that deploys a Keystone CR with `spec.gateway` against the real Envoy Gateway and asserts real `Accepted=True` from the controller, not a simulated patch)
- `renovate.json` (pin the Envoy Gateway chart version with a `customManagers` entry mirroring the `FLUX_OPERATOR_VERSION` pattern introduced by CC-0085, so chart bumps auto-PR)
- Nothing in `operators/keystone/` — `spec.gateway`, the webhook, the reconciler, and `HTTPRouteReady` are already in place from CC-0065

**Acceptance Criteria**:
- [ ] `deploy/kind/base/envoy-gateway.yaml` ships a `HelmRelease/envoy-gateway` in `envoy-gateway-system` that reaches `Ready=True` during `make deploy-infra` within the existing `HELMRELEASE_TIMEOUT` window
- [ ] `deploy/kind/base/openstack-gateway.yaml` ships a `GatewayClass/envoy` and a `Gateway/openstack-gw` in namespace `openstack`; on a fresh `make deploy-infra` run, `kubectl get gateway openstack-gw -n openstack -o jsonpath='{.status.conditions[?(@.type=="Programmed")].status}'` returns `True`
- [ ] `hack/kind-config.yaml` exposes the Envoy proxy NodePort on host `443` via `extraPortMappings` so `https://keystone.127-0-0-1.nip.io/v3` is reachable on the developer's machine with no further port-forward
- [ ] The Quick Start Step 7 sample CR sets `spec.gateway` with `parentRef.name: openstack-gw`, `hostname: keystone.127-0-0-1.nip.io`, `path: /`
- [ ] `status.endpoint` on the Quick Start CR reports `https://keystone.127-0-0-1.nip.io/v3` after reconciliation, matching `spec.gateway.hostname`
- [ ] `curl -k https://keystone.127-0-0-1.nip.io/v3` returns HTTP 200 with a `{"version": {"id": "v3", ...}}` JSON body on a freshly deployed kind cluster — no `/etc/hosts` edit required, no `kubectl port-forward` running
- [ ] With `OS_AUTH_URL=https://keystone.127-0-0-1.nip.io/v3`, `openstack token issue` succeeds — no `kubectl port-forward`, no DNS / hosts-file editing
- [ ] `docs/quick-start.md` `## Access Keystone from your local machine` section is rewritten and no longer references `kubectl port-forward svc/keystone-api`; the self-signed TLS handling (trust the CA or pass `-k` / `OS_INSECURE`) is documented inline; a short note explains the `nip.io` wildcard so readers understand why no local DNS config is needed
- [ ] `docs/reference/infrastructure/e2e-deployment.md` diagram (`:104-108` block) shows the Gateway install immediately after the Gateway API CRDs step
- [ ] `tests/e2e/keystone/gateway-quick-start/` asserts `HTTPRoute.status.parents[0].conditions[?(@.type=="Accepted")].status=True` arrives from the real Envoy Gateway controller within the Chainsaw timeout envelope, without any manual status patching
- [ ] `renovate.json` has a `customManagers` entry that pins the Envoy Gateway chart version with `minimumReleaseAge: "3 days"` and disabled major bumps, same pattern as `FLUX_OPERATOR_VERSION`
- [ ] `deploy/flux-system/kustomization.yaml`, `deploy/flux-system/fluxinstance.yaml`, and `deploy/flux-system/releases/*` are **not** modified — production overlay posture is unchanged
- [ ] The existing `kubectl port-forward svc/keystone-api` path still works for users on networks that block `nip.io` resolution; it is moved into a "Fallback" subsection but not deleted

**Non-Goals**:
- Installing Envoy Gateway (or any Gateway controller) in the production `deploy/flux-system/` overlay. The operator stays platform-agnostic — customers pick their own Gateway implementation, exactly as the CRD reference already documents (`docs/reference/keystone-crd.md:316-318`: *"The Gateway and GatewayClass are infrastructure concerns managed outside the operator"*). This feature is strictly a kind-overlay demo convenience, same posture as CC-0082 (OpenBao UI) and CC-0086 (Flux Web UI).
- Replacing the port-forward path for the rest of the Quick Start (Headlamp, OpenBao UI). Those stay on `kubectl port-forward` because they serve cluster-internal HTTP on non-443 ports; exposing them through the Gateway would require per-service HTTPRoute objects that do not match their current lifecycle.
- Cross-namespace `parentRef` support. `Gateway/openstack-gw` lives in the same `openstack` namespace as the Keystone CR to avoid `ReferenceGrant` — the operator explicitly does not manage `ReferenceGrant` (`keystone_types.go:348-354`: *"Cross-namespace references require a ReferenceGrant in the target namespace (out of scope for this operator)"*).
- Bringing up `MetalLB` in kind. Envoy Gateway's `EnvoyProxy` CR supports `NodePort` directly, which is the lighter-weight path and matches kind's single-node posture.
- Switching the E2E httproute suite from its simulated parent-status patch (`tests/e2e/keystone/httproute/chainsaw-test.yaml:99-100`) to the real controller. That suite exists to exercise the operator reconciler in isolation — a new suite under `gateway-quick-start/` covers the real-controller path instead.
- Editing `/etc/hosts` or shipping CoreDNS rewrite rules. The explicit design choice is public-wildcard DNS (`nip.io`) so the Quick Start has zero host-side DNS configuration. If `nip.io` is unreachable on a developer's network, the retained port-forward fallback covers that case.
- Wiring OIDC / SSO / real CAs on the kind Gateway. The Quick Start cluster is single-user localhost-only; self-signed is correct, matching the OpenBao UI's posture (`docs/quick-start.md:242-245`: *"Your browser will warn that the certificate is not trusted — this is expected for a kind cluster"*).

**References**:
- CC-0065 / #238 / commit `0b2c7d9` — the HTTPRoute sub-reconciler and `spec.gateway` field
- CRD reference for `spec.gateway`: `docs/reference/keystone-crd.md:590-636`
- Existing `spec.gateway` E2E fixture: `tests/e2e/keystone/httproute/01-keystone-cr.yaml`
- Envoy Gateway docs: https://gateway.envoyproxy.io/docs/
- Envoy Gateway `EnvoyProxy` NodePort pattern: https://gateway.envoyproxy.io/docs/tasks/operations/customize-envoyproxy/
- nip.io wildcard DNS: https://nip.io/
- Prior art for kind-only Flux releases: `deploy/kind/base/headlamp.yaml`, #246 (OpenBao UI, CC-0082), #257 (Flux Web UI, CC-0086)

**Labels:** enhancement

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/265


berendt:
> Reviewed by planwerk-review c600a8d with Claude Code



## CRITICAL (1)

### C-001: REQ-013 curl smoke check never executed in CI
**File**: `tests/e2e/infrastructure/gateway-quick-start-smoke/chainsaw-test.yaml:155-161` — **Fix**: ASK — **Confidence**: verified

**Problem**: REQ-013 requires `curl -k https://keystone.127-0-0-1.nip.io/v3` to return HTTP 200 with a Keystone v3 JSON body, and this is the only suite running that curl. The suite contains a Keystone-CRD presence guard that exits 0 (SKIP) when the CRD is absent. The smoke suite only runs in the `e2e-infra` job (.github/workflows/ci.yaml:444), and `deploy/kind/base/kustomization.yaml:106` patches the keystone-operator with `suspend: true`, so the Keystone CRD is never present in `e2e-infra`. The SKIP message points to `tests/e2e/keystone/gateway-quick-start/chainsaw-test.yaml`, but that suite has no curl step — only declarative status assertions. As a result, REQ-013 is documented but unverified by any CI run.

**Code**:
```
if ! kubectl get crd keystones.keystone.openstack.c5c3.io >/dev/null 2>&1; then
  echo "SKIP: Keystone CRD not present on this cluster"
  echo "      (operator not deployed — expected in the"
  echo "      e2e-infra job). The operator-path coverage lives"
  echo "      in tests/e2e/keystone/gateway-quick-start/."
  exit 0
fi
```

**Action Required**: Either (a) move/duplicate the smoke suite into `tests/e2e/keystone/gateway-quick-start-smoke/` so the operator-installed `e2e-operator` matrix job picks it up, or (b) install the keystone-operator after `make deploy-infra` in the `e2e-infra` job (matching .github/workflows/ci.yaml:638) and run the infrastructure suite. Option (a) keeps `e2e-infra` strictly infra-only and is simpler.

**Suggested Fix**:
```
Move the smoke suite to `tests/e2e/keystone/gateway-quick-start-smoke/chainsaw-test.yaml` so the `e2e-operator` matrix (which installs the keystone-operator) runs it; the CRD presence guard then succeeds and the curl step actually executes against `https://keystone.127-0-0-1.nip.io/v3`.
```

---

## WARNING (4)

### W-001: httproute test races real Envoy Gateway controller after CC-0088
**File**: `tests/e2e/keystone/httproute/chainsaw-test.yaml:99-100` — **Fix**: ASK — **Confidence**: likely

**Problem**: The httproute fixture was authored under the assumption that no Gateway controller runs in the kind cluster. CC-0088 introduces a real Envoy Gateway and a real `Gateway/openstack-gw`. The fixture uses `parentRef.name: openstack-gw` and patches `status.parents` via `kubectl patch --subresource=status --type=merge`. JSON merge-patch semantics replace the entire `parents` list rather than merging per controller slot, so the test entry overwrites the real controller's RouteParentStatus. The real controller will eventually re-add its slot (since the listener hostname `keystone.127-0-0-1.nip.io` does not match the fixture's `keystone.e2e.example.com`, it will write `Accepted=False, NoMatchingListenerHostname`). Steady state works, but a tight race between Step 5 and the next reconcile can briefly leave only the real `Accepted=False` entry, producing hard-to-attribute CI flakes.

**Code**:
```
kubectl patch httproute keystone-httproute-api -n $NAMESPACE \
  --subresource=status --type=merge -p '{
    "status": {
      "parents": [{
        "parentRef": {"name": "openstack-gw"},
        "controllerName": "e2e.example.com/fake-gateway-controller",
        ...
```

**Action Required**: Either (a) change `parentRef.name` to a value that resolves to no real Gateway (e.g. `not-a-real-gateway`) so the real controller never writes to this route's status, or (b) switch the patch to `--type=json` with an `add` op targeting `/status/parents/-` and assert the merged shape.

**Suggested Fix**:
```
Change the fixture's `parentRef.name` to a value that no real Gateway resolves (e.g. `not-a-real-gateway`) so the real Envoy Gateway controller never writes a RouteParentStatus slot for this HTTPRoute, eliminating the merge-patch race.
```

---

### W-002: overlay: kind label is unused — no chainsaw selector wired up
**File**: `tests/e2e/infrastructure/gateway-quick-start-smoke/chainsaw-test.yaml` — **Fix**: ASK — **Confidence**: verified

**Problem**: Both new chainsaw tests carry `metadata.labels.overlay: kind` and document the filter idiom `chainsaw test --selector 'overlay!=kind' tests/e2e/...`. Neither `Makefile:333` (`make e2e`) nor `.github/workflows/ci.yaml:{444,645}` invokes chainsaw with `--selector`. Production-context filtering relies entirely on the runtime presence guards (Gateway/openstack-gw, GatewayClass/envoy, Keystone CRD). The label and the selector framing are dead weight — they imply belt-and-braces protection that is not actually wired.

**Code**:
```
metadata:
  labels:
    overlay: kind
```

**Action Required**: Either remove the `overlay: kind` label and the `--selector` framing from both file headers, or wire `--selector "${CHAINSAW_SELECTOR:-}"` into `make e2e` and pin `CHAINSAW_SELECTOR=overlay!=kind` in any non-kind invocation. The phrasing should match what is actually wired.

**Suggested Fix**:
```
Remove `metadata.labels.overlay: kind` and the `--selector 'overlay!=kind'` documentation block from both new chainsaw test headers, since no Make target or CI workflow invokes chainsaw with `--selector`; the runtime presence guards already provide the production-context filtering the label was supposed to back up.
```

---

### W-003: Quick Start does not document host port 443 binding requirement
**File**: `docs/quick-start.md` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: `hack/kind-config.yaml:19-23` maps host `127.0.0.1:443` → containerPort `31443`. Binding host TCP 443 requires `CAP_NET_BIND_SERVICE` on Linux. Rootful Docker is fine; rootless Docker, Podman, or hosts with `net.ipv4.ip_unprivileged_port_start>=443` will fail at `kind create cluster` with a non-obvious bind error. Step 2 will work transparently for most users but break silently on rootless setups. The retained `kubectl port-forward` fallback section is the right escape hatch but is not pointed to from Step 2.

**Code**:
```
extraPortMappings:
  - containerPort: 31443
    hostPort: 443
    listenAddress: "127.0.0.1"
    protocol: TCP
```

**Action Required**: Add a short paragraph under Step 2 noting that the new `extraPortMappings` requires Docker to bind host TCP 443; users on rootless Docker / Podman / hosts with raised `ip_unprivileged_port_start` should use the `kubectl port-forward` fallback (link to the Fallback subsection).

**Suggested Fix**:
```
Add a paragraph under Step 2 of docs/quick-start.md: "The kind config binds host TCP port 443 (mapped to NodePort 31443). On Linux this requires `CAP_NET_BIND_SERVICE`, which is granted by default with rootful Docker. If you run rootless Docker or Podman, or your host sets `net.ipv4.ip_unprivileged_port_start` to 443 or higher, `kind create cluster` will fail with a bind error — use the `kubectl port-forward` fallback below instead." Cross-link to the existing Fallback subsection.
```

---

### W-004: Anonymous Docker Hub chart pull deferred without follow-up ticket link
**File**: `deploy/kind/base/envoy-gateway.yaml:86-88` — **Fix**: ASK — **Confidence**: verified

**Problem**: The HelmRelease pulls the envoy-gateway chart from `oci://docker.io/envoyproxy`, an anonymous Docker Hub source throttled to 100 pulls / 6h per source IP. CI runners behind shared NAT can surface TOOMANYREQUESTS from source-controller, flaking the envoy-gateway HelmRelease wait in `hack/deploy-infra.sh` Phase 3. The supply-chain header acknowledges this and defers mitigation to a future CC-ticket, but no ticket ID is referenced — the next reviewer will have to re-derive the rationale, and the smoke flake is likely to land in someone's PR within weeks.

**Code**:
```
Supply-chain note: `oci://docker.io/envoyproxy` is an anonymous Docker
Hub pull. Docker Hub throttles anonymous pulls to 100/6h per source IP,
so CI runners behind shared NAT can occasionally surface TOOMANYREQUESTS
from source-controller, causing the envoy-gateway HelmRelease wait in
`hack/deploy-infra.sh` Phase 3 to flake. Mitigation is deferred to a
future CC-ticket...
```

**Action Required**: File the follow-up ticket to mirror the chart into `oci://ghcr.io/c5c3/charts` (or equivalent) and replace the supply-chain comment's "future CC-ticket that mirrors..." wording with the concrete ticket ID.

**Suggested Fix**:
```
Open a follow-up ticket (e.g. CC-XXXX) covering chart mirroring to `oci://ghcr.io/c5c3/charts` and update the supply-chain note in `deploy/kind/base/envoy-gateway.yaml` to reference that ticket ID instead of the abstract "future CC-ticket".
```

---

## INFO (3)

### I-001: Redundant namespace creation in envoy-gateway HelmRelease
**File**: `deploy/kind/base/envoy-gateway.yaml:67-106` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Remove redundant namespace creation when namespace is already declared

**Problem**: Lines 67-70 declare `Namespace/envoy-gateway-system` while line 106 sets `install.createNamespace: true` on the HelmRelease. Both kustomize apply and Helm's pre-install hook can create the namespace, and `meta.helm.sh/release-namespace` annotations diverge between the two paths. The precedent file `deploy/kind/base/headlamp.yaml` does the same thing, so locally consistent — but the convention itself is suspect.

**Code**:
```
apiVersion: v1
kind: Namespace
metadata:
  name: envoy-gateway-system
...
  install:
    createNamespace: true
```

**Action Required**: Drop `install.createNamespace: true` from the HelmRelease since the Namespace YAML in the same file already creates it. Apply the same change to `headlamp.yaml` in a follow-up to align the convention.

**Suggested Fix**:
```
Remove the `install.createNamespace: true` line from the envoy-gateway HelmRelease at deploy/kind/base/envoy-gateway.yaml:106 because the Namespace resource at lines 67-70 in the same file already declares the namespace; open a follow-up to apply the same cleanup to deploy/kind/base/headlamp.yaml.
```

---

### I-002: Envoy Gateway version pin >=1.3.0 trails upstream stable
**File**: `deploy/kind/base/envoy-gateway.yaml` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: Envoy Gateway's 1.4 line shipped well before April 2026, so pinning `version: ">=1.3.0 <2.0.0"` makes Renovate's first auto-PR walk through several minor bumps. Not a defect — the packageRule already automerges minor/patch with a 3-day cooldown, so the pin self-corrects within days of merge.

**Code**:
```
version: ">=1.3.0 <2.0.0"
```

**Action Required**: Optional: bump the initial pin to the current stable minor before merge to avoid an avoidable Renovate noise burst, or leave it and let the automerge flow absorb the bumps.

**Suggested Fix**:
```
Optionally raise the initial lower bound to the current upstream stable minor (verify against the Envoy Gateway releases page) so Renovate does not emit a sequence of bump PRs immediately after merge; otherwise leave as-is and rely on the existing minor/patch automerge rule.
```

---

### I-003: CRD reference admonition omits /v3 suffix on Keystone endpoint
**File**: `docs/reference/keystone-crd.md:699` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The admonition references `https://keystone.127-0-0-1.nip.io`, but the canonical Quick Start endpoint (and the value `status.endpoint` resolves to) is `https://keystone.127-0-0-1.nip.io/v3`. Every other doc reference uses the `/v3` form.

**Code**:
```
> ... a ready-made `Gateway/openstack-gw` ships in the kind overlay
> (`deploy/kind/base/openstack-gateway.yaml`) and is reachable on the host
> at `https://keystone.127-0-0-1.nip.io` — see the
```

**Action Required**: Append `/v3` to the URL on line 699 so it matches the rest of the documentation and the actual Keystone endpoint.

**Suggested Fix**:
```
> ... a ready-made `Gateway/openstack-gw` ships in the kind overlay
> (`deploy/kind/base/openstack-gateway.yaml`) and is reachable on the host
> at `https://keystone.127-0-0-1.nip.io/v3` — see the
```

---

## Summary

Solid, well-scoped CC-0088 implementation that respects the kind-only / production-untouched invariant: clean two-overlay separation, mechanical production-posture enforcement, excellent decision-comments at every non-obvious choice, and real-controller assertions in the new gateway-quick-start chainsaw suite. The central concern is that REQ-013's curl smoke check is documented but never actually executed by CI because of a CRD presence guard that always trips in the only job that runs it. Secondary issues are a likely race between the pre-existing httproute fixture and the now-real Envoy Gateway controller, and an `overlay: kind` label/selector idiom that no Make target or workflow actually invokes.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 1 |
| WARNING  | 4 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Do not merge until REQ-013 is exercised by a CI run (Finding 1). The httproute race (Finding 2) and the unused `overlay: kind` selector framing (Finding 3) should be fixed in this PR or in an immediate fast-follow. The remaining warnings (port 443 docs, supply-chain ticket link) and the INFO items are low-risk and can land as follow-ups.

PR #270, opened 2026-04-23
feat(CC-0089): Add Prometheus metrics and ServiceMonitor sub-reconciler

berendt:
# GitHub Issue #122: Add Prometheus metrics and ServiceMonitor sub-reconciler

**Description**: Register custom Prometheus collectors in the Keystone operator: reconciliation duration histograms per sub-reconciler, reconciliation error counters by condition type, fernet/credential key rotation age gauges (last rotation timestamp), and database sync job duration/failure counters. Add a ServiceMonitor CR emitted by a sub-reconciler or Helm template so Prometheus auto-discovers the operator metrics endpoint on port 8080.

**Motivation**: The operator has a metrics port (8080) configured in the Helm chart but emits no custom metrics. For production operational visibility, operators need to monitor reconciliation health, detect degraded conditions, and alert on key rotation failures. Without metrics, operators are flying blind.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_secrets.go`
- `operators/keystone/helm/keystone-operator/templates/`

**Acceptance Criteria**:
- [ ] Reconciliation duration histogram registered per sub-reconciler
- [ ] Error counter by condition type incremented on failures
- [ ] Fernet/credential key age gauges expose last rotation timestamps
- [ ] ServiceMonitor template added to Helm chart (opt-in via values.yaml)
- [ ] Metrics endpoint verified accessible on port 8080
- [ ] Grafana dashboard JSON or mixin provided as documentation

---
Source: https://github.com/C5C3/forge/issues/122


berendt:



## WARNING (3)

### W-001: One-directional drift guard for subReconcilerConditionTypes
**File**: `operators/keystone/internal/controller/instrumentation.go:65-88` — **Fix**: ASK — **Confidence**: verified

**Problem**: The drift guard TestSubReconcilerConditionTypesCoversAllNames only enforces that map values are a subset of subConditionTypes. There is no test asserting that every instrumentSubReconciler call site in Reconcile has a corresponding key in subReconcilerConditionTypes. The fallback at line 85 silently uses an empty condition_type label. The doc comment at line 70-72 claims the drift-guard prevents this fallback from firing in practice, but the drift guard does not actually catch this direction. A future contributor who adds a new sub-reconciler call but forgets to add it to the map will silently emit reconcile_errors_total with an empty condition_type label, degrading alerting. TestReconcileEmitsDurationForEverySubReconciler only verifies the names already in the map.

**Code**:
```
if err != nil {
    instrumentRecordError(name, subReconcilerConditionTypes[name])
}
```

**Action Required**: Either fail loudly when a name is missing from the map (panic on missing key in test mode, or log a one-time warning in production), or extend the drift guard to walk the AST/source for instrumentSubReconciler call literals and verify their first-argument string is in the map. Even the simpler fix — emitting condition_type="UNKNOWN" instead of "" — would be visible in alerts.

**Suggested Fix**:
```
Either fail loudly when a name is missing from the map (panic on missing key in test mode, or log a one-time warning in production), or extend the drift guard to walk the AST/source for instrumentSubReconciler call literals and verify their first-argument string is in the map. Even the simpler fix — emitting condition_type="UNKNOWN" instead of "" — would be visible in alerts.
```

---

### W-002: db_sync metrics omit upgrade-phase and schema-check Jobs
**File**: `operators/keystone/internal/controller/reconcile_database.go:294-299` — **Fix**: ASK — **Confidence**: verified

**Problem**: recordDBSyncTerminalState is called only in the non-upgrade path. The expand/migrate/contract Jobs (reconcileExpand, reconcileMigrate, reconcileContract) and the schema-check Job (buildSchemaCheckJob) all produce terminal states with the same operational impact (db schema mutations), but emit no metric. During an active upgrade the dashboard's 'db_sync duration p95 and failure rate' panel goes blank. The doc says 'db_sync Job outcomes' which is technically narrow, but operators alerting on keystone_operator_db_sync_total{result="failed"} will silently miss every expand/migrate/contract failure.

**Code**:
```
done, err := database.RunDBSyncJob(ctx, r.Client, r.Scheme, keystone, buildDBSyncJob(keystone, configMapName))
// Emit db_sync metrics on the terminal-transition observation path
// regardless of (done, err)...
r.recordDBSyncTerminalState(ctx, keystone)
```

**Action Required**: Either (a) call recordDBSyncTerminalState for each phase Job (it already keys by Job name <keystone>-db-sync — generalize to take a job-name arg and persist a per-phase last-UID annotation), or (b) explicitly document the limitation in the metrics reference doc and the runbook so operators know to alert on the upgrade phase via DatabaseReady condition transitions instead.

**Suggested Fix**:
```
Either (a) extend recordDBSyncTerminalState to take a job-name argument and call it for each phase Job (expand/migrate/contract/schema-check), persisting a per-phase last-UID annotation; or (b) document in docs/reference/keystone-operator-metrics.md and the runbook that db_sync metrics cover only the standard db-sync Job and operators must alert on DatabaseReady condition transitions for upgrade-phase failures.
```

**Related**: W-003

---

### W-003: recordDBSyncTerminalState emits at-most-once only when patch succeeds
**File**: `operators/keystone/internal/controller/reconcile_database.go:142-178` — **Fix**: ASK — **Confidence**: verified

**Problem**: The function logs at V(1) and returns early when the annotation patch fails. By that point the metric has already been emitted via metrics.RecordDBSync at line 151. If the patch fails (transient apiserver error, conflict on a slow reconcile), the next reconcile will re-emit the metric for the same Job. The single-emit guarantee in the metrics reference doc (docs/reference/keystone-operator-metrics.md:170-172) overstates this — in practice it is at-most-one increment per successful annotation persist. For operators alerting on absolute counter rates this is acceptable (inflation bounded by patch retries within a Job's lifetime); for those summing over windows, it is a small skew during apiserver instability windows.

**Code**:
```
if err := r.Patch(ctx, patchTarget, patch); err != nil {
    log.FromContext(ctx).V(1).Info(
        "recordDBSyncTerminalState: patching last-observed Job UID failed; "+
            "next reconcile may re-emit the db_sync metric",
        ...
    )
    return
}
```

**Action Required**: Persist the annotation before emitting the metric so a patch failure does not double-count. Alternatively, soften the doc to say 'at most once per (Job UID, successful annotation persist) — transient apiserver failures may cause a single re-emit on the next reconcile'.

**Suggested Fix**:
```
Reorder the function so the annotation patch persists before metrics.RecordDBSync is called: attempt the Patch first, return on failure with a deferred-emission log, and only call metrics.RecordDBSync(keystone.Name, keystone.Namespace, result, duration) after the patch succeeds. Alternatively, update docs/reference/keystone-operator-metrics.md:170-172 to document that transient apiserver failures may cause a single re-emit.
```

**Related**: W-002

---

## INFO (3)

### I-001: Integration test couples to current Update semantics via Immutable Secret
**File**: `operators/keystone/internal/controller/integration_test.go:3870-3917` — **Fix**: ASK — **Confidence**: likely

**Problem**: TestReconcileErrorsTotalIncrementsOnInducedFailure induces a SecretsReady error by pre-creating a destination Secret as Immutable=true with placeholder data, banking on reconcileDBConnectionSecret calling Update (rejected by the apiserver) rather than Create. If a future refactor migrates that sub-reconciler to Server-Side Apply, the test will silently start passing for the wrong reason (no error induced). The doc comment at line 3879-3886 acknowledges the coupling but doesn't suggest an alternative.

**Code**:
```
The test induces a SecretsReady error by pre-creating a destination Secret as Immutable=true with placeholder data, banking on reconcileDBConnectionSecret calling Update (rejected by the apiserver) rather than Create.
```

**Action Required**: Add a comment near the assertion stating the test expects the Update path, and consider rewiring via an interceptor (interceptor.NewClient) that reliably injects an error from reconcileDBConnectionSecret regardless of the underlying API operation — the rest of the controller test suite already uses this pattern (testReconciler with interceptor.Funcs).

**Suggested Fix**:
```
Refactor the test to use interceptor.NewClient with interceptor.Funcs to inject a deterministic error from reconcileDBConnectionSecret regardless of whether the controller calls Update or Apply. This matches the pattern used elsewhere in the controller test suite (testReconciler) and decouples the test from the underlying API operation.
```

---

### I-002: Helm serviceMonitor.interval regex rejects the literal 0
**File**: `operators/keystone/helm/keystone-operator/values.schema.json:162` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The pattern '^([0-9]+(ns|us|µs|ms|s|m|h))+$' rejects '0'. Prometheus operator's Duration type accepts '0' to mean 'use the global default interval', but this regex requires a unit suffix. This is unlikely to come up in practice but limits flexibility. The current pattern correctly forbids decimals, which matches Prometheus operator's Duration regex.

**Code**:
```
"pattern": "^([0-9]+(ns|us|µs|ms|s|m|h))+$"
```

**Action Required**: Update the regex to also permit a bare '0' (e.g. '^(0|([0-9]+(ns|us|µs|ms|s|m|h))+)$') so users can opt into the global default scrape interval.

**Suggested Fix**:
```
"pattern": "^(0|([0-9]+(ns|us|µs|ms|s|m|h))+)$"
```

---

### I-003: Metrics endpoint serves plain HTTP without authentication or TLS
**File**: `internal/common/bootstrap/manager.go:128-130` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The bootstrap sets only metricsserver.Options{BindAddress: ":8080"}; controller-runtime's default is plain HTTP without auth. The how-to doc (docs/how-to/enable-keystone-operator-metrics.md) doesn't mention this — Prometheus picks it up correctly because the scheme defaults to http, but operators with stricter cluster policies (network policies blocking plain HTTP, mTLS service mesh) need to know. The PR is consistent with the existing setup; this is a documentation gap, not a regression.

**Code**:
```
metricsserver.Options{BindAddress: ":8080"}
```

**Action Required**: Add a short 'Security & network considerations' section to docs/how-to/enable-keystone-operator-metrics.md noting that the endpoint serves plain HTTP on port 8080 (no auth, no TLS) and is intended for cluster-internal scraping; consumers needing TLS should front the operator with a sidecar or update the bootstrap to use SecureServing: true with a cert.

**Suggested Fix**:
```
Add a 'Security & network considerations' section to docs/how-to/enable-keystone-operator-metrics.md stating: the metrics endpoint serves plain HTTP on port 8080 with no authentication or TLS, intended for cluster-internal scraping only. Consumers requiring TLS should front the operator with a sidecar or enable SecureServing in the manager bootstrap with a cert.
```

---

## Summary

Well-engineered, thoroughly-tested Prometheus instrumentation layer with proper cardinality bounds, per-CR lifecycle cleanup, and bidirectional drift guards between collectors, reference doc, and dashboard. Three WARNING-level quality issues stand out: the drift guard for subReconcilerConditionTypes only checks one direction (silent empty-label emission risk), db_sync metrics miss expand/migrate/contract and schema-check Jobs (blank dashboard during upgrades), and recordDBSyncTerminalState emits the metric before persisting the dedupe annotation (possible double-count on patch failure). Test depth, comments on subtle invariants, and architectural choices (collectors struct + sync.Once, TestRecorder indirection) are all high quality.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Merge after addressing the three WARNINGs — most importantly Finding 2 (db_sync coverage of upgrade-phase Jobs) which needs either a code extension or an explicit documentation carve-out so operators do not silently miss upgrade-time failures. Finding 1 has a cheap fix that closes a real future-proofing gap, and Finding 3 should either reorder the patch/emit sequence or soften the doc claim. The INFO findings are optional polish.

operators/keystone/internal/metrics/cardinality_test.go
gndrmnn, line 1: Move this test to `collectors_test.go`

operators/keystone/internal/metrics/collectors.go
```diff
@@ -0,0 +1,204 @@
...
+func newCollectorsForTest(reg prometheus.Registerer) *collectors {
+	c := newCollectors()
+	if err := c.register(reg); err != nil {
+		// Test registries are always empty in new Prometheus registries;
+		// a registration error here is a programmer bug in the test
+		// setup and must be surfaced loudly.
+		panic(fmt.Sprintf("metrics: test registry rejected collectors: %v", err))
+	}
+	return c
+}
```
gndrmnn: If this constructor is test-only, move it to the test file `collectors_test.go`

operators/keystone/internal/metrics/collectors.go
```diff
@@ -0,0 +1,204 @@
...
+	c.dbSyncTotal.WithLabelValues(keystone, namespace, result).Inc()
+	c.dbSyncDuration.WithLabelValues(keystone, namespace).Observe(duration.Seconds())
+}
+
+func (c *collectors) deleteForKeystone(name, namespace string) {
+	labels := prometheus.Labels{"keystone": name, "namespace": namespace}
+	c.keyRotationAge.DeletePartialMatch(labels)
+	c.dbSyncTotal.DeletePartialMatch(labels)
+	c.dbSyncDuration.DeletePartialMatch(labels)
+}
```
gndrmnn: If your internal methods are solely used in tests, move their definition to the test file `collectors_test.go`.

operators/keystone/internal/metrics/contract_test.go
gndrmnn, line 1: No, we are not going to unit test documentation against code. Remove `contract_test.go`

operators/keystone/internal/controller/instrumentation_test.go
```diff
@@ -207,3 +258,176 @@ func TestSubReconcilerConditionTypesCoversAllNames(t *testing.T) {
...
+	bl, ok := expr.(*ast.BasicLit)
+	if !ok || bl.Kind != token.STRING {
+		return "", false
+	}
+	v, err := strconv.Unquote(bl.Value)
+	if err != nil {
+		return "", false
+	}
+	return v, true
+}
```
gndrmnn: We are *absolutely* not going to test source code via literal string analysis. This is not acceptable remove lines 262-432

gndrmnn (changes requested):

gndrmnn (approved):

Issue #271, opened 2026-04-23
Race between operator finalizer Delete and ESO PushSecret adoption leaks Fernet and credential keys in OpenBao


berendt:
**Category**: bug | **Scope**: Medium

**Description**: The openbao-finalizer introduced in CC-0079 (PR #260, commit 09029b9) has a race with ESO's PushSecret adoption that can leak Fernet and credential key material in OpenBao KV-v2. `finalizeOpenBaoSecrets` (`operators/keystone/internal/controller/reconcile_secrets.go:185-197`) calls `r.Delete` on `<keystone>-fernet-keys-backup` and `<keystone>-credential-keys-backup` without first confirming that ESO has adopted the PushSecret by installing its own finalizer (`pushsecret.externalsecrets.external-secrets.io/finalizer`). If the operator's Delete races ahead of ESO's first reconcile — for example because the ESO PushSecret workqueue is backed up under parallel load — Kubernetes removes the PushSecret object immediately (no finalizer to hold it) and ESO never reaches the `DeletionPolicy=Delete` branch that would issue `DeleteSecret` against the remote. The KV-v2 paths `kv-v2/openstack/keystone/fernet-keys` and `kv-v2/openstack/keystone/credential-keys` remain populated, carry `custom_metadata: managed-by=external-secrets` but no `deletion_time`, and — because there is no PushSecret left to reference them — stay orphaned until manual intervention.

**Motivation**: Reproduced on `main` in CI run [24842115250/job 72722485985](https://github.com/C5C3/forge/actions/runs/24842115250/job/72722485985) (commit 4526cd7). The `keystone-deletion-cleanup` E2E step-6 asserts both KV-v2 paths are `not-found` after PushSecret deletion; both still contained `version: 1` data with `created_time` `2026-04-23T15:27:51.982Z` (Fernet) and `2026-04-23T15:27:52.089Z` (credential) — i.e., ESO's first push landed 13 s AFTER `Ready=True` (`15:27:38`) and only ~1.6 s before the PushSecret was garbage-collected at `15:27:53.6`. The immediately preceding CI run [24832475387](https://github.com/C5C3/forge/actions/runs/24832475387) on the PR merge commit 09029b9 was green — only a `golang:1.25` Docker-digest bump (#261) lay between the two runs — confirming this is a timing-sensitive race, not a regression from the digest bump.

[...]

PR #276, opened 2026-04-24
feat(CC-0091): Fix finalizer/PushSecret race leaking keys in OpenBao

berendt:
# GitHub Issue #271: Race between operator finalizer Delete and ESO PushSecret adoption leaks Fernet and credential keys in OpenBao

**Category**: bug | **Scope**: Medium

**Description**: The openbao-finalizer introduced in CC-0079 (PR #260, commit 09029b9) has a race with ESO's PushSecret adoption that can leak Fernet and credential key material in OpenBao KV-v2. `finalizeOpenBaoSecrets` (`operators/keystone/internal/controller/reconcile_secrets.go:185-197`) calls `r.Delete` on `<keystone>-fernet-keys-backup` and `<keystone>-credential-keys-backup` without first confirming that ESO has adopted the PushSecret by installing its own finalizer (`pushsecret.externalsecrets.external-secrets.io/finalizer`). If the operator's Delete races ahead of ESO's first reconcile — for example because the ESO PushSecret workqueue is backed up under parallel load — Kubernetes removes the PushSecret object immediately (no finalizer to hold it) and ESO never reaches the `DeletionPolicy=Delete` branch that would issue `DeleteSecret` against the remote. The KV-v2 paths `kv-v2/openstack/keystone/fernet-keys` and `kv-v2/openstack/keystone/credential-keys` remain populated, carry `custom_metadata: managed-by=external-secrets` but no `deletion_time`, and — because there is no PushSecret left to reference them — stay orphaned until manual intervention.

**Motivation**: Reproduced on `main` in CI run [24842115250/job 72722485985](https://github.com/C5C3/forge/actions/runs/24842115250/job/72722485985) (commit 4526cd7). The `keystone-deletion-cleanup` E2E step-6 asserts both KV-v2 paths are `not-found` after PushSecret deletion; both still contained `version: 1` data with `created_time` `2026-04-23T15:27:51.982Z` (Fernet) and `2026-04-23T15:27:52.089Z` (credential) — i.e., ESO's first push landed 13 s AFTER `Ready=True` (`15:27:38`) and only ~1.6 s before the PushSecret was garbage-collected at `15:27:53.6`. The immediately preceding CI run [24832475387](https://github.com/C5C3/forge/actions/runs/24832475387) on the PR merge commit 09029b9 was green — only a `golang:1.25` Docker-digest bump (#261) lay between the two runs — confirming this is a timing-sensitive race, not a regression from the digest bump.

The blast radius is not limited to the E2E test: in production, a user who creates a Keystone CR and deletes it within seconds (CI automation, GitOps apply-then-revert, misconfigured HelmRelease, prematurely-cancelled CR) would leak Fernet tokens and credential-encryption keys in OpenBao — violating REQ-010 (CC-0079) and the Deletion Cleanup guarantee documented in `docs/reference/keystone-reconciler.md`. A compromised KV read path would then expose historical token-signing material that the cluster administrator believes was purged.

Both Fernet and credential paths share the same defect and must be fixed symmetrically:

- **Fernet**: `fernetKeysPushSecret` (`reconcile_fernet.go:403-429`) targets `RemoteKey: openstack/keystone/fernet-keys`. Leak surface: active Fernet signing keys used to mint Keystone tokens; stale copies enable offline token forgery if the KV backend is later compromised.
- **Credential**: `credentialKeysPushSecret` (`reconcile_credential.go:409-435`) targets `RemoteKey: openstack/keystone/credential-keys`. Leak surface: credential-encryption keys used to protect OpenStack service credentials (EC2 access-keys, application credentials) at rest in the Keystone DB; stale copies enable offline decryption of any persisted credentials.

Both builders construct PushSecrets with identical shape — same `DeletionPolicy=Delete`, same `ClusterSecretStoreRef=openbao-cluster-store`, same `EnsurePushSecret` call site via `internal/common/secrets/secrets.go:128` — so one code path fix covers both resources. Neither `reconcileFernetKeys` nor `reconcileCredentialKeys` gates its `*Ready` condition on the PushSecret's own `status.conditions[type=Ready]`, so there is no upstream readiness gate that naturally prevents a user's Delete from overtaking ESO's first push.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_secrets.go` (`finalizeOpenBaoSecrets`, `openBaoBackupPushSecretNames`, `setOpenBaoFinalizerBlockedCondition`)
- `operators/keystone/internal/controller/reconcile_fernet.go` (`fernetKeysPushSecret`, step 7 of `reconcileFernetKeys`)
- `operators/keystone/internal/controller/reconcile_credential.go` (`credentialKeysPushSecret`, the symmetric step in `reconcileCredentialKeys`)
- `operators/keystone/internal/controller/keystone_controller.go` (`reconcileDeleteOpenBao` — new blocked reason surfaces here)
- `operators/keystone/internal/controller/integration_test.go` (deletion-cleanup cases currently assume ESO finalizer is always present when the handler runs)
- `operators/keystone/internal/controller/reconcile_secrets_test.go` (new unit cases for the adoption wait)
- `tests/e2e/keystone/deletion-cleanup/chainsaw-test.yaml` (step-6 needs a bounded retry to tolerate residual ESO `DeleteSecret` latency after step-4 observes PushSecret gone)
- `docs/reference/keystone-reconciler.md` (document the adoption wait + the new `WaitingForESOAdoption` reason)

**Acceptance Criteria**:
- [ ] `finalizeOpenBaoSecrets` performs a `Get` on each backup PushSecret before issuing `Delete`; for any PushSecret whose `DeletionTimestamp` is zero AND whose finalizer list does not contain `pushsecret.externalsecrets.external-secrets.io/finalizer`, the handler records `SecretsReady=False/OpenBaoFinalizerBlocked` with reason `WaitingForESOAdoption` (distinct from the existing `OpenBaoFinalizerBlocked` reason for post-Delete Terminating waits) and returns `(done=false, nil)` without issuing Delete.
- [ ] Once ESO adoption is observed (ESO finalizer present OR `DeletionTimestamp` already set on the PushSecret), the handler issues Delete exactly as today and Pass-2 semantics (`Get` NotFound → success) remain unchanged.
- [ ] Adoption check covers both `fernet-keys-backup` and `credential-keys-backup` via the existing `openBaoBackupPushSecretNames` iteration — no duplication per resource.
- [ ] New unit tests in `reconcile_secrets_test.go`:
  - PushSecret without ESO finalizer, no DeletionTimestamp → `finalizeOpenBaoSecrets` returns `(false, nil)`, blocked condition reason = `WaitingForESOAdoption`, no Delete call issued (verified via fake-client action log).
  - Same PushSecret with ESO finalizer added → Delete is issued once per call; Get-NotFound on second pass yields `(true, nil)`.
  - Mixed state (Fernet adopted, credential not) → blocked condition names the unadopted resource, no Delete issued on either; subsequent pass after adoption completes both.
- [ ] New integration test in `integration_test.go` covers the "Keystone CR deleted before ESO adopts PushSecret" path, verifying that (a) Keystone stays in Terminating with `SecretsReady=False/OpenBaoFinalizerBlocked/WaitingForESOAdoption`, (b) once the test harness adds the ESO finalizer to both backup PushSecrets the operator proceeds to Delete, (c) Keystone eventually finalizes, (d) the envtest-backed PushSecret objects reach `NotFound`.
- [ ] Chainsaw step-6 (`tests/e2e/keystone/deletion-cleanup/chainsaw-test.yaml:217-252`) replaces the one-shot `bao kv get` with a bounded retry loop (e.g., 30 s total, 3 s interval) so residual `DeleteSecret` latency between step-4 (PushSecret gone) and the OpenBao check does not re-flake — the operator fix plus a tolerant assertion together.
- [ ] `docs/reference/keystone-reconciler.md` documents: (i) why the adoption wait exists (ESO PushSecret workqueue race), (ii) the new `WaitingForESOAdoption` reason alongside the existing blocked reason, (iii) the two-stage finalizer lifecycle (wait-for-adoption → Delete → wait-for-gone), (iv) that both Fernet and credential PushSecrets are covered by the same handler.
- [ ] No expansion of the RBAC surface — `delete` on `pushsecrets` remains the only extra verb already granted by #260. Adding `get` on `pushsecrets` is already covered by the existing rule.
- [ ] CI `chainsaw/keystone-deletion-cleanup` is green across 10 consecutive runs (manual `gh workflow run` trigger) to confirm flake resolution.

**Labels:** bug

**Labels:** bug

---
Source: https://github.com/C5C3/forge/issues/271


berendt:
> Reviewed by planwerk-review 889e22e with Claude Code



## WARNING (2)

### W-001: FinalizingOpenBaoSecrets event now fires on every requeue during Pass-0 adoption wait (regresses CC-0079 exactly-once semantics)
**File**: `operators/keystone/internal/controller/keystone_controller.go:369-381` — **Fix**: ASK — **Confidence**: verified

**Problem**: The pre-existing exactly-once semantics per termination relied on Pass-1 immediately flipping the PushSecret to Terminating so hasLiveOpenBaoBackupPushSecrets would return false on subsequent requeues. Pass-0 breaks that invariant: while waiting for ESO adoption, the PushSecret stays non-Terminating, so hasLiveOpenBaoBackupPushSecrets returns true on every 15s RequeueSecretPolling requeue and the reconciler keeps emitting the FinalizingOpenBaoSecrets Normal Event. Because SetupWithManager does not watch PushSecret, this is a steady ~1-event-per-15s stream until ESO adopts — the opposite of the CC-0079 exactly-once guarantee documented by the nearby comment.

**Code**:
```
	hasLiveCleanupWork, err := r.hasLiveOpenBaoBackupPushSecrets(ctx, keystone)
	if err != nil {
		return ctrl.Result{}, err
	}
	if hasLiveCleanupWork {
		r.Recorder.Event(keystone, corev1.EventTypeNormal, "FinalizingOpenBaoSecrets",
			"Cleaning up OpenBao backup PushSecrets before removing Keystone")
	}
```

**Action Required**: Gate the FinalizingOpenBaoSecrets emit on Pass-0 being cleared: either move the emit into finalizeOpenBaoSecrets right before Pass-1, or teach hasLiveOpenBaoBackupPushSecrets to treat unadopted PushSecrets like already-Terminating for emission purposes. Add an integration-test assertion in TestIntegrationKeystone_DeleteRacingESOAdoption stage 1 that counts FinalizingOpenBaoSecrets events and expects ≤ 1 across the adoption-wait window.

**Suggested Fix**:
```
Compute adoption state in the same pass used for hasLiveCleanupWork and suppress the FinalizingOpenBaoSecrets emit while any backup PushSecret is still unadopted (Pass-0 not yet cleared). Either fold the adoption check into hasLiveOpenBaoBackupPushSecrets so it returns false while adoption is pending, or relocate the Recorder.Event call into finalizeOpenBaoSecrets immediately before Pass-1 fires. Also add an integration-test assertion counting FinalizingOpenBaoSecrets events in TestIntegrationKeystone_DeleteRacingESOAdoption stage 1 with an upper bound of 1.
```

**Related**: I-001

---

### W-002: Misleading doc comment on addESOAdoptionFinalizerToPushSecret test helper
**File**: `operators/keystone/internal/controller/integration_test.go:2936-2950` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The helper's doc comment claims that client.Delete on a PushSecret holding only esoPushSecretFinalizer will garbage-collect the object immediately rather than flip it into Terminating. That is false — both the controller-runtime fake client and real Kubernetes treat any non-empty metadata.finalizers as blocking; the specific finalizer string is not interpreted. The test still passes because teardown calls clearESOFinalizerFromPushSecret, but the comment will mislead future maintainers who copy-paste the helper expecting immediate GC.

**Code**:
```
// addESOAdoptionFinalizerToPushSecret attaches only esoPushSecretFinalizer —
// the adoption signal. Without esoCleanupFinalizer, a client.Delete on the
// PushSecret will garbage-collect it immediately rather than flipping it into
// Terminating. Use this helper to simulate ESO adopting a PushSecret that is
// not expected to hold up deletion — e.g. credential-keys-backup in the
// OpenBaoFinalizerBlocked test, which must satisfy Pass-0 without also
// simulating a stuck kv-v2 purge (CC-0091, REQ-001, REQ-003, REQ-007).
```

**Action Required**: Rewrite the comment to describe what the helper actually models: a PushSecret that ESO has adopted (Pass-0 passes) but whose Delete flow is not being used to simulate a slow remote DeleteSecret. Make clear that any finalizer still blocks etcd removal.

**Suggested Fix**:
```
// addESOAdoptionFinalizerToPushSecret attaches only esoPushSecretFinalizer —
// the adoption signal. This simulates the narrow window after ESO has
// adopted a PushSecret (first reconcile has installed the adoption
// finalizer) but before a client.Delete has fired. Any finalizer still
// blocks etcd removal, so a subsequent Delete flips the object into
// Terminating exactly as it does with addESOFinalizerToPushSecret — use
// this helper when the test only needs to satisfy Pass-0 and does not
// care about modelling ESO's kv-v2 purge latency (CC-0091, REQ-001,
// REQ-003, REQ-007).
```

---

## INFO (3)

### I-001: Controller does not watch PushSecret, so Pass-0 adoption wait adds up to 15s of deletion latency
**File**: `operators/keystone/internal/controller/keystone_controller.go:567-604` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: EnsurePushSecret (internal/common/secrets/secrets.go:133) sets a controller reference so Keystone owns PushSecret, but SetupWithManager does not declare Owns(&esov1alpha1.PushSecret{}). When ESO installs its adoption finalizer on the PushSecret, the controller does not wake up and must wait for the next RequeueSecretPolling (15s) tick before advancing past Pass-0. Correctness is unaffected but deletion takes up to 15s longer per unadopted PushSecret on the first reconcile after a fast-deleted CR.

**Code**:
```
		Owns(&batchv1.CronJob{}).
```

**Action Required**: Add Owns(&esov1alpha1.PushSecret{}) to the builder chain in SetupWithManager so the adoption-wait latency collapses from up to 15s to watch-event delivery latency (single-digit ms). Safe as a one-liner follow-up PR to keep this fix minimal.

**Suggested Fix**:
```
		Owns(&batchv1.CronJob{}).
		Owns(&esov1alpha1.PushSecret{}).
```

**Related**: W-001

---

### I-002: Pass-0 Get is redundant with Pass-2 Get on the happy path
**File**: `operators/keystone/internal/controller/reconcile_secrets.go:207-270` — **Fix**: ASK — **Confidence**: verified

**Problem**: Each reconcile now issues two Get calls per backup PushSecret (one in Pass-0 at lines 207-231, one in Pass-2 at lines 253-270). On the common happy path — both PushSecrets adopted, Pass-1 fires Delete, Pass-2 still observes Terminating — this is 4 Gets + 2 Deletes per reconcile, double the pre-CC-0091 API call count. Minor cluster API-server load cost, not a correctness issue, but worth noting for throughput planning on clusters with many Keystone CRs churning.

**Code**:
```
Pass-0 Get at reconcile_secrets.go:207-231 and Pass-2 Get at reconcile_secrets.go:253-270 both fetch the same PushSecret per name.
```

**Action Required**: Consider having Pass-0 cache the fetched PushSecret object state into a local map keyed by name so Pass-2 can skip the re-Get on the happy path. This is arguably over-engineering for a 2-element iteration, so defer unless cluster-wide throughput becomes a concern.

**Suggested Fix**:
```
Introduce a small intra-reconcile cache (local map[string]*esov1alpha1.PushSecret) populated in Pass-0 and consulted in Pass-2, with fallthrough to Get on cache miss. Only apply if throughput profiling later justifies it — the 2-element loop makes this premature optimization today.
```

---

### I-003: TestFinalizeOpenBaoSecrets_MixedAdoptionState uses BeNumerically(">=", 2) where Equal(2) is expected
**File**: `operators/keystone/internal/controller/reconcile_secrets_test.go:1299-1300` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The assertion BeNumerically(">=", 2) is looser than the contract requires. Pass-1 iterates over a bounded 2-name list and must issue exactly one Delete per adopted PushSecret once Pass-0 clears — a future regression that accidentally issues an extra Delete (e.g., a retry inside a helper) would slip past this assertion unnoticed.

**Code**:
```
	g.Expect(deleteCount).To(BeNumerically(">=", 2), ...)
```

**Action Required**: Tighten the assertion to Equal(2) so that any extra Delete call is caught.

**Suggested Fix**:
```
	g.Expect(deleteCount).To(Equal(2),
		"Pass-1 must fire Delete exactly once on every adopted PushSecret once the Pass-0 gate opens")
```

---

## Summary

A well-scoped, well-motivated fix for the CC-0091 delete race: the three-pass design is clean, the symmetric fernet/credential coverage via a single openBaoBackupPushSecretNames loop is consistent with the existing CC-0079 pattern, and the WaitingForESOAdoption vs. OpenBaoFinalizerBlocked split gives SREs the right signal. Test coverage is unusually thorough (unit tests per branch plus a full envtest integration test across all three lifecycle stages) and the docs update accurately references the motivating CI run. The main concern is that the CC-0079 exactly-once event-emission contract was not reconsidered for Pass-0 and now regresses to a ~1-event-per-15s stream during the very race this PR targets; the misleading adoption-finalizer helper comment is a smaller but real documentation defect. Remaining findings (missing PushSecret watch, redundant Get, loose numeric assertion) are follow-up opportunities rather than blockers.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 2 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Address Finding 1 (event-emission regression) and Finding 2 (misleading helper comment) before merging. The INFO findings (missing PushSecret watch, redundant Get, tightening the deleteCount assertion) can land as follow-up PRs and should not block this change.

Issue #273, opened 2026-04-23
Watch PushSecret in Keystone reconciler to remove 15s deletion-cleanup latency


berendt:
**Category**: enhancement | **Scope**: Small

**Description**: `KeystoneReconciler.SetupWithManager` (`operators/keystone/internal/controller/keystone_controller.go:582-604`) currently watches `corev1.Secret`, `mariadbv1alpha1.MariaDB`, and `esov1.ClusterSecretStore` but **not** `esov1alpha1.PushSecret`. As a result, the openbao-finalizer cleanup added in CC-0079/CC-0091 only re-evaluates Pass-0 (adoption gate) and Pass-2 (garbage-collection wait) at the next periodic requeue — `RequeueSecretPolling = 15 * time.Second` (`operators/keystone/internal/controller/requeue_intervals.go:22`). Each PushSecret state transition during teardown — ESO installing its cleanup finalizer, ESO releasing it after the kv-v2 purge, the API server garbage-collecting the Terminating object — therefore costs **up to 15 s** of avoidable wait, and a worst-case Keystone deletion accumulates ~30 s of pure polling latency across the three state changes the operator has to observe.

**Motivation**: Surfaced while diagnosing CI run [24852835587](https://github.com/C5C3/forge/actions/runs/24852835587/job/72757755094) on PR #272 (CC-0091): the keystone integration suite went from 605 s on `main` to >600 s after adding three Pass-0/Pass-1/Pass-2 wait cycles to `TestIntegration_FinalizerLifecycle_AddAndRemove`, `TestIntegration_FinalizerBrownfieldDeletion`, and the new `TestIntegrationKeystone_DeleteRacingESOAdoption`, hitting the default `go test -timeout 10m`. The unblock fix (Makefile bump to `-timeout 20m`) treats the symptom; the root cause is that the controller is blind to PushSecret events and has to poll. A `Watches(&esov1alpha1.PushSecret{}, …)` with a `pushSecretToKeystoneMapper` mirroring the existing `secretToKeystoneMapper` (`keystone_controller.go:609-…`) would collapse those waits to milliseconds, both in production teardown and in envtest.

**Proposed approach**:
1. Add `Watches(&esov1alpha1.PushSecret{}, handler.EnqueueRequestsFromMapFunc(pushSecretToKeystoneMapper(mgr.GetClient())))` to `SetupWithManager`.
[...]

PR #272, opened 2026-04-23
feat(CC-0091): Fix finalizer/PushSecret race leaking keys in OpenBao

berendt:
# GitHub Issue #271: Race between operator finalizer Delete and ESO PushSecret adoption leaks Fernet and credential keys in OpenBao

**Category**: bug | **Scope**: Medium

**Description**: The openbao-finalizer introduced in CC-0079 (PR #260, commit 09029b9) has a race with ESO's PushSecret adoption that can leak Fernet and credential key material in OpenBao KV-v2. `finalizeOpenBaoSecrets` (`operators/keystone/internal/controller/reconcile_secrets.go:185-197`) calls `r.Delete` on `<keystone>-fernet-keys-backup` and `<keystone>-credential-keys-backup` without first confirming that ESO has adopted the PushSecret by installing its own finalizer (`pushsecret.externalsecrets.external-secrets.io/finalizer`). If the operator's Delete races ahead of ESO's first reconcile — for example because the ESO PushSecret workqueue is backed up under parallel load — Kubernetes removes the PushSecret object immediately (no finalizer to hold it) and ESO never reaches the `DeletionPolicy=Delete` branch that would issue `DeleteSecret` against the remote. The KV-v2 paths `kv-v2/openstack/keystone/fernet-keys` and `kv-v2/openstack/keystone/credential-keys` remain populated, carry `custom_metadata: managed-by=external-secrets` but no `deletion_time`, and — because there is no PushSecret left to reference them — stay orphaned until manual intervention.

**Motivation**: Reproduced on `main` in CI run [24842115250/job 72722485985](https://github.com/C5C3/forge/actions/runs/24842115250/job/72722485985) (commit 4526cd7). The `keystone-deletion-cleanup` E2E step-6 asserts both KV-v2 paths are `not-found` after PushSecret deletion; both still contained `version: 1` data with `created_time` `2026-04-23T15:27:51.982Z` (Fernet) and `2026-04-23T15:27:52.089Z` (credential) — i.e., ESO's first push landed 13 s AFTER `Ready=True` (`15:27:38`) and only ~1.6 s before the PushSecret was garbage-collected at `15:27:53.6`. The immediately preceding CI run [24832475387](https://github.com/C5C3/forge/actions/runs/24832475387) on the PR merge commit 09029b9 was green — only a `golang:1.25` Docker-digest bump (#261) lay between the two runs — confirming this is a timing-sensitive race, not a regression from the digest bump.

[...]

PR #282, opened 2026-04-24
feat(CC-0092): Watch PushSecret in Keystone reconciler to cut delete latency

berendt:
# GitHub Issue #273: Watch PushSecret in Keystone reconciler to remove 15s deletion-cleanup latency

**Category**: enhancement | **Scope**: Small

**Description**: `KeystoneReconciler.SetupWithManager` (`operators/keystone/internal/controller/keystone_controller.go:582-604`) currently watches `corev1.Secret`, `mariadbv1alpha1.MariaDB`, and `esov1.ClusterSecretStore` but **not** `esov1alpha1.PushSecret`. As a result, the openbao-finalizer cleanup added in CC-0079/CC-0091 only re-evaluates Pass-0 (adoption gate) and Pass-2 (garbage-collection wait) at the next periodic requeue — `RequeueSecretPolling = 15 * time.Second` (`operators/keystone/internal/controller/requeue_intervals.go:22`). Each PushSecret state transition during teardown — ESO installing its cleanup finalizer, ESO releasing it after the kv-v2 purge, the API server garbage-collecting the Terminating object — therefore costs **up to 15 s** of avoidable wait, and a worst-case Keystone deletion accumulates ~30 s of pure polling latency across the three state changes the operator has to observe.

**Motivation**: Surfaced while diagnosing CI run [24852835587](https://github.com/C5C3/forge/actions/runs/24852835587/job/72757755094) on PR #272 (CC-0091): the keystone integration suite went from 605 s on `main` to >600 s after adding three Pass-0/Pass-1/Pass-2 wait cycles to `TestIntegration_FinalizerLifecycle_AddAndRemove`, `TestIntegration_FinalizerBrownfieldDeletion`, and the new `TestIntegrationKeystone_DeleteRacingESOAdoption`, hitting the default `go test -timeout 10m`. The unblock fix (Makefile bump to `-timeout 20m`) treats the symptom; the root cause is that the controller is blind to PushSecret events and has to poll. A `Watches(&esov1alpha1.PushSecret{}, …)` with a `pushSecretToKeystoneMapper` mirroring the existing `secretToKeystoneMapper` (`keystone_controller.go:609-…`) would collapse those waits to milliseconds, both in production teardown and in envtest.

**Proposed approach**:
1. Add `Watches(&esov1alpha1.PushSecret{}, handler.EnqueueRequestsFromMapFunc(pushSecretToKeystoneMapper(mgr.GetClient())))` to `SetupWithManager`.
2. Implement `pushSecretToKeystoneMapper` next to the existing mappers — namespace-scoped `List`, then match the PushSecret name against `openBaoBackupPushSecretNames(keystone)` to enqueue only the owning CR (avoids waking unrelated Keystones in shared namespaces).
3. Consider a `builder.WithEventFilter(predicate.GenerationChangedPredicate{})`-style filter to ignore status-only updates from ESO that do not affect finalizers/DeletionTimestamp, so steady-state ESO sync churn does not flood the queue. The two events the operator actually cares about are: (a) `pushsecret.externalsecrets.external-secrets.io/finalizer` added/removed, (b) `DeletionTimestamp` set/cleared.
4. Unit-test the mapper alongside `secretToKeystoneMapper` in `keystone_controller_test.go` — confirm name-match enqueue, name-mismatch skip, cross-namespace skip, list-error tolerance.
5. After landing, the integration suite's three `Eventually(…, eventuallyTimeout, …)` finalizer cycles should each shrink from ~15-20 s to ~1-2 s; the Makefile `-timeout 20m` bump can stay in place as headroom but stops being load-bearing.

**Acceptance**:
- `pushSecretToKeystoneMapper` exists with unit tests covering enqueue/skip/error paths.
- `SetupWithManager` registers the watch.
- `TestIntegrationKeystone_DeleteRacingESOAdoption` end-to-end runtime drops measurably (target: <30 s vs. current ~50-70 s).
- No regression in the openbao-finalizer adoption tests (`TestFinalizeOpenBaoSecrets_*` in `reconcile_secrets_test.go`).

**Refs**: CC-0079, CC-0091, REQ-001, REQ-007. Spin-off from PR #272.

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/273


berendt:
> Reviewed by planwerk-review c600a8d with Claude Code



## INFO (2)

### I-001: Documentation overstates empty-namespace safety guarantee
**File**: `operators/keystone/internal/controller/keystone_controller.go:832-835` — **Fix**: ASK — **Confidence**: verified

**Problem**: The mapper docblock and docs/reference/keystone-reconciler.md claim the mapper 'never enqueues cross-namespace' and that an event with no namespace returns nil. The code uses client.InNamespace(obj.GetNamespace()) unconditionally; controller-runtime treats client.InNamespace("") as cluster-wide, so the docs describe a defensive behaviour the code does not actually enforce. In practice PushSecret is a namespaced resource and the apiserver guarantees a non-empty namespace, but the documentation reads as if the code handles the empty case explicitly.

**Code**:
```
// Namespace scoping is load-bearing — REQ-002 requires that a PushSecret event in ns-b
// never wake a Keystone that lives in ns-a, so the List MUST carry
// client.InNamespace(obj.GetNamespace()) only (never cluster-wide). On a List
// error the mapper logs via log.FromContext and returns nil per the
```

**Action Required**: Pick one: tighten the doc claim to match what the code actually enforces, or add an explicit empty-namespace guard at the top of pushSecretToKeystoneMapper so the claim holds by construction. Update both the in-source docblock and docs/reference/keystone-reconciler.md.

**Suggested Fix**:
```
Option A (docs-only): rewrite the docblock and the Namespace scope row in docs/reference/keystone-reconciler.md to read 'A PushSecret event whose namespace matches no Keystone returns nil. PushSecrets are namespaced resources, so the apiserver guarantees obj.GetNamespace() is non-empty in practice.' Option B (code-side): add `if obj.GetNamespace() == "" { return nil }` as the first statement of pushSecretToKeystoneMapper, then keep the existing doc claim.
```

**Related**: I-002

---

### I-002: Test name misleading about empty-namespace coverage
**File**: `operators/keystone/internal/controller/keystone_controller_test.go:3361-3375` — **Fix**: ASK — **Confidence**: verified

**Problem**: TestPushSecretToKeystoneMapper_EmptyNamespaceReturnsNil reads as if it exercises the case where the event object's namespace is the empty string, but the body actually covers 'no Keystone CRs live in the (non-empty) namespace default'. The mismatch between test name and behaviour reinforces the documentation overstatement in the related finding.

**Code**:
```
// TestPushSecretToKeystoneMapper_EmptyNamespaceReturnsNil verifies the
// empty-list path: when no Keystone CR lives in the event's namespace, the
// mapper returns nothing without erroring (CC-0092, REQ-003).
func TestPushSecretToKeystoneMapper_EmptyNamespaceReturnsNil(t *testing.T) {
```

**Action Required**: If the doc claim about empty-namespace handling is kept (i.e. the related finding is resolved by adding a code guard), add a sibling test that constructs pushSecretObj("anything-backup", "") and asserts the mapper returns nil. If the doc claim is tightened instead, rename this test to TestPushSecretToKeystoneMapper_NoKeystonesInNamespaceReturnsNil to match what it actually exercises.

**Suggested Fix**:
```
Rename the function and its docblock to TestPushSecretToKeystoneMapper_NoKeystonesInNamespaceReturnsNil so the name reflects 'no Keystones in the (non-empty) namespace'. If the related finding is resolved by adding a code guard for empty namespace, additionally introduce TestPushSecretToKeystoneMapper_PushSecretWithoutNamespaceReturnsNil that builds pushSecretObj("anything-backup", "") and asserts the mapper returns nil.
```

**Related**: I-001

---

## Summary

Focused, well-motivated change converting the Owns()-based PushSecret subscription into a Watches() + name-mapper + relevance-predicate trio, eliminating per-ESO-sync-tick workqueue churn and shaving up to ~15s of polling latency from the OpenBao finalizer state machine. The mapper and predicate are pure, mirror existing patterns (secretToKeystoneMapper, mariaDBToKeystoneMapper), and ship with comprehensive unit-test coverage of every admission and skip path; the sourcery review feedback is already addressed. The only remaining items are a paired documentation/test-naming inconsistency around empty-namespace handling — the docs describe a defensive behaviour the code does not actually enforce, and a test name reinforces the overstatement.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 0 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Approve and merge after addressing the two linked INFO-level findings (either tighten the docs or add an empty-namespace guard, and align the test name/coverage accordingly). No blocking, critical, or warning-level issues.

gndrmnn (approved):

Issue #281, opened 2026-04-24
test(e2e): Expand webhook/CEL rejection matrix for Keystone CR (#277 Phase 1)


berendt:
## Context

Parent: #277, Phase 1 → *Webhook / CEL validation matrix*.

Today `tests/e2e/keystone/invalid-cr/` covers exactly two rejection cases: invalid cron in `spec.fernet.rotationSchedule` and duplicate `plugins[*].configSection` values (see `tests/e2e/keystone/invalid-cr/chainsaw-test.yaml`, REQ-008/009 under CC-0012).

The validation rules in `operators/keystone/api/v1alpha1/keystone_types.go` (CEL via `+kubebuilder:validation:XValidation`) and `operators/keystone/api/v1alpha1/keystone_webhook.go` (defense-in-depth) cover considerably more misconfigurations than that. This sub-issue closes the E2E gap for every rejection path that is already implemented.

## Scope — new rejection cases

The existing test is extended with eleven input files and matching `apply` steps. Each row is taken 1:1 from the Phase 1 checklist in #277:

| # | Scenario | Enforced by | Expected `fieldPath` / key string |
|---|----------|-------------|------------------------------------|
| 1 | `spec.database` with **both** `clusterRef` and `host` set | CEL on `DatabaseSpec` (keystone_types.go:61) + webhook (keystone_webhook.go:195) | `spec.database` / `exactly one of clusterRef or host must be set` |
| 2 | `spec.cache` with **both** `clusterRef` and `servers` set | CEL on `CacheSpec` (types.go:66) + webhook (webhook.go:185) | `spec.cache` / `exactly one of clusterRef or servers must be set` |
| 3 | `spec.autoscaling` with neither `targetCPUUtilization` nor `targetMemoryUtilization` | CEL on `AutoscalingSpec` (types.go:207) + webhook | `spec.autoscaling` / `at least one of targetCPUUtilization or targetMemoryUtilization` |
| 4 | `spec.policyOverrides` with neither `rules` nor `configMapRef` | CEL (types.go:100) + webhook | `spec.policyOverrides` / `at least one of rules or configMapRef must be set` |
| 5 | `spec.policyOverrides.rules` with an empty key | CEL (types.go:101) + webhook | `spec.policyOverrides.rules` / `policy rule name must not be empty` |
[...]

Issue #277, opened 2026-04-24
Meta: Expand E2E test coverage for Keystone operator


berendt:
## Background

The Keystone operator currently ships ~44 E2E tests (31 happy-path Chainsaw tests under `tests/e2e/keystone/`, 9 Chaos Mesh tests under `tests/e2e-chaos/`, 2 infrastructure-stack tests, plus Tempest runs for 2025.2 and 2026.1). Broad E2E coverage is our USP, and a recent audit surfaced concrete gaps across webhook validation, failure/recovery paths, rotation edge cases, chaos scenarios, multi-instance behaviour, and observability.

This meta issue tracks the rollout plan for closing those gaps so nothing gets lost and links all sub-issues.

Approach: start with quick wins that need no new infrastructure (pure Chainsaw YAML, small scripts), then add tests that require additional infra (multi-node kind, load generator, Envoy Gateway), expand chaos coverage, and finish with long-running / baseline regression gates.

### Phase 1 — Quick wins (no new infrastructure)

Webhook / CEL validation matrix
- [x] Webhook rejects `database` with both `clusterRef` and `host` set
- [x] Webhook rejects `cache` with both `clusterRef` and `servers` set
- [x] Webhook rejects `autoscaling` with neither CPU nor memory target
- [x] Webhook rejects `policyOverrides` with neither `rules` nor `configMapRef`, and empty rule keys
- [x] Webhook rejects empty `networkPolicy.ingress`
- [x] Webhook rejects `replicas: 0` / negative values
- [x] Webhook rejects HPA with `minReplicas > maxReplicas`
- [x] Webhook rejects `fernet` / `credentialKeys` with `maxActiveKeys < 2`
- [x] Webhook rejects empty or malformed `spec.image.tag`
- [x] Webhook rejects `topologySpreadConstraints` with `maxSkew: 0`
- [x] Webhook rejects mutation of immutable fields (e.g. `spec.database.clusterRef`)

Semantic / state checks
- [ ] Assert `status.endpoint` matches the expected `http://<svc>.<ns>.svc:5000/v3`
- [ ] Assert `ownerReferences` on every operator-managed object point to the Keystone CR
- [ ] Assert `status.observedGeneration == metadata.generation` after spec changes (CC-0072)
- [ ] Assert `condition.lastTransitionTime` is monotonic across flips
- [ ] ConfigMap drift — manual edit of generated ConfigMap is reverted by the operator
[...]

PR #283, opened 2026-04-25
feat(CC-0094): test(e2e): Expand webhook/CEL rejection matrix for Keystone

berendt:
# GitHub Issue #281: test(e2e): Expand webhook/CEL rejection matrix for Keystone CR (#277 Phase 1)

## Context

Parent: #277, Phase 1 → *Webhook / CEL validation matrix*.

Today `tests/e2e/keystone/invalid-cr/` covers exactly two rejection cases: invalid cron in `spec.fernet.rotationSchedule` and duplicate `plugins[*].configSection` values (see `tests/e2e/keystone/invalid-cr/chainsaw-test.yaml`, REQ-008/009 under CC-0012).

The validation rules in `operators/keystone/api/v1alpha1/keystone_types.go` (CEL via `+kubebuilder:validation:XValidation`) and `operators/keystone/api/v1alpha1/keystone_webhook.go` (defense-in-depth) cover considerably more misconfigurations than that. This sub-issue closes the E2E gap for every rejection path that is already implemented.

## Scope — new rejection cases

The existing test is extended with eleven input files and matching `apply` steps. Each row is taken 1:1 from the Phase 1 checklist in #277:

| # | Scenario | Enforced by | Expected `fieldPath` / key string |
|---|----------|-------------|------------------------------------|
| 1 | `spec.database` with **both** `clusterRef` and `host` set | CEL on `DatabaseSpec` (keystone_types.go:61) + webhook (keystone_webhook.go:195) | `spec.database` / `exactly one of clusterRef or host must be set` |
| 2 | `spec.cache` with **both** `clusterRef` and `servers` set | CEL on `CacheSpec` (types.go:66) + webhook (webhook.go:185) | `spec.cache` / `exactly one of clusterRef or servers must be set` |
| 3 | `spec.autoscaling` with neither `targetCPUUtilization` nor `targetMemoryUtilization` | CEL on `AutoscalingSpec` (types.go:207) + webhook | `spec.autoscaling` / `at least one of targetCPUUtilization or targetMemoryUtilization` |
| 4 | `spec.policyOverrides` with neither `rules` nor `configMapRef` | CEL (types.go:100) + webhook | `spec.policyOverrides` / `at least one of rules or configMapRef must be set` |
| 5 | `spec.policyOverrides.rules` with an empty key | CEL (types.go:101) + webhook | `spec.policyOverrides.rules` / `policy rule name must not be empty` |
| 6 | `spec.networkPolicy.ingress: []` (empty) | CEL on `NetworkPolicySpec` (types.go:236) + webhook | `spec.networkPolicy.ingress` / `at least one ingress source must be specified` |
| 7 | `spec.replicas: 0` **and** `spec.replicas: -1` (two variants) | `+kubebuilder:validation:Minimum=1` + webhook | `spec.replicas` / `replicas must be at least 1` |
| 8 | `spec.autoscaling.minReplicas > spec.autoscaling.maxReplicas` | webhook (webhook.go ~428) | `spec.autoscaling.minReplicas` / `must not exceed maxReplicas` |
| 9 | `spec.fernet.maxActiveKeys: 2` **and** `spec.credentialKeys.maxActiveKeys: 2` | `+kubebuilder:validation:Minimum=3` + webhook | `spec.fernet.maxActiveKeys` / `must be at least 3` (same for `credentialKeys`) |

> Note: the parent checklist says `maxActiveKeys < 2`, but the code enforces `>= 3`. The test follows the code; the parent checklist will be corrected when the box is ticked.

## Explicitly out of scope (prerequisites, separate tickets)

Three checkboxes from Phase 1 / "Webhook matrix" require webhook features that do **not** exist today. They need feature tickets first; the E2E tests follow once the feature lands (see #277, *Out of scope*):

- Empty / malformed `spec.image.tag` — `commonv1.ImageSpec.Tag` (`internal/common/types/types.go:14`) has neither `MinLength` nor a regex pattern.
- `topologySpreadConstraints[*].maxSkew: 0` — upstream `corev1.TopologySpreadConstraint.MaxSkew` carries no CRD-level minimum; no defense-in-depth in the Keystone webhook today.
- Mutation of immutable fields (`spec.database.clusterRef`, analogously `spec.cache.clusterRef`) — `ValidateUpdate` does not currently compare old vs. new.

Those three points will be split out of the Phase 1 list into their own `feat(CC-xxxx): harden <field> validation` issues, each followed by a dedicated test.

## Test design

Location: `tests/e2e/keystone/invalid-cr/` (extend the existing test — **no** new directory; every case shares the scope "webhook rejects an otherwise minimal Keystone CR").

File layout after the change:

```
tests/e2e/keystone/invalid-cr/
├── 00-invalid-cron.yaml                             (existing)
├── 01-duplicate-plugins.yaml                        (existing)
├── 02-database-both-modes.yaml                      (new)
├── 03-cache-both-modes.yaml                         (new)
├── 04-autoscaling-no-target.yaml                    (new)
├── 05-policy-overrides-no-source.yaml               (new)
├── 06-policy-overrides-empty-rule-key.yaml          (new)
├── 07-networkpolicy-empty-ingress.yaml              (new)
├── 08-replicas-zero.yaml                            (new)
├── 09-replicas-negative.yaml                        (new)
├── 10-hpa-min-greater-than-max.yaml                 (new)
├── 11-fernet-maxactivekeys-below-minimum.yaml       (new)
├── 12-credentialkeys-maxactivekeys-below-minimum.yaml (new)
└── chainsaw-test.yaml                               (extended)
```

Assertion pattern per step (identical to `00-invalid-cron.yaml`, so CI reports stay homogeneous):

```yaml
- try:
  - apply:
      file: 02-database-both-modes.yaml
      expect:
      - check:
          ($error != null): true
          (contains($error, 'spec.database')): true
          (contains($error, 'exactly one of clusterRef or host must be set')): true
```

Input YAMLs: each file carries a **complete, otherwise valid** minimal CR where exactly **one** field violates the rule under test. That keeps the error-message assertion sharp and makes follow-on regressions (e.g. a new required field) fail loudly. Template: the existing `00-invalid-cron.yaml` / `01-duplicate-plugins.yaml`.

All CRs remain **unapplied** in cluster state (admission reject), so no cleanup steps and no `delete` blocks are needed. The test fits inside the 30 s `apply` timeout from `tests/e2e/chainsaw-config.yaml` and does not affect `parallel: 4` execution.

## Definition of Done

- [ ] `chainsaw test tests/e2e/keystone/invalid-cr` green locally against `make kind-up`
- [ ] `make e2e` green (rest of the Keystone suite unchanged, no new flakes)
- [ ] Parallel-4 CI job `e2e` green in at least two consecutive runs (anti-flake)
- [ ] All eleven new steps use the `(contains($error, …))` pattern from the existing test; no dependency on an exact HTTP-code string
- [ ] No new infra components in `deploy/kind/`, no deviations from `tests/e2e/chainsaw-config.yaml`
- [ ] Parent #277: the nine covered checkboxes ticked, the three prerequisite points linked as follow-up issues
- [ ] Commit prefix `test(CC-xxxx)` in line with the repo style (feature code to be assigned by a maintainer)

## Risks / open questions

- **Error-message drift**: the CEL and webhook layers can produce slightly different strings (e.g. for `replicas: 0` the CRD layer says `"should be greater than or equal to 1"`, the webhook says `"replicas must be at least 1"`). Mitigation: match on the common substring (`"replicas"`) plus `($error != null): true`. Please confirm in review.
- **Validator ordering**: CRD schema → CEL → validating webhook. If CEL rejects first, the request never reaches the webhook and the webhook-specific string is absent from `$error`. The test therefore matches the semantic core message, not the producing layer.
- Feature code (`CC-xxxx`) to be assigned by a maintainer; suggest extending the sequence past CC-0093.

---
Source: https://github.com/C5C3/forge/issues/281


berendt:
> Reviewed by planwerk-review 889e22e with Claude Code



## WARNING (2)

### W-001: verify-invalid-cr-fixtures is not actually gating e2e-operator
**File**: `.github/workflows/ci.yaml:214-220` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The chainsaw-test.yaml comment (lines 19-25), CI workflow comment (.github/workflows/ci.yaml:205-213), and docs (keystone-crd.md paragraph after the step table) all claim the job 'fails the build before the cluster-bound e2e-operator job runs'. However, the verify-invalid-cr-fixtures job has no needs: block and is not listed in needs: of build-e2e-images (line 462) or e2e-operator (line 599). They run in parallel; the only 'gating' is the overall PR fail status. The documentation overstates the actual dependency graph.

**Code**:
```
214:  verify-invalid-cr-fixtures:
215:    runs-on: ubuntu-latest
216:    timeout-minutes: 5
217:    steps:
218:      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
219:      - name: Verify CC-0094 invalid-CR fixture drift and structure
220:        run: make verify-invalid-cr-fixtures
```

**Action Required**: Add `needs: [verify-invalid-cr-fixtures]` to `build-e2e-images` so the documented gating actually holds. The check is sub-second and free, so this is the cheapest alignment with the documentation.

**Suggested Fix**:
```
  build-e2e-images:
    needs: [changes, lint, shellcheck, test, test-integration, verify-codegen, verify-invalid-cr-fixtures]
```

---

### W-002: One-sided drift detection: orphan files on disk are silently tolerated
**File**: `tests/e2e/keystone/invalid-cr/_generate.py:302-324` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: _generate.py:302-324 iterates only over FIXTURES. If a contributor removes an entry from FIXTURES but forgets to `git rm` the corresponding YAML, the orphan file remains on disk and `--check` exits 0. test_generate.py also only walks FIXTURES, so the cross-reference test does not catch it either. If the orphan is also removed from chainsaw-test.yaml, it becomes invisible dead code.

**Code**:
```
302	def main(argv: list[str]) -> int:
303	    check_only = "--check" in argv
304	    here = Path(__file__).resolve().parent
305	    drift: list[str] = []
306	    for fixture in FIXTURES:
307	        rendered = render(fixture)
308	        target = here / fixture.filename
309	        if target.exists() and target.read_text() == rendered:
310	            continue
```

**Action Required**: After the FIXTURES loop, list all NN-*.yaml files matching the CC-0094 pattern and report any not in {f.filename for f in FIXTURES} as drift (or unlink them when not in --check mode).

**Suggested Fix**:
```
    expected = {f.filename for f in FIXTURES}
    pattern = re.compile(r"^(0[2-9]|1[0-9])-.+\.yaml$")
    for path in here.iterdir():
        if path.is_file() and pattern.match(path.name) and path.name not in expected:
            if check_only:
                drift.append(f"{path.name} (orphan: not in FIXTURES)")
            else:
                path.unlink()
                print(f"removed orphan {path.name}")
```

---

## INFO (1)

### I-001: Missing explicit encoding="utf-8" in _generate.py
**File**: `tests/e2e/keystone/invalid-cr/_generate.py:309-314` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: _generate.py:309, 314 calls target.read_text() and target.write_text(rendered) without an encoding= argument, relying on locale.getpreferredencoding(). The generated comments contain em-dashes (—, U+2014, multi-byte UTF-8). test_generate.py:81 correctly uses encoding="utf-8". Inconsistent, and on a runner with a non-UTF-8 locale (e.g. minimal containers, some Windows defaults) the round-trip could mis-encode and mark every fixture as drifted.

**Code**:
```
309	        if target.exists() and target.read_text() == rendered:
310	            continue
311	        if check_only:
312	            drift.append(fixture.filename)
313	        else:
314	            target.write_text(rendered)
```

**Action Required**: Thread encoding="utf-8" through both read_text and write_text calls to match test_generate.py and guarantee deterministic behavior across locales.

**Suggested Fix**:
```
        if target.exists() and target.read_text(encoding="utf-8") == rendered:
            continue
        if check_only:
            drift.append(fixture.filename)
        else:
            target.write_text(rendered, encoding="utf-8")
```

---

## Summary

This is a well-executed test-expansion PR for CC-0094: validation paths are correctly traced from fixtures to webhook/CEL code, the dual-layer admission ordering is understood and documented (including the mutating-webhook coercion trap justifying the omission of 08-replicas-zero.yaml), and the canonical-scaffold + unit-test + CI verification stack effectively prevents drift. Two warnings are worth addressing: the documented gating of e2e-operator by verify-invalid-cr-fixtures is not actually wired up, and the drift detection is one-sided (orphan YAML files slip through). The encoding inconsistency in _generate.py is a small polish item.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 2 |
| INFO     | 1 |

> [!IMPORTANT]
> **Recommendation**: Mergeable after addressing W-001 to align the workflow with the documented gating claim. W-002 and I-001 are recommended improvements but do not block merge.

Issue #285, opened 2026-04-26
Drop -api suffix from operator-managed sub-resources


berendt:
**Category**: enhancement | **Scope**: Medium

**Description**: Remove the `-api` suffix from every Keystone sub-resource the operator creates so that the cluster-local DNS name of the Keystone Service — and therefore the URL seeded into the OpenStack service catalog — drops from `keystone-api.<ns>.svc.cluster.local` to plain `keystone.<ns>.svc.cluster.local`. Today `openstack endpoint list` on a freshly-bootstrapped cluster shows an asymmetric pair of hostnames:

```
| internal | http://keystone-api.openstack.svc.cluster.local:5000/v3 |
| admin    | http://keystone-api.openstack.svc.cluster.local:5000/v3 |
| public   | https://keystone.127-0-0-1.nip.io:8443/v3               |
```

The `public` row already uses the bare `keystone` label thanks to CC-0088 (`spec.gateway.hostname` → `https://keystone.127-0-0-1.nip.io/v3`). The internal/admin rows still carry the `-api` suffix because the bootstrap Job hardcodes that pattern in `operators/keystone/internal/controller/reconcile_bootstrap.go:67-68` (`bootstrapServiceURL()` → `fmt.Sprintf("http://%s-api.%s.svc.cluster.local:5000/v3", keystone.Name, keystone.Namespace)`), and because the underlying Service itself is named via the centralised helper `apiResourceName()` at `operators/keystone/internal/controller/reconcile_hpa.go:23-28` which appends `-api` to `keystone.Name`. The same helper drives the names of every sub-resource the operator owns: Deployment (`reconcile_deployment.go:152`), Service (`reconcile_deployment.go:453`), PodDisruptionBudget (`reconcile_deployment.go:281`), HPA (`reconcile_hpa.go:36,81`), NetworkPolicy (`reconcile_networkpolicy.go:43,150`), HTTPRoute (`reconcile_httproute.go:147,225,249`), and the `internalAPIURL()` helper (`reconcile_httproute.go:97`) which feeds the HTTPRoute backendRef. The `-api` suffix is redundant: the operator is single-purpose for Keystone, the Kubernetes resource kind already disambiguates `Deployment/keystone` from `Service/keystone`, and the suffix actively diverges from the public-facing hostname customers and operators interact with.

Concrete steps:

[...]

PR #288, opened 2026-04-26
feat(CC-0095): Drop -api suffix from operator-managed sub-resources

berendt:
# GitHub Issue #285: Drop -api suffix from operator-managed sub-resources

**Description**: Remove the `-api` suffix from every Keystone sub-resource the operator creates so that the cluster-local DNS name of the Keystone Service — and therefore the URL seeded into the OpenStack service catalog — drops from `keystone-api.<ns>.svc.cluster.local` to plain `keystone.<ns>.svc.cluster.local`. Today `openstack endpoint list` on a freshly-bootstrapped cluster shows an asymmetric pair of hostnames:

```
| internal | http://keystone-api.openstack.svc.cluster.local:5000/v3 |
| admin    | http://keystone-api.openstack.svc.cluster.local:5000/v3 |
| public   | https://keystone.127-0-0-1.nip.io:8443/v3               |
```

The `public` row already uses the bare `keystone` label thanks to CC-0088 (`spec.gateway.hostname` → `https://keystone.127-0-0-1.nip.io/v3`). The internal/admin rows still carry the `-api` suffix because the bootstrap Job hardcodes that pattern in `operators/keystone/internal/controller/reconcile_bootstrap.go:67-68` (`bootstrapServiceURL()` → `fmt.Sprintf("http://%s-api.%s.svc.cluster.local:5000/v3", keystone.Name, keystone.Namespace)`), and because the underlying Service itself is named via the centralised helper `apiResourceName()` at `operators/keystone/internal/controller/reconcile_hpa.go:23-28` which appends `-api` to `keystone.Name`. The same helper drives the names of every sub-resource the operator owns: Deployment (`reconcile_deployment.go:152`), Service (`reconcile_deployment.go:453`), PodDisruptionBudget (`reconcile_deployment.go:281`), HPA (`reconcile_hpa.go:36,81`), NetworkPolicy (`reconcile_networkpolicy.go:43,150`), HTTPRoute (`reconcile_httproute.go:147,225,249`), and the `internalAPIURL()` helper (`reconcile_httproute.go:97`) which feeds the HTTPRoute backendRef. The `-api` suffix is redundant: the operator is single-purpose for Keystone, the Kubernetes resource kind already disambiguates `Deployment/keystone` from `Service/keystone`, and the suffix actively diverges from the public-facing hostname customers and operators interact with.

Concrete steps:

1. Change `apiResourceName()` in `operators/keystone/internal/controller/reconcile_hpa.go:26-28` to return `keystone.Name` directly. Optionally rename the helper to something neutral (`coreResourceName` or `subResourceName`) since the `api` framing was only meaningful while the suffix existed; keep the rename mechanical.
2. Refactor `bootstrapServiceURL()` in `operators/keystone/internal/controller/reconcile_bootstrap.go:67-68` to delegate to `apiResourceName(keystone)` instead of re-implementing the `%s-api` pattern. After this, the bootstrap URL automatically tracks any future change to the helper — divergence is impossible by construction.
3. Update the two remaining hardcoded literals in the Deployment template at `operators/keystone/internal/controller/reconcile_deployment.go:171,178` (the container name and the named container port both spelled `"keystone-api"`). Change them to `"keystone"`. The named port is referenced through Service `targetPort` (which uses an int, not a name, see `reconcile_deployment.go:461`) and through the HTTPRoute `backendRef` port (an int, see `reconcile_httproute.go:236-238`), so this is a local cosmetic rename — no cross-resource selector needs updating.
4. Walk the tests under `operators/keystone/internal/controller/` and update assertions that currently expect `*-api.<ns>.svc.cluster.local`, `test-keystone-api`, `keystone-api` container/port names, etc. The known-affected files are `reconcile_deployment_test.go`, `reconcile_healthcheck_test.go`, `reconcile_networkpolicy_test.go`, `reconcile_hpa_test.go`, `reconcile_httproute_test.go`, `reconcile_bootstrap_test.go`, `keystone_controller_test.go`, `integration_test.go`. Roughly 100+ assertion sites in total.
5. Walk the e2e fixtures and update container-name selectors and PodChaos pod-name selectors:
   - `tests/e2e/keystone/resources/chainsaw-test.yaml` (container name assertions at `:49`, `:98`)
   - `tests/e2e/keystone/uwsgi/chainsaw-test.yaml` (container name assertions at `:49`, `:106`)
   - `tests/e2e/keystone/graceful-shutdown/chainsaw-test.yaml:73`
   - `tests/e2e/keystone/configmap-no-secrets/chainsaw-test.yaml:166`
   - `tests/e2e-chaos/api-pod-kill-pdb/chainsaw-test.yaml` and `tests/e2e-chaos/api-pod-kill-pdb/01-podchaos.yaml` (PodChaos `name: kill-keystone-api` and label-selector references — these target Pods by labelSelector, not by name, so verify the actual selectors after the rename)
6. Walk the docs and architecture references and rewrite cluster-internal hostname examples from `keystone-api.openstack.svc.cluster.local` to `keystone.openstack.svc.cluster.local`. The Quick Start has at least three sites that explicitly print the internal URL (`docs/quick-start.md:638,699` and the Fallback subsection at `:736-740`), the architecture docs under `architecture/docs/09-implementation/` and `architecture/docs/05-deployment/` reference the `keystone-api` container name and Service in code examples, and `docs/reference/keystone-crd.md`, `docs/reference/keystone-reconciler.md`, `docs/reference/keystone-upgrade-flow.md`, `docs/guides/observability.md`, `docs/guides/advanced-configuration.md`, `docs/guides/day-2-operations.md`, and `docs/reference/chaos-e2e-tests.md` all contain matching strings. About 70+ doc sites in total. The Fallback `kubectl port-forward svc/keystone-api ...` invocation in `docs/quick-start.md:736-740` and the matching unit doc test in `tests/unit/docs/quick_start_access_section_test.sh` need to be updated together so the Fallback section keeps passing its assertion gate.
7. Update the migration-relevant comments — the apologetic `// containerResources() for the pattern used by the keystone-api container.` strings scattered across `reconcile_policyvalidation.go:168`, `reconcile_fernet.go:317`, `reconcile_credential.go:322`, `reconcile_database.go:620`, and `reconcile_bootstrap.go:134` should be touched up to drop the `-api` framing so future readers don't get confused about why a bare `keystone` container is being described as `keystone-api`.
8. Document the breaking change in the Keystone CRD reference and in the upgrade-flow reference — existing OpenStack control planes have `http://keystone-api.<ns>.svc.cluster.local:5000/v3` baked into the `endpoint` rows of the Keystone DB. After the rename, the Service `keystone-api` no longer exists, so any other OpenStack service whose `keystone_authtoken.www_authenticate_uri` / `auth_url` points at the old hostname will fail. The operator's bootstrap reconciler runs `keystone-manage bootstrap` which is idempotent and updates the existing admin/internal/public endpoint rows in-place; on the next reconcile after the upgrade, the catalog will self-heal to the new hostname. Document this explicitly in the upgrade-flow guide so operators know to either: (a) re-roll the bootstrap Job (the operator's BootstrapReady condition flips to False on Generation change, which forces re-bootstrap), or (b) run `openstack endpoint set` against the two cluster-internal rows after the upgrade. Add a one-line callout to the CRD reference explaining that the operator-managed Service is now named after the CR (`metadata.name`) without suffix.

**Motivation**: Three concrete drivers.

(a) **Symmetry with the public URL.** CC-0088 landed `https://keystone.127-0-0-1.nip.io/v3` as the documented public access path in the Quick Start, and the CRD's `status.endpoint` is derived as `https://{spec.gateway.hostname}/v3` — both consistently use the bare `keystone` label. The internal/admin rows in the catalog should match: there is no architectural reason for the internal hostname to carry an extra `-api` suffix when the public one does not, and the asymmetry shows up in every fresh Quick Start `openstack endpoint list` output.

(b) **The `-api` suffix is dead semantic weight.** It dates from a time when an operator might have managed multiple sibling components per CR (`keystone-api`, `keystone-something-else`), but in this codebase the Keystone operator only manages a single workload per CR. `Deployment/keystone`, `Service/keystone`, `HPA/keystone`, `PodDisruptionBudget/keystone`, `NetworkPolicy/keystone`, `HTTPRoute/keystone` — the kind disambiguates them perfectly, the `metadata.name` already matches the CR `metadata.name`, and a `kubectl describe keystone <name> && kubectl get all -l app.kubernetes.io/instance=<name>` walk reads cleanly without a parallel `<name>-api` family of objects sitting alongside.

(c) **`bootstrapServiceURL()` reimplements the naming convention instead of delegating to `apiResourceName()`.** That is a latent bug independent of the suffix question: any future change to the Service name pattern would silently desync the bootstrap-seeded catalog URL from the actual Service hostname. Folding this single hardcoded `fmt.Sprintf` into the central helper closes the divergence channel at the same time as the rename, with no extra cost.

This change is the natural follow-up to CC-0088 (#282) and lines the internal posture up with the public posture established by that feature.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_hpa.go:26-28` — flip `apiResourceName()` body to `return keystone.Name`; consider renaming the helper to a suffix-neutral identifier
- `operators/keystone/internal/controller/reconcile_bootstrap.go:67-68` — refactor `bootstrapServiceURL()` to delegate to `apiResourceName(keystone)` so the bootstrap URL tracks the helper automatically
- `operators/keystone/internal/controller/reconcile_deployment.go:171,178` — rename container name and port name from `"keystone-api"` to `"keystone"`
- All controller-package tests that assert on resource names, container/port names, or cluster-internal URLs (the `*_test.go` files listed under "Concrete steps" #4 above)
- e2e fixtures listed under "Concrete steps" #5 above
- Documentation sites listed under "Concrete steps" #6 above (`docs/`, `architecture/docs/`, plus the `tests/unit/docs/quick_start_access_section_test.sh` gate that locks the Quick Start Fallback subsection wording)
- Doc-comment cleanups listed under "Concrete steps" #7 above

**Acceptance Criteria**:
- [ ] `apiResourceName(keystone)` returns `keystone.Name` (no suffix) — verified by a focused unit test
- [ ] On a freshly bootstrapped cluster, `openstack endpoint list` returns `internal` and `admin` rows with URL `http://keystone.<ns>.svc.cluster.local:5000/v3` (no `-api`)
- [ ] `kubectl get svc -n openstack -l app.kubernetes.io/instance=<keystone-cr-name>` returns a single Service named `<keystone-cr-name>` (matching `metadata.name` of the CR), with no Service named `<keystone-cr-name>-api`
- [ ] `kubectl get deploy,hpa,pdb,networkpolicy,httproute -n openstack -l app.kubernetes.io/instance=<keystone-cr-name>` reports objects all named `<keystone-cr-name>`
- [ ] `bootstrapServiceURL()` no longer contains the literal string `"-api"`; its return value is computed by composing `apiResourceName(keystone)` with the namespace and the `:5000/v3` suffix
- [ ] `grep -rn 'keystone-api'` across `operators/`, `tests/`, `docs/`, `architecture/docs/`, `hack/`, `deploy/` returns zero hits — every site is either renamed to `keystone` or, where the literal must remain (legacy migration callouts in upgrade docs), explicitly annotated as a pre-rename artefact
- [ ] All controller-package unit tests pass against the renamed fixtures
- [ ] The chainsaw e2e suites under `tests/e2e/keystone/` and `tests/e2e-chaos/` pass on a kind cluster after the rename, including the PodChaos suite that selects API Pods by label
- [ ] The Quick Start `tests/unit/docs/quick_start_access_section_test.sh` gate still passes — the Fallback subsection now references `kubectl port-forward svc/keystone` and the primary section continues to omit any port-forward
- [ ] An upgrade callout is added to `docs/reference/keystone-upgrade-flow.md` explaining the Service rename and the catalog self-heal path
- [ ] The CRD reference (`docs/reference/keystone-crd.md`) documents the new sub-resource naming convention (no `-api` suffix) and links to the upgrade callout
- [ ] Reconciliation of an existing pre-rename CR on the upgrade path is verified end-to-end: the new Deployment/Service/HPA come up with the new names, the old `<name>-api` objects are GC'd by ownerReference cascade (or explicitly deleted by a one-shot migration step in the controller — see Non-Goals), the BootstrapReady condition flips back to True, and `openstack endpoint list` reflects the new internal URL
- [ ] No selector or label changes are required: `selectorLabels()` and `commonLabels()` already key off `keystone.Name` only and are unaffected

**Non-Goals**:
- Adding a runtime migration step inside the controller to delete the old `<name>-api` Service/Deployment/HPA/PDB/NetworkPolicy/HTTPRoute objects on first reconcile after the upgrade. OwnerReference cascade should handle this when the operator stops creating the old-named objects, but it relies on the controller-runtime cache catching the orphan. If field testing shows orphans persist, follow up with a dedicated migration sub-reconciler — out of scope here.
- Renaming the CRD itself, the `Keystone` Go type, or the `keystone.c5c3.io` API group. The change is strictly to operator-managed sub-resource names, not the CR.
- Changing the public URL or the `spec.gateway.hostname` semantics. CC-0088 already settled the public posture; this issue only aligns the internal posture with it.
- Touching `keystone_authtoken` configuration in any other OpenStack service. The catalog self-heal via `keystone-manage bootstrap` covers Keystone itself; downstream services that hardcode `keystone-api.<ns>.svc.cluster.local` in their config are outside this operator's surface area and need their own updates by their respective operators or operators of operators.
- Backwards-compatibility shim: the operator will not create both `<name>` and `<name>-api` Services in parallel. The rename is a clean cut at one Generation boundary, documented in the upgrade flow.

**References**:
- CC-0088 / #282 / commit `88323b9` — the public-side rename to `keystone.127-0-0-1.nip.io` that motivates internal/public symmetry
- `apiResourceName()` central helper: `operators/keystone/internal/controller/reconcile_hpa.go:26-28`
- `bootstrapServiceURL()` hardcoded pattern: `operators/keystone/internal/controller/reconcile_bootstrap.go:67-68`
- Container/port name literals: `operators/keystone/internal/controller/reconcile_deployment.go:171,178`
- Quick Start internal-URL examples: `docs/quick-start.md:638,699,736-740`

**Labels:** enhancement

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/285


berendt:



## WARNING (3)

### W-001: apiResourceName() doc comment broken mid-sentence by inline marker
**File**: `operators/keystone/internal/controller/reconcile_hpa.go:23-31` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Avoid brittle line-number references in comments

**Problem**: The canonical doc comment for the central apiResourceName() helper has the 'CC-0095 legacy' annotation jammed into the middle of a clause, interrupting 'Service' from 'DNS'. Reading the comment top-to-bottom is jarring, and this is the function's primary documentation — every reviewer landing here will trip on it.

**Code**:
```
// apiResourceName returns the canonical name for Keystone operator-managed
// sub-resources (Deployment, HPA, Service, PodDisruptionBudget, NetworkPolicy,
// HTTPRoute). Centralised here so the naming convention is defined in one
// place. Since CC-0095 the helper returns the bare CR name with no suffix —
// the historical "{name}-api" framing was dropped to align internal Service // CC-0095 legacy: pre-rename name referenced for traceability.
// DNS with the public hostname posture established by CC-0088.
func apiResourceName(keystone *keystonev1alpha1.Keystone) string {
	return keystone.Name
}
```

**Action Required**: Restructure the comment so the literal '{name}-api' sits at end-of-clause and the marker follows naturally on the same line.

**Suggested Fix**:
```
// apiResourceName returns the canonical name for Keystone operator-managed
// sub-resources (Deployment, HPA, Service, PodDisruptionBudget, NetworkPolicy,
// HTTPRoute). Centralised here so the naming convention is defined in one
// place. Since CC-0095 the helper returns the bare CR name with no suffix,
// aligning internal Service DNS with the public hostname posture established
// by CC-0088 and dropping the historical "{name}-api" framing. // CC-0095 legacy: pre-rename name referenced for traceability.
func apiResourceName(keystone *keystonev1alpha1.Keystone) string {
	return keystone.Name
}
```

**Related**: W-002, I-003

---

### W-002: Awkward inline markers break sentences in test comment blocks
**File**: `operators/keystone/internal/controller/reconcile_httproute_test.go:843-847` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: Same issue as the apiResourceName comment: the 'CC-0095 legacy' marker is necessary to satisfy the residue scanner, but its placement breaks reading flow. Affects multiple locations: reconcile_httproute_test.go:843-847 (parenthetical interrupted), reconcile_httproute_test.go:868-872 (mid-clause), reconcile_deployment_test.go:1606-1611 (mid-parenthetical), and integration_test.go:3776-3788 (mid-sentence).

**Code**:
```
// (renamed from `<cr-name>-api` // CC-0095 legacy: pre-rename name referenced for traceability.
// by CC-0095, REQ-004):
```

**Action Required**: Restructure each comment so the literal '<cr-name>-api' sits at end-of-line (period or end-of-paren) and the marker follows naturally. Apply to all four locations.

**Suggested Fix**:
```
// TestBuildKeystoneDeployment_NameMatchesCR pins the Deployment ObjectMeta.Name
// to the bare CR name (no `-api` suffix). Symmetric with the Service, PDB, and
// HPA name guards: together they assert the operator emits sub-resources at
// `<cr-name>` rather than the legacy `<cr-name>-api` form (CC-0095, REQ-003 // CC-0095 legacy: pre-rename name referenced for traceability.
// and REQ-004).
```

**Related**: W-001, I-003

---

### W-003: Quick-start fallback test no longer pins the bare service name precisely
**File**: `tests/unit/docs/quick_start_access_section_test.sh:137` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: Use exact-match assertions instead of loose comparisons

**Problem**: The string 'kubectl port-forward svc/keystone' is a substring of 'kubectl port-forward svc/keystone-api'. After this PR the test passes for the new svc/keystone form, but it would also pass if a future regression reintroduced svc/keystone-api — exactly the scenario this test should guard against. The negative-assertion at line 81 already correctly catches the regression because 'keystone' is a substring of 'keystone-api', so only the positive Fallback assertion at line 137 is loose.

**Code**:
```
  if echo "$fallback_tail" | grep -q 'kubectl port-forward svc/keystone'; then
    echo "  PASS: Fallback section contains the port-forward command"
```

**Action Required**: Anchor the grep on the trailing space before the namespace flag so the assertion pins the bare service name.

**Suggested Fix**:
```
  if echo "$fallback_tail" | grep -q 'kubectl port-forward svc/keystone '; then
    echo "  PASS: Fallback section contains the port-forward command"
```

---

## INFO (5)

### I-001: Unnecessary marker in gateway-quick-start-smoke chainsaw-test.yaml
**File**: `tests/e2e/keystone/gateway-quick-start-smoke/chainsaw-test.yaml:280-282` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The residue scanner's literal pattern is 'keystone-api' and templated patterns are '{name}-api|<name>-api|<cr-name>-api|{cr-name}-api'. The line contains only `-api` (in backticks, no prefix) which matches none of these — the marker is unnecessary. Additionally, using C-style '//' inside a YAML '#' comment is inconsistent; YAML uses '#'.

**Code**:
```
        # `keystone-smoke` (matches the Keystone CR name; CC-0095
        # dropped the historical `-api` suffix). // CC-0095 legacy: pre-rename suffix referenced for traceability.
```

**Action Required**: Remove the unnecessary CC-0095 legacy marker entirely.

**Suggested Fix**:
```
        # `keystone-smoke` (matches the Keystone CR name; CC-0095
        # dropped the historical `-api` suffix).
```

---

### I-002: Acceptance criterion 'old <name>-api objects GC'd by ownerReference cascade' is unachievable
**File**: `docs/reference/keystone-upgrade-flow.md:557-602` — **Fix**: ASK — **Confidence**: verified

**Problem**: The PR description's AC says the old <name>-api objects are GC'd by ownerReference cascade. This cannot work: the legacy objects' ownerReferences point to the live Keystone CR which is still present. GC only deletes orphans whose owners are gone. The PR documents this correctly and provides a manual cleanup runbook.

**Code**:
```
Acceptance criterion from PR description: "the old `<name>-api` objects are GC'd by ownerReference cascade (or explicitly deleted by a one-shot migration step in the controller — see Non-Goals)"
```

**Action Required**: Update the GitHub issue / PR description to reflect that cascade-GC was found infeasible. Consider opening a follow-up ticket for the cleanup sub-reconciler mentioned in the warning block at line 601. No code change required.

**Suggested Fix**:
```
Update PR description / linked issue to clarify cascade-GC is infeasible and manual cleanup runbook in keystone-upgrade-flow.md is the supported path. Optionally open follow-up ticket for cleanup sub-reconciler.
```

---

### I-003: Self-referential markers triggered by the marker text itself
**File**: `operators/keystone/internal/controller/reconcile_deployment_test.go:1582` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The CC-0095 legacy marker is required at lines 1582 and 1601 because the marker text itself contains 'keystone-api'. The production code line and assertion message are clean — only the annotation comment forces the marker. This is harmless but slightly tautological.

**Code**:
```
g.Expect(deploy.Spec.Template.Spec.Containers[0].Name).To(Equal("keystone"),
    "container Name must be 'keystone' (renamed by CC-0095, REQ-003)") // CC-0095 legacy: legacy name "keystone-api" referenced for traceability.
```

**Action Required**: Rephrase the marker to avoid mentioning the literal 'keystone-api', so the marker becomes self-justifying rather than self-triggering.

**Suggested Fix**:
```
g.Expect(deploy.Spec.Template.Spec.Containers[0].Name).To(Equal("keystone"),
    "container Name must be 'keystone' (renamed by CC-0095, REQ-003)") // CC-0095 legacy: pre-rename name referenced for traceability.
```

**Related**: W-001, W-002

---

### I-004: chaos-e2e-tests.md frontmatter omits CC-0095
**File**: `docs/reference/chaos-e2e-tests.md:4` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The 'feature:' frontmatter lists CC-0047, CC-0048, CC-0049, CC-0054, CC-0066 but the file body was modified for CC-0095 (line 480 plus chaos-test table updates). Other docs that touched CC-0095 added it to their frontmatter (keystone-crd.md, keystone-reconciler.md, keystone-upgrade-flow.md). For consistency, CC-0095 should be added.

**Code**:
```
feature: CC-0047, CC-0048, CC-0049, CC-0054, CC-0066
```

**Action Required**: Append CC-0095 to the feature list in the frontmatter.

**Suggested Fix**:
```
feature: CC-0047, CC-0048, CC-0049, CC-0054, CC-0066, CC-0095
```

---

### I-005: bootstrapServiceURL consistency assertion uses loose ContainSubstring
**File**: `operators/keystone/internal/controller/reconcile_bootstrap_test.go:487-549` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: TestBootstrapServiceURL_ComposesViaHelper uses ContainSubstring(apiResourceName(ks)) which is too loose to actually pin the relationship. If apiResourceName returns 'keystone' and bootstrapServiceURL embeds 'keystoneapi' (no dot), the assertion still passes. The neighboring literal-equality assertion does provide the real guard, so the second assertion is mostly redundant.

**Code**:
```
g.Expect(url).To(ContainSubstring(apiResourceName(ks)),
    "bootstrapServiceURL host segment must equal apiResourceName(ks) so future helper changes propagate (CC-0095, REQ-002)")
```

**Action Required**: Either tighten the assertion to MatchRegexp(`/`+apiResourceName(ks)+`\.`) to pin the host segment boundary, or drop the redundant assertion. Not blocking.

**Suggested Fix**:
```
Replace the loose ContainSubstring assertion with MatchRegexp(`/`+apiResourceName(ks)+`\.`) to anchor on the host segment boundary, or remove the redundant assertion since the neighboring literal-equality check already provides the real guard.
```

---

## Summary

This is a clean, thorough rename PR with strong test discipline (per-builder name guards, dual unit + integration coverage, residue scanner CI gate, doc-section assertion tests) and a well-thought-out manual upgrade runbook for the orphan-cleanup gap. The L1->L5 commit progression makes review easy. The main quality concerns are awkward inline placement of CC-0095 legacy markers in 5 Go comment blocks (notably the canonical apiResourceName() doc comment) interrupting sentence flow, and a loosened Quick Start fallback test assertion that no longer pins the bare service name precisely. The 'OwnerReference cascade GC' acceptance criterion is physically impossible but the PR correctly handled it via manual runbook documentation.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 5 |

> [!IMPORTANT]
> **Recommendation**: Address Findings 1, 2, and 5 before merge; the rest are polish/follow-up that can be handled in a follow-up PR or left as-is.

Issue #286, opened 2026-04-26
Make Trust Flush active by default so expired trust delegations are cleaned up without explicit opt-in


berendt:
**Category**: enhancement | **Scope**: Medium

**Description**: Flip the default semantics of `spec.trustFlush` from *opt-in* to *active by default* so that a freshly bootstrapped Keystone CR cleans up expired `trust` delegations from day one, without the operator having to remember to add a `trustFlush:` block to their CR. Today `spec.trustFlush` is a nil-able pointer (`*TrustFlushSpec` at `operators/keystone/api/v1alpha1/keystone_types.go:75-79`), and the reconciler explicitly skips CronJob creation when the pointer is nil — `reconcile_trustflush.go:34-47` sets `TrustFlushReady=True` with `reason: TrustFlushNotRequired, message: \"Trust flush is not configured\"` and returns. As a result, every Keystone CR that does not include `spec.trustFlush` accumulates expired trust rows in the `trust` table indefinitely, growing the table monotonically and degrading queries that scan it (`keystone-manage trust_flush` is the only supported way to GC that table — Keystone does not delete rows on token expiry).

The CRD already declares the right defaults at the leaf level: `Schedule` carries `+kubebuilder:default=\"0 * * * *\"` (`keystone_types.go:340-342`) and `Suspend` carries `+kubebuilder:default=false` (`keystone_types.go:344-346`). Those defaults only fire once the parent struct is materialized. Today nothing materializes it on the unset path, so the leaf defaults are dead code for any CR that omits the field. This issue closes that gap by having the defaulting webhook materialize an empty `TrustFlushSpec` when the pointer is nil — the CRD-layer leaf defaults then fill in `schedule`/`suspend`, and the reconciler's existing Path 1 (`reconcile_trustflush.go:49-61`) takes over and creates the CronJob with the documented hourly schedule.

Concrete steps:

1. Add a defaulting block to `KeystoneWebhook.Default()` at `operators/keystone/api/v1alpha1/keystone_webhook.go:80-132`, alongside the existing `Replicas`, `Bootstrap.AdminUser`, `Cache.Backend` defaults:
   ```go
   if obj.Spec.TrustFlush == nil {
       obj.Spec.TrustFlush = &TrustFlushSpec{
           Schedule: \"0 * * * *\",
           Suspend:  false,
       }
   }
   ```
[...]

PR #289, opened 2026-04-26
feat(CC-0096): Enable Trust Flush by default for expired delegations

berendt:
# GitHub Issue #286: Make Trust Flush active by default so expired trust delegations are cleaned up without explicit opt-in

**Description**: Flip the default semantics of `spec.trustFlush` from *opt-in* to *active by default* so that a freshly bootstrapped Keystone CR cleans up expired `trust` delegations from day one, without the operator having to remember to add a `trustFlush:` block to their CR. Today `spec.trustFlush` is a nil-able pointer (`*TrustFlushSpec` at `operators/keystone/api/v1alpha1/keystone_types.go:75-79`), and the reconciler explicitly skips CronJob creation when the pointer is nil — `reconcile_trustflush.go:34-47` sets `TrustFlushReady=True` with `reason: TrustFlushNotRequired, message: \"Trust flush is not configured\"` and returns. As a result, every Keystone CR that does not include `spec.trustFlush` accumulates expired trust rows in the `trust` table indefinitely, growing the table monotonically and degrading queries that scan it (`keystone-manage trust_flush` is the only supported way to GC that table — Keystone does not delete rows on token expiry).

The CRD already declares the right defaults at the leaf level: `Schedule` carries `+kubebuilder:default=\"0 * * * *\"` (`keystone_types.go:340-342`) and `Suspend` carries `+kubebuilder:default=false` (`keystone_types.go:344-346`). Those defaults only fire once the parent struct is materialized. Today nothing materializes it on the unset path, so the leaf defaults are dead code for any CR that omits the field. This issue closes that gap by having the defaulting webhook materialize an empty `TrustFlushSpec` when the pointer is nil — the CRD-layer leaf defaults then fill in `schedule`/`suspend`, and the reconciler's existing Path 1 (`reconcile_trustflush.go:49-61`) takes over and creates the CronJob with the documented hourly schedule.

Concrete steps:

1. Add a defaulting block to `KeystoneWebhook.Default()` at `operators/keystone/api/v1alpha1/keystone_webhook.go:80-132`, alongside the existing `Replicas`, `Bootstrap.AdminUser`, `Cache.Backend` defaults:
   ```go
   if obj.Spec.TrustFlush == nil {
       obj.Spec.TrustFlush = &TrustFlushSpec{
           Schedule: \"0 * * * *\",
           Suspend:  false,
       }
   }
   ```
   Setting `Schedule` and `Suspend` explicitly here avoids relying on the CRD-layer marker firing in the same admission round-trip — the webhook is the single source of truth, mirroring the `Replicas = 3` / `Bootstrap.Region = \"RegionOne\"` style at `keystone_webhook.go:81-98`. The `+kubebuilder:default` markers stay in place as a defense-in-depth fallback for callers that bypass the webhook (e.g. envtest without the defaulting webhook wired up).
2. Update the field godoc on `KeystoneSpec.TrustFlush` (`keystone_types.go:75-79`) and on `TrustFlushSpec` (`keystone_types.go:335-339`) to reflect the new posture: *\"When unset, the defaulting webhook materializes a populated TrustFlushSpec so the operator runs `keystone-manage trust_flush` hourly by default. To pause trust flush without removing the CronJob, set `suspend: true`. There is no nil-back path — admission re-defaults the pointer.\"*
3. Drop the `Path 2` branch from `reconcile_trustflush.go:34-47` — once the webhook always materializes the pointer, `keystone.Spec.TrustFlush == nil` is unreachable in normal operation. Keep `deleteCronJob()` itself for finalizer/garbage-collection paths and for envtest/unit-test scenarios that construct CRs without the webhook in the loop. Replace the inline nil-branch with a one-line guard that defends against the bypass case (envtest, kubectl-without-admission-webhook): if `keystone.Spec.TrustFlush == nil`, log a warning and treat it as the (now legacy) opt-out path so the controller does not panic.
4. Decide what `TrustFlushReady` should report when the new default is active. Today the two outcomes are `TrustFlushReady` (CronJob present) and `TrustFlushNotRequired` (CronJob absent). After the flip, the latter only fires on the legacy bypass path. Recommendation: keep both reasons (the constant strings are referenced by `reconcile_trustflush_test.go:104,182,230` and by `tests/e2e/keystone/trust-flush/chainsaw-test.yaml`) and document that `TrustFlushNotRequired` is now a legacy bypass-only signal. Do **not** introduce a new `TrustFlushSuspended` reason — `suspend: true` keeps the CronJob and the reason should remain `TrustFlushReady`, mirroring how an Autoscaling-suspended HPA still reports `AutoscalingReady`.
5. Update the integration tests:
   - `operators/keystone/internal/controller/reconcile_trustflush_test.go` — the three `TrustFlushNil_*` cases (`TestReconcileTrustFlush_TrustFlushNil_NoExistingCronJob_SetsTrustFlushNotRequired` at `:168`, plus the deletion-path test at `:226`) construct the CR by hand and bypass the webhook. They will keep working as-is and continue to exercise the legacy bypass branch from step 3 — annotate them with a comment noting the bypass posture so future readers do not reintroduce the now-deleted production code path.
   - `operators/keystone/internal/controller/keystone_controller_test.go` and `integration_test.go` — any test that asserts on `TrustFlushReady` reason needs a one-line audit. Cases that go through the webhook (admission-enabled envtest fixtures) will start reporting `TrustFlushReady` instead of `TrustFlushNotRequired`. Update assertions accordingly.
   - `operators/keystone/api/v1alpha1/keystone_webhook_test.go` — add a `TestDefault_TrustFlushNil_MaterializesDefaultSpec` case mirroring the existing `Replicas`/`AdminUser` defaulting tests, asserting that `spec.trustFlush` is non-nil with `schedule == \"0 * * * *\"` and `suspend == false` after `Default()`.
6. Update the e2e fixture at `tests/e2e/keystone/trust-flush/00-keystone-cr.yaml:30-32`. The explicit `trustFlush:` block remains valid and exercises the user-set path. Add a sibling fixture (e.g. `tests/e2e/keystone/trust-flush-default/00-keystone-cr.yaml`) that omits `spec.trustFlush` entirely and asserts (a) the `{name}-trust-flush` CronJob is created with `spec.schedule == \"0 * * * *\"`, (b) `TrustFlushReady=True` with reason `TrustFlushReady`, and (c) the `keystone` CR's serialized spec at the API server contains the materialized `trustFlush:` block (`kubectl get keystone -o yaml`). The existing `01-patch-disable-trust-flush.yaml:12` patches `trustFlush: null` to disable — after the flip, the webhook re-defaults that to a populated struct on the next admission. Replace that patch with `trustFlush: {suspend: true}` and update the matching assertions in `chainsaw-test.yaml:136+` to expect `TrustFlushReady` (not `TrustFlushNotRequired`) plus a CronJob with `spec.suspend == true` and zero recent jobs. Document the rationale in the chainsaw test header.
7. Update the docs:
   - `docs/reference/keystone-crd.md:144,464-468` — change the row description from *\"When set, the operator creates a CronJob…\"* to *\"Defaults to a populated struct (hourly schedule). To pause without removing the CronJob, set `suspend: true`. The field cannot be unset back to nil — admission re-defaults it.\"* Update the `TrustFlushSpec` section narrative at `:461-468` and the CronJob mapping table accordingly.
   - `docs/quick-start.md:589` — change *\"Trust-flush CronJob created (if `spec.trustFlush` is set)\"* to *\"Trust-flush CronJob created — defaults to hourly\"*.
   - `docs/guides/observability.md:60` — same edit as the Quick Start row.
   - `docs/guides/advanced-configuration.md:192` — note the new default-on posture and link to `suspend: true` as the documented way to pause.
   - `docs/reference/keystone-reconciler.md:280,2406-2418,2628` — adjust the sub-reconciler description so the *Path 1*/*Path 2* split is described as legacy-bypass-only.
8. Document the **breaking-on-upgrade** behavior in `docs/reference/keystone-upgrade-flow.md`. Existing CRs that omit `spec.trustFlush` on a cluster running the old operator version will get a populated `spec.trustFlush` materialized at the next admission round-trip after the operator upgrade (any reconcile that round-trips the CR through the API server, plus any user-issued `kubectl apply`/`kubectl edit`). On the next reconcile after that, the operator creates the `{name}-trust-flush` CronJob, and within the configured interval the first `keystone-manage trust_flush` Job fires. The first run can be expensive on long-lived clusters that have never run trust flush — the trust table may contain millions of expired rows. Add a one-line callout that recommends running `keystone-manage trust_flush --date <recent>` manually once before the upgrade, or staggering the rollout, on clusters where the trust table is known to be very large.

**Motivation**: Three drivers, in order of weight.

(a) **Operationally correct default.** Trust delegations expire by `expires_at` but Keystone never garbage-collects the rows. Without `keystone-manage trust_flush` running on a schedule, the `trust` table grows unbounded for the lifetime of the deployment. Every Keystone deployment in production needs trust flush enabled; making it opt-in pushes that decision to every operator individually, and the silent-degradation failure mode (slow `trust` queries and growing DB size months after first install) is exactly the kind of thing CRD defaults exist to prevent. The leaf-level kubebuilder defaults on `Schedule` and `Suspend` show the original intent was already a sensible default; the only thing missing is materialization of the parent struct.

(b) **Symmetry with `credentialKeys`.** `CredentialKeysSpec` is a non-pointer field (`keystone_types.go:73`) with `+kubebuilder:default=\"0 0 * * 0\"` on `RotationSchedule` and `+kubebuilder:default=3` on `MaxActiveKeys`. A CR that omits `credentialKeys` still gets credential-key rotation on a weekly schedule because the CRD-layer defaults fire on the materialized non-nil struct. Trust flush is conceptually the same kind of *housekeeping CronJob the operator owns* and should default on for the same reason. The asymmetry today (`fernet` and `credentialKeys` default-on, `trustFlush` default-off) is historical, not principled.

(c) **The current `TrustFlushNotRequired` condition is misleading.** It is reported on every Keystone CR that omits `spec.trustFlush`, and the user-visible message is *\"Trust flush is not configured\"*. That phrasing reads as a positive signal (\"the operator says everything is fine\") when in fact the cluster is running with an unbounded-growth `trust` table. The condition is not actionable — there is no error to surface — so any operator who reads it interprets it as \"no action needed.\" Flipping the default removes the misleading condition from the happy path entirely; it remains as a legacy bypass-only signal that is never seen in production.

**Affected Areas**:
- `operators/keystone/api/v1alpha1/keystone_webhook.go:80-132` — add `if obj.Spec.TrustFlush == nil { obj.Spec.TrustFlush = &TrustFlushSpec{Schedule: \"0 * * * *\", Suspend: false} }` block alongside the existing `Replicas`/`AdminUser`/`Cache.Backend` defaults
- `operators/keystone/api/v1alpha1/keystone_types.go:75-79,335-339` — godoc rewrite reflecting the default-on posture
- `operators/keystone/api/v1alpha1/keystone_webhook_test.go` — add `TestDefault_TrustFlushNil_MaterializesDefaultSpec` next to the existing `TestDefault_*` cases
- `operators/keystone/internal/controller/reconcile_trustflush.go:34-47` — reframe the nil-branch as legacy-bypass-only with a warning log; keep `deleteCronJob()` for finalizer paths
- `operators/keystone/internal/controller/reconcile_trustflush_test.go:168-230` — annotate the `TrustFlushNil_*` cases as bypass-path tests
- `operators/keystone/internal/controller/keystone_controller_test.go`, `integration_test.go` — adjust any assertion that runs through the webhook and expects `reason: TrustFlushNotRequired` to expect `reason: TrustFlushReady`
- `tests/e2e/keystone/trust-flush/01-patch-disable-trust-flush.yaml:12` — change `trustFlush: null` to `trustFlush: {suspend: true}` and update the matching `chainsaw-test.yaml:136+` assertions
- `tests/e2e/keystone/trust-flush-default/` (new) — fixture + chainsaw test for the omitted-`trustFlush` path
- `docs/reference/keystone-crd.md:144,461-475` — narrative + table edits
- `docs/quick-start.md:589`, `docs/guides/observability.md:60`, `docs/guides/advanced-configuration.md:192`, `docs/reference/keystone-reconciler.md:280,2406-2418,2628`, `docs/reference/keystone-e2e-tests.md:64,842-845` — wording updates and the new `trust-flush-default` suite reference
- `docs/reference/keystone-upgrade-flow.md` — upgrade callout for clusters with never-flushed trust tables

**Acceptance Criteria**:
- [ ] `KeystoneWebhook.Default()` materializes `obj.Spec.TrustFlush` to `&TrustFlushSpec{Schedule: \"0 * * * *\", Suspend: false}` when the pointer is nil — verified by a focused unit test
- [ ] Applying a Keystone CR that omits `spec.trustFlush` results in (a) `kubectl get keystone -o yaml` showing a populated `spec.trustFlush:` block, (b) a CronJob `{name}-trust-flush` with `schedule: \"0 * * * *\"` and `suspend: false`, and (c) `TrustFlushReady=True` with `reason: TrustFlushReady`
- [ ] Setting `spec.trustFlush.suspend: true` on an existing CR pauses the CronJob (`spec.suspend == true`) without deleting it; `TrustFlushReady=True` with `reason: TrustFlushReady` is preserved
- [ ] Patching `spec.trustFlush: null` on a webhook-enabled cluster causes admission to re-default the pointer back to a populated struct — the user cannot unset the field
- [ ] On a cluster bypassing the defaulting webhook (envtest without the defaulter wired up, or `--validate=false` paths), the reconciler falls back to the legacy nil-branch, deletes any existing CronJob, sets `TrustFlushReady=True` with `reason: TrustFlushNotRequired`, emits a structured controller-log entry (`reason=webhook-bypass`, plus `namespace`/`name` keys), and records a Kubernetes Warning Event of reason `TrustFlushBypass` on the Keystone CR (visible via `kubectl describe`)
- [ ] The new `tests/e2e/keystone/trust-flush-default` chainsaw suite passes on a kind cluster, asserting the omitted-field happy path end-to-end
- [ ] The existing `tests/e2e/keystone/trust-flush` chainsaw suite still passes, with the disable-path step rewritten to use `suspend: true`
- [ ] All controller-package unit tests pass, including the relabelled bypass-path cases in `reconcile_trustflush_test.go`
- [ ] `docs/reference/keystone-crd.md` documents the new default and the `suspend: true` opt-out; `docs/quick-start.md` no longer suggests trust flush is opt-in; `docs/reference/keystone-upgrade-flow.md` includes the upgrade callout
- [ ] On a brownfield cluster upgraded from the previous operator version, a Keystone CR that previously omitted `spec.trustFlush` ends up with `spec.trustFlush` materialized after the next admission round-trip — verified by issuing `kubectl patch keystone <name> --type=merge -p '{}'` post-upgrade (or any other spec write — `kubectl apply`, `kubectl edit`, GitOps reconciliation), at which point the operator creates the `{name}-trust-flush` CronJob on the next reconcile and the first trust-flush Job fires within the hourly window. Note: the implementation relies on `CREATE`/`UPDATE` admission webhooks, so an unmodified brownfield CR sitting in etcd does not get materialized by operator reconciles alone (status subresource writes bypass the defaulting webhook). Opt-in controller-driven materialization on first reconcile is out-of-scope for this PR and tracked separately

**Non-Goals**:
- Adding a new `Disabled bool` field to `TrustFlushSpec`. The existing `Suspend` field already covers the *\"keep the CronJob, do not run\"* use case, and a `Disabled` field would duplicate that without a clean delete path. If a user genuinely wants no CronJob at all, the answer is *\"this operator owns trust flush; you cannot opt out\"* — same posture as fernet rotation.
- Changing the field shape from `*TrustFlushSpec` to a non-pointer `TrustFlushSpec`. That would be the *cleanest* CRD-level solution (the kubebuilder leaf defaults would fire automatically on materialized empty structs), but it is a deeper API change with knock-on effects on existing clients and on the deepcopy generation. The webhook-Default approach achieves the same observable behavior with a single-line code change and no API-shape break.
- Backporting the new default to older `keystone.openstack.c5c3.io/v1alpha1` clients. The `v1alpha1` API is still pre-stable and operators are expected to track operator releases; a one-line upgrade-flow callout is sufficient.
- Auto-running an initial bulk `keystone-manage trust_flush --date <past>` on the first reconcile after upgrade to drain a backlog. That is an operator-initiated decision (it can be expensive on a never-flushed cluster) and remains documented in the upgrade flow rather than encoded in the controller.
- Surfacing trust-flush job-failure metrics or alerting beyond the existing `TrustFlushReady` condition. Job-level observability is a separate concern (out of scope; would belong to a future *\"surface CronJob outcomes on the CR status\"* enhancement).

**References**:
- CC-0057 — original trust-flush feature (introduced `*TrustFlushSpec` and the `TrustFlushReady`/`TrustFlushNotRequired` conditions)
- `KeystoneSpec.TrustFlush` field: `operators/keystone/api/v1alpha1/keystone_types.go:75-79`
- `TrustFlushSpec` struct with leaf defaults: `operators/keystone/api/v1alpha1/keystone_types.go:335-351`
- Reconciler nil-branch: `operators/keystone/internal/controller/reconcile_trustflush.go:34-47`
- Defaulting webhook anchor: `operators/keystone/api/v1alpha1/keystone_webhook.go:76-132`
- Existing prior art for *\"materialize an empty struct so leaf defaults fire\"* — `Replicas = 3` and `Bootstrap.Region = \"RegionOne\"` defaulting at `keystone_webhook.go:81-98`
- Conceptual symmetry: `CredentialKeysSpec` non-pointer with leaf defaults at `keystone_types.go:73,323-333`
- Existing trust-flush e2e suite: `tests/e2e/keystone/trust-flush/`

**Labels:** enhancement

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/286


berendt:



## WARNING (5)

### W-001: Documentation pre-flush command uses GNU-only `date` flags
**File**: `docs/reference/keystone-upgrade-flow.md:392-395` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The `$(date -u -d '7 days ago' +%FT%TZ)` is evaluated by the operator's local shell before being passed to `kubectl exec`. `date -d` is a GNU coreutils extension; it does not exist on macOS/BSD. Operators following this runbook from a Mac workstation will get `date: illegal option -- d`. Note that `+%FT%TZ` also contains a literal `Z` not a timezone code, but the GNU compatibility issue is the bigger problem.

**Code**:
```
   kubectl exec -n openstack deploy/<name>-api -- \
     keystone-manage trust_flush --date $(date -u -d '7 days ago' +%FT%TZ)
```

**Action Required**: Replace the GNU-only date substitution with a portable form — instruct the operator to substitute the date manually, compute it inline in Python/Go inside the container, or document both GNU and BSD `-v` forms.

**Suggested Fix**:
```
   # Replace 2026-04-19T00:00:00Z with a date 7 days before today (UTC, ISO 8601).
   kubectl exec -n openstack deploy/<name>-api -- \
     keystone-manage trust_flush --date 2026-04-19T00:00:00Z
```

---

### W-002: Bypass-log namespace assertion is vacuous due to substring collision
**File**: `operators/keystone/internal/controller/reconcile_trustflush_test.go:606-607` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The bypass log message in reconcile_trustflush.go:43-48 contains the literal string "defaulting" inside "webhook defaulting did not run". So `ContainSubstring("default")` matches the message text itself, not the namespace value. The assertion would pass even if the keystone.Namespace field were never logged. The CR-name assertion has a similar but milder issue — "test-keystone" is unique enough that it really does only appear from the structured field.

**Code**:
```
   g.Expect(combined).To(ContainSubstring("default"),
       "bypass log must include the CR namespace (CC-0096, REQ-002)")
```

**Action Required**: Assert against the structured-key format funcr emits (e.g. `"namespace"="default"` and `"name"="test-keystone"`) rather than a bare substring that collides with the message text.

**Suggested Fix**:
```
	g.Expect(combined).To(ContainSubstring("\"namespace\"=\"default\""),
		"bypass log must include the CR namespace as a structured key (CC-0096, REQ-002)")
	g.Expect(combined).To(ContainSubstring("\"name\"=\"test-keystone\""),
		"bypass log must include the CR name as a structured key (CC-0096, REQ-002)")
```

**Related**: W-003

---

### W-003: Test name and code comment claim "Warning" but log uses Info level
**File**: `operators/keystone/internal/controller/reconcile_trustflush.go:43-48` — **Fix**: ASK — **Confidence**: verified

**Problem**: The bypass branch calls `log.FromContext(ctx).Info(...)` — Info level. logr does not have a separate "Warning" level (it has V(0), V(1), …). Naming the test `TestReconcileTrustFlush_BypassNil_LogsWarning` and the doc "log a warning" is imprecise — anyone grepping log pipelines for `severity=warning` will not find these entries. The bypass condition (webhook didn't run) is also operationally significant since admission-time validation was skipped, so a Warning Event on the CR would surface that visibly via `kubectl describe`.

**Code**:
```
log.FromContext(ctx).Info(...)  // bypass log call at reconcile_trustflush.go:43-48
// Test: TestReconcileTrustFlush_BypassNil_LogsWarning (reconcile_trustflush_test.go:584-586)
```

**Action Required**: Either rename the test to `TestReconcileTrustFlush_BypassNil_LogsBypass` and revise comments to say "log" not "warning", or (preferred) use `r.Recorder.Event(keystone, corev1.EventTypeWarning, "TrustFlushBypass", "...")` so the bypass surfaces as a real Kubernetes Warning Event on the CR — matching the existing pattern other reconcilers use.

**Suggested Fix**:
```
Decide between (1) terminology cleanup (rename test + comments to "LogsBypass") and (2) emitting a real corev1.EventTypeWarning Event on the keystone CR so the operationally significant bypass is visible in `kubectl describe`. Option 2 matches existing reconciler patterns and is the recommended fix.
```

**Related**: W-002

---

### W-004: Self-negating bullet in upgrade-flow doc
**File**: `docs/reference/keystone-upgrade-flow.md:351-354` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The list "which can happen via any of:" is followed by a bullet that immediately negates itself ("does **not** typically happen"). Readers naturally read it as "this IS one of the ways", then have to backtrack — structurally confusing.

**Code**:
```
- A controller-driven status write that results in a re-admission of the spec
  (does **not** typically happen — status subresource writes bypass the
  defaulting webhook). Operators who want the CronJob created promptly should
  trigger an explicit no-op spec update (see "Forcing materialization" below).
```

**Action Required**: Either remove the self-negating bullet or split it out as a separate Note paragraph that does not appear in the list of ways materialization can be triggered.

**Suggested Fix**:
```
- A fresh `kubectl apply -f` (or GitOps reconciliation) of the CR manifest.
- A `kubectl edit` or `kubectl patch` on the same object.

> **Note:** Status subresource writes (which the controller does on its own)
> bypass the defaulting webhook, so the operator's reconciliation alone will
> not trigger materialization. Use one of the spec writes above, or see
> "Forcing materialization" below.
```

---

### W-005: Acceptance criterion overstates automatic brownfield upgrade behavior
**File**: `(PR description)` — **Fix**: ASK — **Confidence**: likely

**Problem**: The criterion states that on a brownfield cluster upgraded from the previous operator version, a Keystone CR that previously omitted spec.trustFlush ends up with spec.trustFlush materialized and a CronJob created within one reconcile. The implementation correctly relies on admission webhooks (CREATE/UPDATE only), and the docs at keystone-upgrade-flow.md:340-354 explicitly state this requires an external write. So on a brownfield cluster where nothing writes the CR after upgrade, the AC will not be satisfied automatically — operators must trigger materialization. The docs are correct; the AC text is overstated.

**Code**:
```
AC bullet: "On a brownfield cluster upgraded from the previous operator version, a Keystone CR that previously omitted spec.trustFlush ends up with spec.trustFlush materialized, a CronJob created within one reconcile…"
```

**Action Required**: Either reword the AC to match documented reality (e.g. "…ends up with spec.trustFlush materialized after the next admission round-trip — verified by issuing `kubectl patch keystone <name> --type=merge -p '{}'` post-upgrade…") or open a follow-up issue to implement opt-in controller-driven materialization (out of scope for this PR).

**Suggested Fix**:
```
Reword the AC to acknowledge that the materialization happens on the next admission round-trip rather than automatically, or scope a follow-up issue for controller-driven materialization on first reconcile.
```

---

## INFO (3)

### I-001: Path numbering inversion is now confusing
**File**: `operators/keystone/internal/controller/reconcile_trustflush.go:26-62` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The function header comment lists "spec.trustFlush set" first and "spec.trustFlush nil" second, but the body has `// Path 2: legacy bypass ...` first (line 37) and `// Path 1: trust flush configured ...` second (line 62). The labels were correct pre-CC-0096 (where opt-in was less common), but now the canonical/dominant production path is "set" (Path 1) and "nil" (Path 2) is the legacy bypass. Reading the function in source order, you encounter "Path 2" before "Path 1".

**Code**:
```
// Two lifecycle paths:
//   - spec.trustFlush set: create or update CronJob ... (CC-0057, REQ-001)
//   - spec.trustFlush nil: legacy bypass ...  (CC-0057, CC-0096, REQ-002).
// ...
// Path 2: legacy bypass ...   (line 37)
// Path 1: trust flush configured ...   (line 62)
```

**Action Required**: Either swap the in-function ordering so the configured path is handled first and the bypass second, or relabel them as "Production path" and "Legacy bypass" so the numbering doesn't mislead readers.

**Suggested Fix**:
```
// Two lifecycle paths:
//   - spec.trustFlush set (production): create or update CronJob ... (CC-0057, REQ-001)
//   - spec.trustFlush nil (legacy bypass): ... (CC-0057, CC-0096, REQ-002).
// ...
// Production path: trust flush configured ...
// ...
// Legacy bypass: webhook did not run ...
```

**Related**: I-003

---

### I-002: Hourly-tick race in trust-flush chainsaw Step 10
**File**: `tests/e2e/keystone/trust-flush/chainsaw-test.yaml:177-194` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The schedule is hourly (`0 * * * *`). The CronJob is created at Step 1 with `suspend: false`; suspend is applied at Step 7. If Step 1 runs at e.g. `:55` and Step 7 runs after `:00`, the Kubernetes CronJob controller will have created a Job (legitimately, while suspend was still false) before the suspend patch took effect. Step 10's assertion would then fail on that legitimate pre-suspend Job. CI runs are typically <10 minutes, so the probability of crossing a `:00` boundary is roughly 17%.

**Code**:
```
          OWNED_JOBS=$(kubectl get jobs -n "$NAMESPACE" \
            -o jsonpath='{range .items[?(@.metadata.ownerReferences[0].name=="keystone-trust-flush-trust-flush")]}{.metadata.name}{"\n"}{end}' \
            2>/dev/null || true)
          if [ -n "$OWNED_JOBS" ]; then
            echo "FAIL: CronJob created Jobs while suspended:"
```

**Action Required**: Filter the assertion to Jobs created strictly after the suspend patch — capture a timestamp before Step 7, then in Step 10 filter on `metadata.creationTimestamp > $captured`. Alternatively, accept up to one pre-suspend Job and only assert no new jobs appear during a 30s sleep+poll after suspending.

**Suggested Fix**:
```
Track this as a low-priority follow-up: capture a timestamp in Step 7 and filter Job assertions in Step 10 to only those created after that timestamp, or sleep+poll after the suspend patch instead of asserting absolute absence.
```

---

### I-003: Bypass-branch comment is imprecise for developer readers
**File**: `operators/keystone/internal/controller/reconcile_trustflush.go:38` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: The PR comment in reconcile_trustflush.go:38 says the bypass branch "runs only when the webhook did not run", but it also runs when callers programmatically construct a *Keystone and pass it directly to reconcileTrustFlush() (e.g. unit tests). The wording "did not run" is fine for ops/runbook readers but slightly imprecise for developers. Non-blocking — existing bypass-path test comments cover the developer angle.

**Code**:
```
// Path 2: legacy bypass — runs only when the webhook did not run.
```

**Action Required**: Refine the comment to acknowledge that the bypass branch also fires for programmatic callers (e.g. envtest/unit-test paths that bypass admission), not only when the cluster's webhook is misconfigured.

**Suggested Fix**:
```
// Legacy bypass: runs when admission webhooks did not materialize the field —
// either the cluster has no webhook configured, or a programmatic caller (envtest/unit test)
// invoked reconcileTrustFlush directly without going through admission.
```

**Related**: I-001

---

## Summary

Well-scoped PR that flips spec.trustFlush to default-on via a defaulting webhook, with strong defense-in-depth (webhook + leaf kubebuilder defaults) and excellent integration test coverage including a UID identity check that proves no delete-and-recreate. Five WARNING-level items are real but easy to fix: a GNU-only `date -d` example that breaks on macOS (W-001), a vacuous substring assertion that would silently pass without a real namespace key (W-002), Info-level logging mislabeled as "Warning" in test names and docs (W-003), a self-negating doc bullet (W-004), and an acceptance criterion that overstates the brownfield upgrade behavior the implementation actually delivers (W-005). Three INFO items round out the review, including a now-inverted Path 1/Path 2 ordering and a possible hourly-tick race in the chainsaw suspend test.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 5 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Mergeable after addressing the WARNING items. None are blocking, but W-001, W-002, and W-004 should be fixed in this PR (or a fast follow-up commit) before merge to avoid shipping a broken runbook example, a silently-passing test assertion, and a self-contradictory doc bullet. W-003 and W-005 are best resolved with a brief team discussion (Event vs. log terminology; AC wording vs. implementation scope) and may warrant either small adjustments here or a follow-up issue.

operators/keystone/config/crd/bases/keystone.openstack.c5c3.io_keystones.yaml
```diff
@@ -996,9 +996,16 @@ spec:
...
+                  TrustFlush configures periodic purging of expired trust delegations
+                  (CC-0057, CC-0096). The defaulting webhook materializes a populated
+                  TrustFlushSpec when the field is unset so that the operator runs
+                  keystone-manage trust_flush hourly by default — there is no nil-back
+                  path on a webhook-enabled cluster, because admission re-defaults the
+                  pointer if a user patches it to null. To pause the schedule without
+                  removing the CronJob, set suspend: true. The pointer shape is preserved
+                  for backwards compatibility with envtest fixtures that bypass the
+                  defaulting webhook; the reconciler's nil-branch is now a legacy
+                  bypass-only path.
```
gndrmnn: This project has never had a stable release yet. The definitions are fine to break. Nobody uses this productively yet. Remove the comment about backwards compatibility.

operators/keystone/helm/keystone-operator/crds/keystone.openstack.c5c3.io_keystones.yaml
```diff
@@ -999,9 +999,16 @@ spec:
...
+                  TrustFlush configures periodic purging of expired trust delegations
+                  (CC-0057, CC-0096). The defaulting webhook materializes a populated
+                  TrustFlushSpec when the field is unset so that the operator runs
+                  keystone-manage trust_flush hourly by default — there is no nil-back
+                  path on a webhook-enabled cluster, because admission re-defaults the
+                  pointer if a user patches it to null. To pause the schedule without
+                  removing the CronJob, set suspend: true. The pointer shape is preserved
+                  for backwards compatibility with envtest fixtures that bypass the
+                  defaulting webhook; the reconciler's nil-branch is now a legacy
+                  bypass-only path.
```
gndrmnn: This project has never had a stable release yet. The definitions are fine to break. Nobody uses this productively yet. Remove the comment about backwards compatibility.

operators/keystone/api/v1alpha1/keystone_types.go
```diff
@@ -72,9 +72,16 @@ type KeystoneSpec struct {
...
+	// TrustFlush configures periodic purging of expired trust delegations
+	// (CC-0057, CC-0096). The defaulting webhook materializes a populated
+	// TrustFlushSpec when the field is unset so that the operator runs
+	// keystone-manage trust_flush hourly by default — there is no nil-back
+	// path on a webhook-enabled cluster, because admission re-defaults the
+	// pointer if a user patches it to null. To pause the schedule without
+	// removing the CronJob, set suspend: true. The pointer shape is preserved
+	// for backwards compatibility with envtest fixtures that bypass the
+	// defaulting webhook; the reconciler's nil-branch is now a legacy
+	// bypass-only path.
```
gndrmnn: This project has never had a stable release yet. The definitions are fine to break. Nobody uses this productively yet. Remove the comment about backwards compatibility.

PR #299, opened 2026-04-27
ci: add chainsaw-lint job to schema-validate Chainsaw test definitions

berendt:
Lints every tests/**/chainsaw-test.yaml plus both chainsaw-config.yaml files via `chainsaw lint` so typos, removed fields, or schema drift after a chainsaw version bump fail fast — before the cluster-bound e2e-operator and e2e-chaos jobs spin up a kind cluster.

Adds the `chainsaw-lint` Make target (uses the already-pinned chainsaw from install-test-deps, prints output only on failure) and an always-on CI job that reuses the shared testdeps cache. Wired into the `needs:` list of build-e2e-images and e2e-chaos so a schema break blocks the heavy jobs early.

AI-assisted: Claude Code

### DIFF
--- a/.github/actions/setup-e2e-infra/action.yaml
+++ b/.github/actions/setup-e2e-infra/action.yaml
@@ -30,22 +30,10 @@ runs:
     - name: Install Flux CLI
       uses: fluxcd/flux2/action@871be9b40d53627786d3a3835a3ddba1e3234bd2 # v2.8.3
 
-    # Cache test tool binaries installed by hack/install-test-deps.sh (chainsaw,
-    # flux, kind, kubectl). The script has skip-if-correct-version logic, so a
-    # cache hit makes install-test-deps a no-op. Key on the script hash so the
-    # cache auto-invalidates when any pinned version changes.
-    - name: Cache test dependency binaries
-      uses: actions/cache@v5
-      with:
-        path: $HOME/.local/bin
-        key: ${{ runner.os }}-testdeps-${{ hashFiles('hack/install-test-deps.sh') }}
-        restore-keys: ${{ runner.os }}-testdeps-
-
-    - name: Install test dependencies
-      shell: bash
-      run: |
-        make install-test-deps
-        echo "${HOME}/.local/bin" >> "$GITHUB_PATH"
+    # Cache + `make install-test-deps` is shared with the lightweight
+    # chainsaw-lint job — keeping it in one composite action ensures the cache
+    # key/path stay in lockstep across consumers.
+    - uses: ./.github/actions/setup-test-deps
 
     - name: Deploy infrastructure stack
       shell: bash
--- a/.github/actions/setup-test-deps/action.yaml
+++ b/.github/actions/setup-test-deps/action.yaml
@@ -1,0 +1,35 @@
+# SPDX-FileCopyrightText: Copyright 2026 SAP SE or an SAP affiliate company
+#
+# SPDX-License-Identifier: Apache-2.0
+
+# Composite action encapsulating the shared cache + `make install-test-deps`
+# step used by every job that needs the pinned chainsaw/flux/kind/kubectl
+# binaries. Extracted so the cache key, restore-keys, and PATH wiring stay in
+# one place: setup-e2e-infra (cluster-bound jobs) and lightweight jobs like
+# chainsaw-lint both consume this and inherit any future tweaks (key bump,
+# additional pinned tool, etc.) for free.
+
+name: Setup test dependencies
+description: >
+  Restore the shared testdeps cache and run `make install-test-deps` so the
+  pinned chainsaw/flux/kind/kubectl binaries are on PATH. No cluster required.
+
+runs:
+  using: composite
+  steps:
+    # Cache test tool binaries installed by hack/install-test-deps.sh (chainsaw,
+    # flux, kind, kubectl). The script has skip-if-correct-version logic, so a
+    # cache hit makes install-test-deps a no-op. Key on the script hash so the
+    # cache auto-invalidates when any pinned version changes.
+    - name: Cache test dependency binaries
+      uses: actions/cache@v5
+      with:
+        path: $HOME/.local/bin
+        key: ${{ runner.os }}-testdeps-${{ hashFiles('hack/install-test-deps.sh') }}
+        restore-keys: ${{ runner.os }}-testdeps-
+
+    - name: Install test dependencies
+      shell: bash
+      run: |
+        make install-test-deps
+        echo "${HOME}/.local/bin" >> "$GITHUB_PATH"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -219,6 +219,21 @@ jobs:
       - name: Verify CC-0094 invalid-CR fixture drift and structure
         run: make verify-invalid-cr-fixtures
 
+  # Schema-lint every Chainsaw test (tests/**/chainsaw-test.yaml) and config
+  # (tests/{e2e,e2e-chaos}/chainsaw-config.yaml) via `chainsaw lint` so typos,
+  # removed fields, or schema drift after a chainsaw version bump fail fast —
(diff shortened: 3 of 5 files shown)


berendt:



## WARNING (2)

### W-001: Stale CI workflow documentation
**File**: `docs/reference/ci-cd/ci-workflow.md:85-121` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The PR adds an always-on gate job (chainsaw-lint) and modifies the needs: arrays of build-e2e-images and e2e-chaos, but leaves the workflow reference doc untouched. The DAG block now misrepresents reality — the documented needs: arrays omit verify-invalid-cr-fixtures (added by CC-0094) and chainsaw-lint (this PR), and the doc still asserts 'The workflow defines 18 jobs' while ci.yaml currently has 22.

**Code**:
```
Gate Jobs (always run):
  lint ─────────────┐
  format-check ─────┤
  shellcheck ───────┤
  verify-codegen ───┤
  test (matrix) ────┼──> build-e2e-images ──> E2E Jobs
  test-integration ─┘
...
build-e2e-images ──> needs: [changes, lint, shellcheck, test, test-integration, verify-codegen]
e2e-chaos ──────> needs: [changes, lint, shellcheck, test, test-integration, verify-codegen]
```

**Action Required**: Add chainsaw-lint to the Gate Jobs box in the DAG, refresh the two stale needs: arrays, and add a ### chainsaw-lint section after ### shellcheck (modeled on docs/reference/ci-cd/ci-workflow.md:200-210) describing what the job validates and how the cache is shared via setup-test-deps. Also fix the pre-existing CC-0094 omission in the same pass.

**Suggested Fix**:
```
Add chainsaw-lint (and verify-invalid-cr-fixtures) to the Gate Jobs DAG block, update both build-e2e-images and e2e-chaos needs: arrays to match ci.yaml, update the '18 jobs' count to 22, and add a ### chainsaw-lint subsection after ### shellcheck following the established 1-paragraph + step-table template.
```

---

### W-002: find | xargs pipeline lacks pipefail and --no-run-if-empty
**File**: `Makefile:155-160` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: Two related fragility points: (1) Makefile does not set SHELL := bash or .SHELLFLAGS := -e -o pipefail -c, so the recipe runs in /bin/sh without pipefail — if find errored (e.g. tests/ was renamed/removed) but xargs exited 0, the Make target would silently succeed. (2) GNU xargs without -r/--no-run-if-empty will run 'chainsaw lint test -f' once with no argument when find produces nothing, producing a confusing 'flag needs an argument: -f' error rather than a clean diagnostic. Neither failure mode triggers in current CI (47 chainsaw-test.yaml + 2 chainsaw-config.yaml files exist) but they become real if the tests/ tree is restructured.

**Code**:
```
chainsaw-lint:
	@command -v chainsaw >/dev/null 2>&1 || { echo 'chainsaw is not installed; run `make install-test-deps` first' >&2; exit 1; }
	@echo "Linting Chainsaw test definitions..."
	@find tests -type f -name 'chainsaw-test.yaml' -print0 | xargs -0 -n1 chainsaw lint test -f
	@echo "Linting Chainsaw configurations..."
	@find tests -type f -name 'chainsaw-config.yaml' -print0 | xargs -0 -n1 chainsaw lint configuration -f
```

**Action Required**: Add -r/--no-run-if-empty to both xargs invocations so empty find results yield a clean no-op instead of a confusing error. Note -r is GNU-only (BSD xargs on macOS does not support it); a comment marking the recipe as Linux-CI-targeted is acceptable per the existing format-check pattern.

**Suggested Fix**:
```
chainsaw-lint:
	@command -v chainsaw >/dev/null 2>&1 || { echo 'chainsaw is not installed; run `make install-test-deps` first' >&2; exit 1; }
	@echo "Linting Chainsaw test definitions..."
	@find tests -type f -name 'chainsaw-test.yaml' -print0 | xargs -0 -r -n1 chainsaw lint test -f
	@echo "Linting Chainsaw configurations..."
	@find tests -type f -name 'chainsaw-config.yaml' -print0 | xargs -0 -r -n1 chainsaw lint configuration -f
```

---

## INFO (3)

### I-001: Composite action has no inputs block for future opt-out
**File**: `.github/actions/setup-test-deps/action.yaml:12-15` — **Fix**: ASK — **Confidence**: uncertain

**Problem**: The new composite action is correct and well-commented. There is no inputs: block, which is fine because no parameters are needed today, but if someone later wants to opt out of the install step (e.g. a pure cache-restore use case), they'll have to either fork the action or add an input.

**Code**:
```
.github/actions/setup-test-deps/action.yaml:12-15 (composite action body without inputs:)
```

**Action Required**: Speculative future-proofing — no change recommended now. Revisit if a second consumer needs cache-restore-only behavior.

**Suggested Fix**:
```
No change recommended at this time; consider adding an inputs: block (e.g. install-deps: true) only when a concrete second use case appears.
```

---

### I-002: PR description claim 'prints output only on failure' is inaccurate
**File**: `Makefile:155-160` — **Fix**: ASK — **Confidence**: verified

**Problem**: The Makefile recipe always prints two @echo "Linting Chainsaw …" status lines plus chainsaw's own per-file output. The status echos themselves print regardless of outcome, so the PR description's claim of silent-on-success is at minimum mildly inaccurate. Whether chainsaw lint itself is silent on success in v0.2.14 could not be verified without network access.

**Code**:
```
	@echo "Linting Chainsaw test definitions..."
	@find tests -type f -name 'chainsaw-test.yaml' -print0 | xargs -0 -n1 chainsaw lint test -f
	@echo "Linting Chainsaw configurations..."
	@find tests -type f -name 'chainsaw-config.yaml' -print0 | xargs -0 -n1 chainsaw lint configuration -f
```

**Action Required**: Either drop the @echo "Linting …" status lines so the recipe is genuinely silent on success, or soften the PR description to 'minimal output on success'. Not blocking.

**Suggested Fix**:
```
Either remove the two @echo status lines so success is truly silent, or update the PR description to say 'minimal output on success' instead of 'prints output only on failure'.
```

---

### I-003: chainsaw-lint listed in e2e-chaos needs is also transitive via build-e2e-images
**File**: `.github/workflows/ci.yaml:704` — **Fix**: ASK — **Confidence**: verified

**Problem**: build-e2e-images already has chainsaw-lint in its needs, so the dependency is transitive in e2e-chaos. The explicit listing is correct and intentional: the if: guard !contains(needs.*.result, 'failure') only inspects direct needs, so listing chainsaw-lint explicitly ensures a chainsaw-lint failure surfaces in needs.*.result for e2e-chaos's gate. The redundancy is harmless and arguably defensive.

**Code**:
```
needs: [changes, lint, shellcheck, test, test-integration, verify-codegen, chainsaw-lint, build-e2e-images]
```

**Action Required**: No change needed — wiring is correct and defensible.

**Suggested Fix**:
```
No change needed; the explicit listing is intentional so chainsaw-lint failures appear in needs.*.result for the if: gate.
```

---

## Summary

A small, well-scoped CI hardening PR that adds a fast-failing chainsaw-lint gate and cleanly extracts shared cache + install-test-deps logic into a new setup-test-deps composite action consumed by both setup-e2e-infra and the new job. The composite-action refactor, always-on gate pattern, SHA-pinned actions, and explicit wiring into both build-e2e-images and e2e-chaos needs: are all done well. Two small issues stand out: the CI workflow reference doc is now further out of date with the actual job graph (DAG, needs: arrays, job count, and a missing ### chainsaw-lint section), and the find | xargs pipelines would benefit from -r/--no-run-if-empty for friendlier behavior if the tests/ tree ever changes shape.

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 2 |
| INFO     | 3 |

> [!IMPORTANT]
> **Recommendation**: Merge after updating docs/reference/ci-cd/ci-workflow.md to reflect the new chainsaw-lint job (and ideally pick up the pre-existing verify-invalid-cr-fixtures omission in the same pass). The xargs -r hardening and PR-description wording fix are nice-to-have follow-ups and do not block merge.

Issue #300, opened 2026-04-28
Restrict mode on mounted Fernet/credential key Secret volumes to silence Keystone's world-readable key_repository warning


berendt:
**Category**: bug | **Scope**: Small

**Description**: Every Keystone pod reconciled by the operator emits the following warning on startup:

```
2026-04-28 14:32:43.466 12 WARNING keystone.common.fernet_utils [None req-f046276d-3ae1-43e5-9729-054a91337e86 - - - - - -] key_repository is world readable: /etc/keystone/fernet-keys
```

The check originates upstream in `keystone.common.fernet_utils.FernetUtils._check_key_repository` and fires whenever the directory hosting the Fernet repository carries the world-read bit (`stat.S_IROTH`). The exact same `FernetUtils` class is reused by the credential provider against `/etc/keystone/credential-keys` (`keystone.credential.providers.fernet.Provider`), so the credential repository will emit the analogous warning under identical conditions — the fernet line just lands first in the log on cold start. The defect is symmetric and both repositories must be fixed together.

Root cause is the volume layout in `operators/keystone/internal/controller/reconcile_deployment.go:246-261`. Both `SecretVolumeSource` blocks are constructed without a `DefaultMode`, so Kubernetes falls back to the volume default — `0644` for projected files and `0755` for the projected directory, both world-readable. Combined with the absence of `PodSecurityContext.FSGroup` on the pod (`reconcile_deployment.go:166-167` sets only `TerminationGracePeriodSeconds`, `TopologySpreadConstraints`, `PriorityClassName`; the container-level `restrictedSecurityContext()` at `security_context.go:21-35` neither does nor can set `fsGroup`), kubelet leaves both projected directories at `0755`.

The same construction is reused by the rotation CronJob templates:

- `operators/keystone/internal/controller/reconcile_fernet.go:357-378` — `fernet-keys-src` Secret volume with no `DefaultMode`; the init container `cp /fernet-keys-src/* /etc/keystone/fernet-keys/` (`reconcile_fernet.go:314`) targets an `EmptyDir` working copy that inherits a world-readable mode.
- `operators/keystone/internal/controller/reconcile_credential.go` — symmetric credential rotation CronJob with the same defect.

Operationally this is a noise + defence-in-depth issue. The keystone process runs as UID/GID `42424` (`security_context.go:16,25-26`) and `ReadOnlyRootFilesystem=true` plus the PSS Restricted profile (CC-0045) bound the in-cluster blast radius. But:

[...]

PR #301, opened 2026-04-28
feat(CC-0099): Restrict mode on mounted Fernet key Secret volumes

berendt:
# GitHub Issue #300: Restrict mode on mounted Fernet/credential key Secret volumes to silence Keystone's world-readable key_repository warning

**Description**: Every Keystone pod reconciled by the operator emits the following warning on startup:

```
2026-04-28 14:32:43.466 12 WARNING keystone.common.fernet_utils [None req-f046276d-3ae1-43e5-9729-054a91337e86 - - - - - -] key_repository is world readable: /etc/keystone/fernet-keys
```

The check originates upstream in `keystone.common.fernet_utils.FernetUtils._check_key_repository` and fires whenever the directory hosting the Fernet repository carries the world-read bit (`stat.S_IROTH`). The exact same `FernetUtils` class is reused by the credential provider against `/etc/keystone/credential-keys` (`keystone.credential.providers.fernet.Provider`), so the credential repository will emit the analogous warning under identical conditions — the fernet line just lands first in the log on cold start. The defect is symmetric and both repositories must be fixed together.

Root cause is the volume layout in `operators/keystone/internal/controller/reconcile_deployment.go:246-261`. Both `SecretVolumeSource` blocks are constructed without a `DefaultMode`, so Kubernetes falls back to the volume default — `0644` for projected files and `0755` for the projected directory, both world-readable. Combined with the absence of `PodSecurityContext.FSGroup` on the pod (`reconcile_deployment.go:166-167` sets only `TerminationGracePeriodSeconds`, `TopologySpreadConstraints`, `PriorityClassName`; the container-level `restrictedSecurityContext()` at `security_context.go:21-35` neither does nor can set `fsGroup`), kubelet leaves both projected directories at `0755`.

The same construction is reused by the rotation CronJob templates:

- `operators/keystone/internal/controller/reconcile_fernet.go:357-378` — `fernet-keys-src` Secret volume with no `DefaultMode`; the init container `cp /fernet-keys-src/* /etc/keystone/fernet-keys/` (`reconcile_fernet.go:314`) targets an `EmptyDir` working copy that inherits a world-readable mode.
- `operators/keystone/internal/controller/reconcile_credential.go` — symmetric credential rotation CronJob with the same defect.

Operationally this is a noise + defence-in-depth issue. The keystone process runs as UID/GID `42424` (`security_context.go:16,25-26`) and `ReadOnlyRootFilesystem=true` plus the PSS Restricted profile (CC-0045) bound the in-cluster blast radius. But:

- The warning trains operators to ignore `WARNING` lines from `keystone.common.fernet_utils`, masking real issues at the same severity (e.g. owner-mismatch warnings emitted by the same function).
- A future log-based alert or an upstream promotion of the check to `ERROR` would page on every pod start across the fleet.
- Mounted Fernet keys are active token-signing material and credential keys protect every persisted EC2/application credential at rest in the Keystone DB; there is no reason any sidecar, ephemeral debug container, or future init container co-scheduled into the pod's mount namespace should be able to read them.

**Motivation**: Reproduced on a single-replica Keystone pod against current `main`. The warning is deterministic, fires on every cold start before `--master` worker fork, and is currently unblocked only by ignoring it. The fix is mechanical:

1. Set `PodSecurityContext.FSGroup = openstackUID` (42424) on the Keystone pod and on both rotation CronJob pods. With FSGroup set, kubelet projects the Secret volume directory at mode `0750` (owner `root`, group `42424`, group-readable but not world-readable) and applies group-read on files via the standard `defaultMode | 050` projection — the openstack process can still read its own keys via the group bit.
2. Add `DefaultMode: ptr.To(int32(0o400))` on every `fernet-keys` / `credential-keys` `SecretVolumeSource` (effective file mode becomes `0450` after the FSGroup `| 050` projection — owner-read + group-read, no world bits).
3. Update the rotation init container `cp` invocation to `install -m 0400 /fernet-keys-src/* /etc/keystone/fernet-keys/` (or `cp -p` followed by `chmod 0400`) so the EmptyDir working copy is non-world-readable after the rotation pod hands off to keystone-manage.

Both Fernet and credential paths share identical builder shape (`reconcile_fernet.go` ↔ `reconcile_credential.go`), so one symmetric fix covers both.

**Affected Areas**:
- `operators/keystone/internal/controller/reconcile_deployment.go` (`buildKeystoneDeployment`, PodSpec at L166-167, Secret volumes at L246-261)
- `operators/keystone/internal/controller/reconcile_fernet.go` (`fernetRotateCronJob` Pod template L357-378, init container at L314)
- `operators/keystone/internal/controller/reconcile_credential.go` (symmetric credential rotation CronJob)
- `operators/keystone/internal/controller/reconcile_deployment_test.go` (currently asserts volume/mount existence at L280-291, L303-315 but no mode assertions)
- `operators/keystone/internal/controller/reconcile_fernet_test.go`, `reconcile_credential_test.go` (rotation CronJob mode assertions)
- `docs/reference/keystone/keystone-reconciler.md` (volume table currently lists `defaultMode: 0555` for the rotation scripts ConfigMap but no mode for either key Secret volume)

**Acceptance Criteria**:
- [ ] `buildKeystoneDeployment` sets `Spec.Template.Spec.SecurityContext.FSGroup = ptr.To(openstackUID)` (reusing the constant from `security_context.go:16`) and `DefaultMode = ptr.To(int32(0o400))` on both the `fernet-keys` and `credential-keys` `SecretVolumeSource`.
- [ ] Fernet and credential rotation CronJob templates apply the same `FSGroup` on the Pod and the same `DefaultMode` on every `*-keys-src` Secret volume.
- [ ] The rotation init container preserves a non-world-readable mode on the EmptyDir working copy (`install -m 0400 ...` or `cp -p` + `chmod 0400`), so post-rotation the keystone container still mounts `/etc/keystone/fernet-keys` non-world-readable.
- [ ] `reconcile_deployment_test.go` asserts `Spec.Template.Spec.SecurityContext.FSGroup == 42424` and `DefaultMode == 0o400` on both Secret volumes in the rendered Deployment.
- [ ] `reconcile_fernet_test.go` and `reconcile_credential_test.go` assert the same on the rotation CronJob and assert the init-container command argv contains the non-world-readable copy/chmod step.
- [ ] An integration or chainsaw test scrapes the keystone container log for 30 s after pod-Ready and asserts the `key_repository is world readable` substring is absent for *both* `fernet-keys` and `credential-keys` paths (negative assertion for each, since the fernet line currently masks the credential line).
- [ ] No regression in `chainsaw/keystone-fernet-rotation` or `chainsaw/keystone-credential-rotation` — the keystone-manage rotation step still reads/writes the working copy under the new permissions.
- [ ] `docs/reference/keystone/keystone-reconciler.md` documents the chosen mode for both Secret volumes alongside the existing `0555` note for the rotation script ConfigMap, with a one-line rationale referencing Keystone's `_check_key_repository` check.

**Refs**: CC-0045 (PSS Restricted profile / `restrictedSecurityContext`), upstream `keystone/common/fernet_utils.py::_check_key_repository`.

**Labels:** bug

**Labels:** bug

---
Source: https://github.com/C5C3/forge/issues/300


berendt:
> Reviewed by planwerk-review c5c4853 with Claude Code



## WARNING (1)

### W-001: Chainsaw test `--since=120s` creates vacuous-pass risk
**File**: `tests/e2e/keystone/key-repository-mode/chainsaw-test.yaml:87` — **Fix**: AUTO-FIX — **Confidence**: likely

**Problem**: The upstream `key_repository is world readable` warning fires once per uWSGI worker startup within seconds of pod creation. Step 3 only runs after Step 2's `Ready=True` succeeds, and on a slow CI cluster a Keystone CR may legitimately take 2–4 minutes (DB sync Job, bootstrap Job, then API Deployment rollout). If the keystone container started 3+ minutes before the `kubectl logs` call, `--since=120s` discards the startup window entirely, both `grep -cF` calls return 0, and the test passes vacuously even when a regression is present. The other 25 chainsaw tests in `tests/e2e/keystone/` use `--tail=N` only — this PR is the first to use `--since=`, with no rationale stated. `--tail=2000` already bounds the buffer; Keystone produces only a few hundred startup lines.

**Code**:
```
LOGS=$(kubectl logs -n openstack "$POD" -c keystone --tail=2000 --since=120s 2>&1)
```

**Action Required**: Drop the `--since=120s` window at line 87 and at the catch-block dump on line 123, keeping `--tail=2000` only. Update the catch-block heading on line 120 from `(full --since=120s)` to `(last 2000 lines)`.

**Suggested Fix**:
```
          LOGS=$(kubectl logs -n openstack "$POD" -c keystone --tail=2000 2>&1)
```

---

## INFO (7)

### I-001: Frontmatter `feature:` field introduces a new doc-meta convention
**File**: `docs/reference/keystone/keystone-reconciler.md:4` — **Fix**: ASK — **Confidence**: verified

**Problem**: `grep -rln "^feature:" docs/` returns only this file. No other reference doc carries a `feature:` field, no schema or VitePress config consumes it, and there's no precedent for the convention. This PR also retroactively tags 20 prior CC tickets in addition to CC-0099 — pulling unrelated tickets into a single edit. The metadata is harmless but introduces a brand-new convention added to one file with no documentation. Cross-references the project pattern "Enforce ticket scope in PRs"; flagged as INFO since the metadata has no behavioral impact.

**Code**:
```
---
title: Keystone Reconciler Architecture
quadrant: operator
feature: CC-0013, CC-0015, CC-0038, CC-0057, CC-0058, CC-0064, CC-0065, CC-0067, CC-0068, CC-0071, CC-0072, CC-0073, CC-0074, CC-0077, CC-0078, CC-0079, CC-0080, CC-0081, CC-0087, CC-0095, CC-0099
---
```

**Action Required**: Remove the `feature:` line from the frontmatter (CC-0099 is already inline-tagged in the table rows that matter), or open a follow-up to standardise it across `docs/reference/`.

**Suggested Fix**:
```
Remove the `feature:` line from the frontmatter, or open a follow-up issue to standardise the field across `docs/reference/` with a documented schema.
```

---

### I-002: Asymmetric integration-test coverage for credential rotation init container
**File**: `operators/keystone/internal/controller/integration_test.go:846` — **Fix**: ASK — **Confidence**: verified

**Problem**: The integration test asserts the new `install -m 0400` argv only for the fernet rotation CronJob. There is no integration-test coverage of `credentialRotationCronJob`'s init container — the existing `TestIntegration_CronJobDetailedSpec` suite is fernet-only. The new unit test `TestCredentialRotationCronJob_CopyKeysInitContainerPreservesNonWorldReadableMode` covers it at the unit level, so this is a coverage symmetry gap rather than a hole. The project patterns "Apply bug fixes to all instances of the same pattern" and "Apply symmetric validation to all similar inputs" apply if a credential integration test is added later.

**Code**:
```
g.Expect(initContainer.Command).To(Equal([]string{"sh", "-c", "install -m 0400 /fernet-keys-src/* /etc/keystone/fernet-keys/"}))
```

**Action Required**: No fix required for this PR — flagged so the asymmetry is visible if a future integration test is added for credential rotation.

**Suggested Fix**:
```
Optionally extend `TestIntegration_CronJobDetailedSpec` (or add a parallel test) to cover the credential rotation CronJob's init container argv with a matching assertion.
```

---

### I-003: PR description claims effective mode `0o450`; kubelet projection produces `0o440`
**File**: `PR description` — **Fix**: ASK — **Confidence**: verified

**Problem**: The PR body states *"effective file mode becomes `0450` after the FSGroup `| 050` projection"*. Kubelet's actual FSGroup projection mirrors the owner's permission bits onto the group via `mode | (mode & 0700) >> 3` — for `0400` input this yields `0440` (owner-read + group-read), not `0450` (which would imply group-execute). The code is correct (`DefaultMode=0o400`) and the resulting behaviour is what the test asserts; only the PR description mis-states the projected mode.

**Code**:
```
(PR body) effective file mode becomes `0450` after the FSGroup `| 050` projection
```

**Action Required**: Update the PR description to state the effective mode is `0440` (not `0450`) after FSGroup projection. No code change needed.

**Suggested Fix**:
```
Edit the PR description to replace `0450` with `0440` and correct the projection formula to `mode | (mode & 0700) >> 3`.
```

---

### I-001: Positive Deviation: feature: front-matter added in full rather than appended
**File**: `docs/reference/keystone/keystone-reconciler.md:4` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Spec said 'append `, CC-0099` to the front-matter `feature:` list at L4', but the field did not exist on origin/main (only `title:` and `quadrant:` were present). PR adds the entire field with 21 CC features ending in CC-0099: `feature: CC-0013, CC-0015, ..., CC-0099`. This is broader coverage than requested by the specification.

**Code**:
```
feature: CC-0013, CC-0015, ..., CC-0099
```

**Action Required**: No change needed — broader coverage than requested. Documents the doc's full feature provenance instead of just CC-0099.

**Suggested Fix**:
```
No change required. The deviation is positive — it documents the full feature provenance of the doc rather than only CC-0099, which is preferable.
```

---

### I-002: Positive Deviation: chainsaw log scrape uses --since=120s --tail=2000 (broader than spec)
**File**: `tests/e2e/keystone/key-repository-mode/chainsaw-test.yaml:87` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Spec called for `kubectl logs <keystone-pod> -c keystone --tail=500` for 30 s. Implementation uses `kubectl logs ... --tail=2000 --since=120s` — wider line buffer and longer history window. The single-shot scrape after Ready=True reliably captures the early-startup warning that fires before the WSGI master fork.

**Code**:
```
kubectl logs <keystone-pod> -c keystone --tail=2000 --since=120s
```

**Action Required**: No change needed — strictly broader coverage; any warning within a 30s observation window is also captured by this 120s history window.

**Suggested Fix**:
```
No change required. The deviation is strictly broader: a 120s/2000-line window is a superset of the 30s/500-line window the spec requested.
```

---

### I-003: chainsaw failure message names CC-0099 + path + count, but does not echo the offending log line itself
**File**: `tests/e2e/keystone/key-repository-mode/chainsaw-test.yaml:92-94` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: REQ-006 scenario 'Test fails clearly when the warning regresses' calls for 'printing the offending log line'. The chainsaw assertion uses `grep -cF` (count only) and prints e.g. `ERROR (CC-0099 REQ-006): keystone container logged '...' ($FERNET_MATCHES occurrence(s))`. The actual matching log line is not echoed in the assertion error itself. The catch block at lines 113-138 does dump the full keystone container log via `kubectl logs --tail=2000 --since=120s`, so the offending line is still visible in failure diagnostics — but not in the immediate error message.

**Code**:
```
FERNET_MATCHES=$(... | grep -cF '...')
if [ "$FERNET_MATCHES" -gt 0 ]; then
  echo "ERROR (CC-0099 REQ-006): keystone container logged 'key_repository is world readable: /etc/keystone/fernet-keys' ($FERNET_MATCHES occurrence(s))"
```

**Action Required**: Optional: change `grep -cF` to `grep -F` and capture matching lines into a variable, then print `Offending line(s): $MATCHES` before exiting non-zero. Not required — the catch block already provides full log diagnostics on failure.

**Suggested Fix**:
```
Capture `grep -F` output (full matching lines) into a variable instead of just the count, and echo the variable contents before the non-zero exit so the offending log line appears directly in the assertion error rather than only in the catch-block log dump.
```

---

### I-004: Credential init-container argv shape (`sh`, `-c`, …) only loosely asserted in unit tests
**File**: `operators/keystone/internal/controller/reconcile_credential_test.go:814` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: REQ-005 scenario 'Init-container shell command remains POSIX `sh -c`'. The credential unit test only checks `len(initContainer.Command) >= 3` and substring matches on `Command[2]`. It does not assert `Command[0] == "sh"` and `Command[1] == "-c"`. The fernet side has equivalent unit-test coverage plus a stricter integration assertion (`integration_test.go:846` uses full slice equality). Source code in `reconcile_credential.go:329` uses the canonical `[]string{"sh", "-c", ...}` shape, so the requirement is functionally met; only the credential unit test does not lock the argv shape.

**Code**:
```
g.Expect(len(initContainer.Command)).To(BeNumerically(">=", 3))
g.Expect(initContainer.Command[2]).To(ContainSubstring("install -m 0400"))
```

**Action Required**: Add explicit assertions on `Command[0]` and `Command[1]` so a future change to e.g. `bash` or a non-shell entrypoint fails the unit test the same way as the fernet integration test would.

**Suggested Fix**:
```
g.Expect(len(initContainer.Command)).To(BeNumerically(">=", 3))
g.Expect(initContainer.Command[0]).To(Equal("sh"))
g.Expect(initContainer.Command[1]).To(Equal("-c"))
g.Expect(initContainer.Command[2]).To(ContainSubstring("install -m 0400"))
```

---

## Summary

The PR cleanly implements a security-hardening change with strong symmetry across the three Pod templates mounting Fernet/credential keys, comprehensive unit-test coverage including regression guards on adjacent fields (PSS-restricted container context, ConfigMap volume modes), and a focused E2E test. The design choice of Pod-level `FSGroup` plus `DefaultMode: 0o400` is correct: kubelet rewrites mode `0400 → 0440` and chgrp's the volume to GID 42424, so the keystone process reads via the group bit while world bits are zero. The only material finding is the chainsaw test's `--since=120s` window, which makes the negative log assertion vacuously satisfiable on slow CI clusters. (includes adversarial review pass)

| Severity | Count |
|----------|-------|
| BLOCKING | 0 |
| CRITICAL | 0 |
| WARNING  | 1 |
| INFO     | 7 |

> [!IMPORTANT]
> **Recommendation**: Merge after replacing `--tail=2000 --since=120s` with `--tail=2000` in the chainsaw test (and updating the catch-block heading). The remaining INFO findings (new `feature:` frontmatter convention, asymmetric integration coverage, PR-body mode mis-statement) are non-blocking and can be addressed in this PR or a follow-up at the author's discretion.

gndrmnn (approved):

Issue #314, opened 2026-04-30
test(e2e): Assert semantic invariants on Keystone CR status (#277 Phase 1)


berendt:
## Context

Parent: #277, Phase 1 → *Semantic / state checks*.

Today the Keystone E2E suite exercises happy-path lifecycle (`tests/e2e/keystone/basic-deployment/`),
recovery (`missing-secret/`, `release-upgrade/`) and condition shapes (`healthcheck/`, `events/`),
but no test pins down the *invariants* the CR must guarantee at all times:

- `status.endpoint` URL format
- `ownerReferences` on every operator-managed object
- `status.observedGeneration` tracking (CC-0072)
- Monotonicity of `condition.lastTransitionTime`
- Immutability of generated ConfigMaps

Each of these is verified by unit/integration tests (e.g.
`operators/keystone/internal/controller/integration_test.go:622` — `TestIntegration_StatusEndpoint`,
`reconcile_deployment_test.go:178` — `TestReconcileDeployment_OwnerReferences`),
but no E2E gate exists. This sub-issue closes that gap.

## Scope — new assertions

| # | Scenario | Enforced by | Assertion |
|---|----------|-------------|-----------|
| 1 | `status.endpoint` matches the cluster-local URL when `spec.gateway` is unset | `reconcile_httproute.go:96` (`internalAPIURL`), `reconcile_deployment.go:111` | `status.endpoint == "http://<keystone.Name>.<ns>.svc.cluster.local:5000/v3"` |
| 2 | Every operator-managed object carries `ownerReferences[0].kind=Keystone`, `controller=true`, `blockOwnerDeletion=true`, `uid==<CR UID>` | `controllerutil.SetControllerReference` calls in `reconcile_deployment.go`, `reconcile_fernet.go`, `reconcile_credential.go`, `reconcile_networkpolicy.go`, `reconcile_httproute.go`, `reconcile_hpa.go`, `reconcile_dbconnection_secret.go`, `internal/common/config` | For each managed kind {Deployment, Service, ConfigMap (immutable `keystone-config-*`), `<name>-db-connection` Secret, `<name>-fernet-keys` Secret, `<name>-credential-keys` Secret, `<name>-fernet-keys-backup` PushSecret, `<name>-credential-keys-backup` PushSecret, `<name>-fernet-rotation` CronJob+SA+Role+RoleBinding, `<name>-credential-rotation` CronJob+SA+Role+RoleBinding} assert the four owner-reference fields via JMESPath |
[...]

PR #318, opened 2026-04-30
feat(CC-0101): Assert semantic invariants on Keystone CR status

berendt:
# GitHub Issue #314: test(e2e): Assert semantic invariants on Keystone CR status (#277 Phase 1)

## Context

Parent: #277, Phase 1 → *Semantic / state checks*.

Today the Keystone E2E suite exercises happy-path lifecycle (`tests/e2e/keystone/basic-deployment/`),
recovery (`missing-secret/`, `release-upgrade/`) and condition shapes (`healthcheck/`, `events/`),
but no test pins down the *invariants* the CR must guarantee at all times:

- `status.endpoint` URL format
- `ownerReferences` on every operator-managed object
- `status.observedGeneration` tracking (CC-0072)
- Monotonicity of `condition.lastTransitionTime`
- Immutability of generated ConfigMaps

Each of these is verified by unit/integration tests (e.g.
`operators/keystone/internal/controller/integration_test.go:622` — `TestIntegration_StatusEndpoint`,
`reconcile_deployment_test.go:178` — `TestReconcileDeployment_OwnerReferences`),
but no E2E gate exists. This sub-issue closes that gap.

## Scope — new assertions

| # | Scenario | Enforced by | Assertion |
|---|----------|-------------|-----------|
| 1 | `status.endpoint` matches the cluster-local URL when `spec.gateway` is unset | `reconcile_httproute.go:96` (`internalAPIURL`), `reconcile_deployment.go:111` | `status.endpoint == "http://<keystone.Name>.<ns>.svc.cluster.local:5000/v3"` |
| 2 | Every operator-managed object carries `ownerReferences[0].kind=Keystone`, `controller=true`, `blockOwnerDeletion=true`, `uid==<CR UID>` | `controllerutil.SetControllerReference` calls in `reconcile_deployment.go`, `reconcile_fernet.go`, `reconcile_credential.go`, `reconcile_networkpolicy.go`, `reconcile_httproute.go`, `reconcile_hpa.go`, `reconcile_dbconnection_secret.go`, `internal/common/config` | For each managed kind {Deployment, Service, ConfigMap (immutable `keystone-config-*`), `<name>-db-connection` Secret, `<name>-fernet-keys` Secret, `<name>-credential-keys` Secret, `<name>-fernet-keys-backup` PushSecret, `<name>-credential-keys-backup` PushSecret, `<name>-fernet-rotation` CronJob+SA+Role+RoleBinding, `<name>-credential-rotation` CronJob+SA+Role+RoleBinding} assert the four owner-reference fields via JMESPath |
| 3 | `observedGeneration` on every sub-condition catches up with `metadata.generation` after a spec change | `keystone_controller.go:542/550` (Ready), `reconcile_secrets.go:74/93/110/126/142/152` and analogous in every sub-reconciler — CC-0072 | After `kubectl patch keystone ... --type=merge -p '{\"spec\":{\"replicas\":2}}'`, assert `metadata.generation > 1` and for **every** condition `observedGeneration == metadata.generation` |
| 4 | `condition.lastTransitionTime` strictly increases when a condition flips True→False→True | `conditions.SetCondition` only bumps `lastTransitionTime` on a status change (apimachinery convention) | Capture `lastTransitionTime` on `DeploymentReady=True`; force a flip via `kubectl scale deploy <name>-keystone --replicas=0`; wait for `DeploymentReady=False`; reset replicas; wait for `DeploymentReady=True`; assert new `lastTransitionTime > previous` (RFC3339 string compare suffices) |
| 5 | Manual mutation of a generated ConfigMap is rejected by the API server | `internal/common/config.CreateImmutableConfigMap` sets `Immutable: true` | `kubectl patch cm <keystone-config-*> --type=merge -p '{\"data\":{\"keystone.conf\":\"x\"}}'` exits non-zero with stderr containing `\"field is immutable\"` — proves the invariant the operator relies on |

## Out of scope (prerequisites, separate tickets)

One Phase 1 checkbox requires a feature that does not exist today:

- **Secret drift policy on `<name>-fernet-keys` / `<name>-credential-keys`** — neither Secret is
  marked `Immutable` (`reconcile_fernet.go:258` and analogous in `reconcile_credential.go`); the
  operator only writes new keys on rotation, so a manual edit persists between rotations. There
  is no implemented "policy" to revert manual edits. Splitting this off:
  → `feat(CC-xxxx): define and enforce drift policy on rotation Secrets`. The E2E test follows
  once the feature lands.

## Test design

Location: new directory `tests/e2e/keystone/semantic-invariants/` (does **not** extend an existing
test — the assertion surface is wide enough that mixing it into e.g. `basic-deployment` would
dilute that test's intent).

```
tests/e2e/keystone/semantic-invariants/
├── 00-keystone-cr.yaml      (managed DB+cache via clusterRef, gateway unset, replicas: 1)
└── chainsaw-test.yaml       (5 steps, one per scope row)
```

Step pattern:

- **Step 1** — apply CR, assert `Ready=True`.
- **Step 2** — `assert` block: `status.endpoint` literal match.
- **Step 3** — `script` step iterating the managed-kind list with `kubectl get ... -o json | jq -e '.metadata.ownerReferences[0]'`. Failing kinds print to stderr; exit 1 on the first miss.
- **Step 4** — `patch` to `replicas: 2`; `assert` block matching `(conditions[*])` with the JMESPath function `(observedGeneration == metadata.generation)`.
- **Step 5** — `script` step: capture `lastTransitionTime` on `DeploymentReady`, scale Deployment to 0, poll until `DeploymentReady=False`, scale back, poll until `DeploymentReady=True`, fail if new timestamp is not strictly greater than the captured one.
- **Step 6** — `script` step: `kubectl patch cm <keystone-config-*>` and assert non-zero exit with `\"field is immutable\"` in stderr (no resource cleanup needed; the patch is rejected pre-commit).

All steps fit inside the 30 s `apply` / 5 m `assert` budget from `tests/e2e/chainsaw-config.yaml`.
No new infrastructure components.

## Definition of Done

- [ ] `chainsaw test tests/e2e/keystone/semantic-invariants` green locally against `make kind-up`
- [ ] `make e2e` green; no flakes in two consecutive `parallel: 4` CI runs
- [ ] All five scope rows asserted; one prerequisite split into its own follow-up issue and linked here
- [ ] No new infra in `deploy/kind/`; no deviations from `tests/e2e/chainsaw-config.yaml`
- [ ] Parent #277: the four covered checkboxes ticked, the Secret-drift checkbox replaced by a link to the new feature ticket
- [ ] Commit prefix `test(CC-xxxx)` once a feature code is assigned

## Risks / open questions

- **Owner-reference list maintenance**: the managed-kind list is derived from today's reconcile surface. Any future managed object (e.g. ServiceMonitor when CC-0089 lands in-cluster) needs to be appended. Mitigation: a comment in the test enumerates the source files (`reconcile_*.go`) so a reviewer adding a new managed kind sees the test as a tracked counterpart.
- **Step 4 timing**: the flip via `kubectl scale` interacts with the Deployment's `progressDeadlineSeconds`. Using `replicas: 0` avoids the deadline path because scale-down is immediate; please confirm in review.
- **Endpoint string drift**: the cluster DNS suffix `.cluster.local` is hard-coded. Clusters that override `--cluster-domain` would fail this test — out of scope here, the kind test cluster uses the default.

---
Source: https://github.com/C5C3/forge/issues/314


berendt:
> Reviewed by planwerk-review 6bdbeb5 with Claude Code



## BLOCKING (2)

### B-001: Step 6 catch block omits CR status and pod logs (REQ-007 violation)
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:498-510` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Step 6's catch block dumps only ConfigMap metadata and events. REQ-007 requires pod logs, CR status, and recent events on every step failure. The spec's 'Catch block does not require ConfigMap content dump for invariants test' scenario excuses only the content dump — not the CR status (`kubectl get keystone -o yaml`) or pod logs (`kubectl logs --tail=60 --previous`).

**Code**:
```
catch:
- script:
    content: |
      echo "=== Keystone config ConfigMaps (metadata only — data omitted by design) ==="
      kubectl get cm -n $NAMESPACE \
        -l "forge.c5c3.io/config-base=keystone-invariants-config" \
        -o custom-columns=NAME:.metadata.name,IMMUTABLE:.immutable,LABELS:.metadata.labels 2>&1 || true
      echo "=== Events ==="
      kubectl get events -n $NAMESPACE --sort-by='.lastTimestamp' 2>&1 | tail -30 || true
```

**Action Required**: Add `kubectl get keystone keystone-invariants -n $NAMESPACE -o yaml` and `kubectl logs -n $NAMESPACE -l app.kubernetes.io/name=keystone-operator --all-containers --tail=60 --previous` to Step 6's catch script.

**Suggested Fix**:
```
catch:
- script:
    content: |
      echo "=== Keystone CR (full YAML) ==="
      kubectl get keystone keystone-invariants -n $NAMESPACE -o yaml 2>&1 || true
      echo "=== Keystone config ConfigMaps (metadata only — data omitted by design) ==="
      kubectl get cm -n $NAMESPACE \
        -l "forge.c5c3.io/config-base=keystone-invariants-config" \
        -o custom-columns=NAME:.metadata.name,IMMUTABLE:.immutable,LABELS:.metadata.labels 2>&1 || true
      echo "=== Operator pod logs ==="
      kubectl logs -n $NAMESPACE -l app.kubernetes.io/name=keystone-operator --all-containers --tail=60 --previous 2>&1 || true
      echo "=== Events ==="
      kubectl get events -n $NAMESPACE --sort-by='.lastTimestamp' 2>&1 | tail -30 || true
```

**Related**: B-002

---

### B-002: Step 3 and Step 5 catch blocks omit pod logs (REQ-007 violation)
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:259-455` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Neither Step 3's catch (lines 259-268) nor Step 5's catch (lines 449-455) invokes `kubectl logs`. REQ-007 and the planned test spec require pod logs (--tail=60) on every step failure. Both steps test reconciler failures (ownerReferences regression, lastTransitionTime stuck) where operator pod logs are the most diagnostic signal. The healthcheck convention explicitly cited by the spec includes pod logs.

**Code**:
```
echo "=== Keystone CR (full YAML) ==="
kubectl get keystone keystone-invariants -n $NAMESPACE -o yaml 2>&1 || true
echo "=== Managed objects (selected by name prefix) ==="
kubectl get -n $NAMESPACE \
  deploy,svc,pdb,cm,secret,... -o name 2>&1 | grep keystone-invariants || true
echo "=== Events ==="
kubectl get events …
```

**Action Required**: Append `kubectl logs -n $NAMESPACE -l app.kubernetes.io/name=keystone-operator --all-containers --tail=60 --previous` to both Step 3 and Step 5 catch blocks. For Step 5, also add the equivalent for keystone-API pods.

**Suggested Fix**:
```
echo "=== Keystone CR (full YAML) ==="
kubectl get keystone keystone-invariants -n $NAMESPACE -o yaml 2>&1 || true
echo "=== Managed objects (selected by name prefix) ==="
kubectl get -n $NAMESPACE \
  deploy,svc,pdb,cm,secret,... -o name 2>&1 | grep keystone-invariants || true
echo "=== Operator pod logs ==="
kubectl logs -n $NAMESPACE -l app.kubernetes.io/name=keystone-operator --all-containers --tail=60 --previous 2>&1 || true
echo "=== Events ==="
kubectl get events …
```

**Related**: B-001

---

## WARNING (3)

### W-001: Step 3 ownerReferences fan-out is missing the TrustFlush CronJob
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:243-247` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The test's iteration list at lines 215-251 and the source-file enumeration at lines 19-27 (mirrored in docs/reference/testing/keystone-e2e-tests.md:626-635) do not mention reconcile_trustflush.go or the <name>-trust-flush CronJob. The defaulting webhook at operators/keystone/api/v1alpha1/keystone_webhook.go:113-118 materialises spec.trustFlush whenever unset, and the fixture at tests/e2e/keystone/semantic-invariants/00-keystone-cr.yaml does not set it, so reconcile_trustflush.go:78-80 will create the <name>-trust-flush CronJob (owned via job.EnsureCronJob). The test currently does not assert that owner reference, contradicting its stated invariant that Step 3 is the tracked counterpart for owner-ref fan-out.

**Code**:
```
          # Credential rotation RBAC + CronJob (reconcile_credential.go)
          check_owner sa "${NAME}-credential-rotate"
          check_owner role.rbac.authorization.k8s.io "${NAME}-credential-rotate"
          check_owner rolebinding.rbac.authorization.k8s.io "${NAME}-credential-rotate"
          check_owner cronjob "${NAME}-credential-rotate"
```

**Action Required**: Add a check_owner cronjob assertion for ${NAME}-trust-flush after the credential-rotate block, and extend the comment list at lines 19-27 and 124-148 (and the docs at lines 626-635) to include reconcile_trustflush.go.

**Suggested Fix**:
```
          # Credential rotation RBAC + CronJob (reconcile_credential.go)
          check_owner sa "${NAME}-credential-rotate"
          check_owner role.rbac.authorization.k8s.io "${NAME}-credential-rotate"
          check_owner rolebinding.rbac.authorization.k8s.io "${NAME}-credential-rotate"
          check_owner cronjob "${NAME}-credential-rotate"

          # Trust flush CronJob (reconcile_trustflush.go) — webhook defaults
          # spec.trustFlush so this CronJob is always materialised
          check_owner cronjob "${NAME}-trust-flush"
```

**Related**: W-002, I-002

---

### W-002: Step 3 ownerReferences fan-out is missing the MariaDB Database/User/Grant CRs
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:249-251` — **Fix**: AUTO-FIX — **Confidence**: verified

**Problem**: The fixture sets spec.database.clusterRef.name: openstack-db (managed mode). reconcile_database.go:279,295 creates MariaDB Database, User, and Grant CRs via database.EnsureDatabase / database.EnsureDatabaseUser, each carrying the standard controller-set owner reference (internal/common/database/database.go:34,91,123). They are persistent (not transient like Jobs) and the existing tests/e2e/keystone/deletion-cleanup test already enumerates them as owned resources for cascading deletion. The semantic-invariants test does not iterate them in Step 3, and reconcile_database.go is absent from both the in-file source list (lines 19-27) and the docs source list.

**Code**:
```
          # PushSecrets (reconcile_fernet.go, reconcile_credential.go)
          check_owner pushsecret.external-secrets.io "${NAME}-fernet-keys-backup"
          check_owner pushsecret.external-secrets.io "${NAME}-credential-keys-backup"
```

**Action Required**: Add check_owner assertions for the three MariaDB CRs (database, user, grant) after the PushSecrets block, and extend the comment list at lines 19-27 and 124-148 (and the docs at lines 626-635) to include reconcile_database.go.

**Suggested Fix**:
```
          # PushSecrets (reconcile_fernet.go, reconcile_credential.go)
          check_owner pushsecret.external-secrets.io "${NAME}-fernet-keys-backup"
          check_owner pushsecret.external-secrets.io "${NAME}-credential-keys-backup"

          # MariaDB CRs (reconcile_database.go) — managed mode only;
          # fixture uses spec.database.clusterRef so all three are created
          check_owner database.k8s.mariadb.com "${NAME}"
          check_owner user.k8s.mariadb.com "${NAME}"
          check_owner grant.k8s.mariadb.com "${NAME}"
```

**Related**: W-001, I-002

---

### W-001: Step exec-timeout overrides deviate from "no override of timeouts" spec
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:155-360` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Three step-level `timeout:` fields override the default `timeouts.exec: 30s` — Step 3 (`2m` at line 155), Step 4 (`6m` at line 286), Step 5 (`12m` at line 360). The planned test spec for REQ-006 explicitly forbids overrides: "no override of timeouts in the test spec". Step 5's 12m alone is more than 2× the cited 5m assert budget. The pragmatic justification (poll loops can't fit a real Deployment flip into 30s/5m) is sound, but contradicts the spec as written.

**Code**:
```
# Step 3 (line 155): timeout: 2m
# Step 4 (line 286): timeout: 6m
# Step 5 (line 360): timeout: 12m
```

**Action Required**: Either reconcile spec text to explicitly allow per-step exec overrides for poll-driven steps, or restructure the polling logic to fit within ≤5m.

**Suggested Fix**:
```
Reach alignment with the spec author on whether per-step exec overrides are acceptable for poll-driven invariant tests, then either update the spec/planned-test-spec to permit them or rework the polling deadlines.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update spec and planned test spec to permit per-step exec overrides for poll-driven steps; document the rationale (Deployment flip physics). | Honest about the actual physical timeline of a Deployment flip; preserves diagnostic poll cadence; no test rewrite. | Spec churn; future authors may use the carve-out more liberally than intended. | LOW | Implementation drifts from spec; future reviewers either re-flag the same issue or assume the spec is wrong without authority to amend it. |
| B | Tighten polling deadlines (drop Step 5's settle window, run inner polls back-to-back) so each step fits within the 5m assert budget without override. | Spec-conformant as written; smaller suite runtime. | More fragile under cluster latency variance; risk of flaky tests under load; loses the explicit settle window. | MED | If the 5m ceiling is not actually achievable under realistic cluster timing, the suite becomes intermittently flaky and the spec is still violated in spirit. |

**Recommended**: A — The spec's 5m ceiling is physically unworkable for a real Deployment scale-flip-restore cycle, so reconciling the spec to match operational reality is more honest than over-tightening the polling and accepting flakiness.

---

## INFO (4)

### I-001: Step 5 lastTransitionTime equality edge case under second-precision timestamps
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:438-446` — **Fix**: ASK — **Confidence**: verified

**Problem**: The strict monotonicity check uses bash [[ "$T1" > "$T0" ]] (lexicographic) on RFC3339 UTC seconds-precision timestamps. Equality is therefore possible: if the False→True transition completes within the same wall-clock second as the captured T0, the test fails even though the transition actually advanced. In practice, terminationGracePeriodSeconds=30 plus the 5s preStop sleep guarantees several seconds between T0 and T1, so the boundary is unreachable under realistic scheduling — but the comment does not mention the equality semantics and a future timing change could surface it.

**Code**:
```
          # apimachinery emits RFC3339 UTC seconds-precision timestamps;
          # bash `[[ > ]]` does a lexicographic compare which is correct
          # for that format.
          if [[ "$T1" > "$T0" ]]; then
            echo "OK: lastTransitionTime advanced ($T0 -> $T1)"
          else
            echo "FAIL: lastTransitionTime did not advance: T0=$T0 T1=$T1" >&2
            exit 1
          fi
```

**Action Required**: Either extend the inline comment to acknowledge the equality boundary (and document why it is unreachable here), or relax the comparison to >= if same-second equality should be treated as success.

**Suggested Fix**:
```
Extend the comment block to explicitly call out that strict > is intentional and the equality case is unreachable because the True→False→True cycle takes at least terminationGracePeriodSeconds=30s plus the 5s preStop sleep — or relax the comparison to >= and treat same-second transitions as success. Project judgment call on which semantics to enforce.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Keep strict > and extend the inline comment to document why same-second equality is unreachable in this test. | Preserves the strongest possible monotonicity assertion; no behavioural change; minimal diff; future readers understand the boundary. | Still flaky if a future change ever shrinks termination/preStop below ~1s. | LOW | A reader may later question whether the strict > is intentional; if termination timings change in the future, the test could flake silently with no documented reason. |
| B | Relax the comparison to [[ "$T1" >= "$T0" ]] so same-second transitions count as success. | Eliminates any same-second flakiness regardless of future timing changes; matches what the spec actually guarantees (non-decreasing). | Weakens the assertion: a regression where lastTransitionTime is never updated would still pass if T1==T0; no longer proves the controller actually wrote a fresh timestamp. | LOW | Future timing changes (shorter grace period, faster preStop) could make the strict > flaky. |
| C | Capture T0 with subsecond precision (e.g., date +%s.%N at fixture-apply time) instead of relying on the controller's seconds-precision timestamp, then compare numerically. | Removes the equality boundary entirely without weakening the assertion. | Mixes wall-clock time with controller-emitted time, which is a different invariant; more complex bash; risk of clock-skew false positives. | MED | None beyond what A/B already cover. |

**Recommended**: A — The reviewer verified that termination/preStop timing makes equality unreachable in this suite, so the strict > carries real signal (it would catch a controller that fails to refresh the timestamp). A documentation-only change preserves that signal at minimal cost; relaxing to >= would weaken the test against a real failure mode for a boundary that cannot be reached.

---

### I-002: Source-file enumeration duplicated between test and docs
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:14-27` — **Fix**: ASK — **Confidence**: verified

**Problem**: The list of reconcile_*.go files contributing managed kinds is maintained in two places: the chainsaw-test header comment (lines 14-27) and the docs reference (docs/reference/testing/keystone-e2e-tests.md:626-635). Both already drift in the same direction — each is missing reconcile_trustflush.go and reconcile_database.go — which is exactly the failure mode duplication produces. The test's claim that it is the tracked counterpart for owner-ref fan-out only holds if the enumeration is kept in sync, and there is currently no mechanism (or even a cross-reference) ensuring that.

**Action Required**: Either consolidate to a single source of truth (e.g., docs only, with the test header pointing to the docs section) or add a short note in each that the other is the canonical mirror so reviewers know to update both.

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Make the docs the single source of truth; replace the in-file enumeration with a one-line pointer to docs/reference/testing/keystone-e2e-tests.md. | Eliminates duplication entirely; one place to update when a new reconciler lands; docs are already the place reviewers look for the test's contract. | Test reader has to context-switch to the docs to see the list; if the docs are reorganised the pointer breaks. | LOW | Both lists continue to drift independently and the next managed-kind addition will likely miss one of them again. |
| B | Keep both lists but add a cross-reference comment in each pointing to the other as the canonical mirror, plus a checklist item in the PR template / docs that touching one requires touching the other. | Preserves the at-a-glance enumeration in the test file (helpful for reviewers reading chainsaw-test.yaml in isolation); low-risk change. | Still relies on reviewer discipline; cross-references rot too. | LOW | Same drift risk as today, just with an explicit pointer. |
| C | Generate the list mechanically (e.g., a small script that greps reconcile_*.go for SetControllerReference / EnsureCronJob calls and renders the table into both the test header and the docs). | Eliminates drift by construction; future-proof. | Significant effort for a small surface area; introduces a generator and a CI check; overkill for two locations. | HIGH | Manual sync continues to be the failure mode. |

**Recommended**: A — Both enumerations have already drifted in the same way (findings 1 and 2), which is direct evidence the duplication is not being maintained. Consolidating to docs-only with a pointer from the test is the lowest-effort fix that actually removes the drift surface; option B preserves the same failure mode and option C is disproportionate for two locations.

**Related**: W-001, W-002

---

### I-001: Positive Deviation: Step 5 relies on operator-driven restore instead of explicit kubectl scale --replicas=1
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:397-436` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Story 4 AC and Task 2.5(d) prescribe `kubectl scale deploy keystone-invariants --replicas=1` to drive False → True. Implementation instead scales to 0 and lets the operator reconcile back to `spec.replicas=2` (set by Step 4). After Step 4's patch an explicit `--replicas=1` would be reverted by the operator anyway, so the spec's prescribed mechanic is self-defeating. The invariant under test (T1 > T0) is still verified via the natural operator path; the deviation is a strict improvement.

**Code**:
```
# Force True → False. The operator's watch fires from this patch
# and reconciles the Deployment back to spec.replicas=2 (set by
# Step 4), so no explicit scale-back is required to drive the
# False → True transition.
kubectl scale deploy -n "$NAMESPACE" "$NAME" --replicas=0
…
# False → True via operator-driven restore (no explicit scale).
```

**Action Required**: Update Story 4 AC and Task 2.5 wording to reflect the actual mechanic: "scales the Deployment to 0 replicas; the operator reconciles it back to spec.replicas (=2 from Step 4) and DeploymentReady flips True". No code change needed.

**Suggested Fix**:
```
Amend Story 4 AC and Task 2.5(d) in the spec to describe the operator-driven restore (scale → 0; operator reconciles back to spec.replicas=2) rather than an explicit kubectl scale --replicas=1.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update Story 4 AC and Task 2.5 in the spec to match the implemented operator-driven restore mechanic. | Spec reflects what the test actually does; future readers aren't misled; the implementation is already strictly better. | Requires a spec PR. | LOW | Spec/implementation drift accumulates; future contributors may 'fix' the test back to the broken prescribed mechanic. |
| B | Revert the implementation to use kubectl scale --replicas=1 to match the spec literally. | Literal spec compliance. | The scale would be reverted by the operator immediately; test mechanic becomes nonsensical. | LOW | — |

**Recommended**: A — The implementation correctly leverages the operator's reconcile loop and the comments document the choice; the spec text is the artifact that needs to catch up.

---

### I-002: Positive Deviation: Step 3 collects all owner-reference failures before exiting
**File**: `tests/e2e/keystone/semantic-invariants/chainsaw-test.yaml:167-256` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Story 2 AC and REQ-002's second scenario describe failing on the first missing/mismatched ownerReference. Implementation's `check_owner` sets `fail=1` and returns; the final `if [ "$fail" -ne 0 ]` exits after every kind has been checked. All failing GVKs+names are reported in a single run, so a reviewer fixing one regression sees other regressions immediately rather than going through fail/fix iterations. REQ-002's exit-1-with-GVK+name requirement is still satisfied; only the AC's "first" wording is overruled by a strictly-better diagnostic surface.

**Code**:
```
# check_owner sets fail=1 and returns rather than exiting; the final
# if [ "$fail" -ne 0 ] exits after every kind has been checked.
```

**Action Required**: Update Story 2 AC to: "fails the test on any missing or mismatched ownerReference and emits all failing GVK+names". No code change needed.

**Suggested Fix**:
```
Amend Story 2 AC text from "fails on the first missing or mismatched ownerReference" to "fails on any missing or mismatched ownerReference and emits all failing GVK+names".
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update Story 2 AC to describe the all-failures-collected behavior. | Spec reflects strictly-better diagnostic surface; future contributors won't 'fix' it back to fail-on-first. | Requires a spec PR. | LOW | Spec/implementation drift; future reviewers re-flag the same deviation. |
| B | Revert check_owner to exit-on-first-failure to match the spec literally. | Literal spec compliance. | Loses diagnostic value; reviewers hit one failure at a time and need multiple fail/fix iterations. | LOW | — |

**Recommended**: A — Collecting all failures before exiting is a clear improvement over fail-on-first for owner-reference regression diagnosis; the spec wording is the cheaper artifact to update.

---

## Summary

The PR adds a well-structured Chainsaw E2E suite that gates five real status invariants (endpoint URL, ownerReferences fan-out, observedGeneration, lastTransitionTime monotonicity, ConfigMap immutability) with careful timing budgets, diagnostic-rich catch blocks, and a fixture that deliberately keeps optional sub-conditions in NotRequired branches so observedGeneration stays exact. The main gap is that Step 3's ownerReferences enumeration omits the TrustFlush CronJob (always materialised by the defaulting webhook) and the MariaDB Database/User/Grant CRs (managed mode, which the fixture uses), which contradicts the test's own claim of being the tracked counterpart for owner-ref invariants. The remaining items are minor: a seconds-precision equality boundary in the lastTransitionTime check (unreachable in practice but undocumented) and the duplication of the source-file list between the test and the docs. (includes adversarial review pass)

| Severity | Count |
|----------|-------|
| BLOCKING | 2 |
| CRITICAL | 0 |
| WARNING  | 3 |
| INFO     | 4 |

> [!IMPORTANT]
> **Recommendation**: Merge after the two Step 3 ownerReferences omissions (TrustFlush CronJob, MariaDB Database/User/Grant) are added — they are mechanical fixes and bring the test in line with what its header comment advertises. The two INFO items can be addressed in the same PR or as a follow-up at the team's discretion.

gndrmnn (approved):

Issue #317, opened 2026-04-30
test(e2e): Reconcile Keystone CR in PSS-restricted namespace (#277 Phase 1)


berendt:
## Context

Parent: #277, Phase 1 → *Security / compliance baseline*.

The operator already builds restricted PodSecurityContexts everywhere it spawns a Pod —
`operators/keystone/internal/controller/security_context.go:21` (`restrictedSecurityContext`)
is referenced from:

- `reconcile_deployment.go:174` (Keystone API Deployment)
- `reconcile_bootstrap.go:157` (bootstrap Job)
- `reconcile_database.go:806` (db-sync, db-expand, db-migrate, schema-check Jobs)
- `reconcile_fernet.go:315/328` (rotation init/main containers)
- `reconcile_credential.go:318/331` (rotation init/main containers)
- `reconcile_trustflush.go:124` (trust flush CronJob)
- `reconcile_policyvalidation.go:174` (policy validation Job)

The chart's operator pod itself sets pod- and container-level restricted contexts in
`helm/keystone-operator/templates/deployment.yaml` (lines 22-27 and 67-78). Unit tests
(`security_context_test.go`, `reconcile_bootstrap_test.go:319`,
`reconcile_fernet_test.go:558`, `reconcile_credential_test.go:565`) verify each call site.

What's missing: an end-to-end gate that proves the **whole** workload — operator + Keystone API
+ every Job and CronJob — admits cleanly into a namespace labelled
`pod-security.kubernetes.io/enforce=restricted`. A regression where any future reconciler
forgets `restrictedSecurityContext()` would be caught by the API server's PSS admission, but
only if exercised at this enforcement level, which today's tests do not do.

## Scope — new test

A single Chainsaw test directory: `tests/e2e/keystone/pod-security-restricted/`.

Scenario:

1. Create a fresh test namespace labelled `pod-security.kubernetes.io/enforce=restricted, pod-security.kubernetes.io/enforce-version=latest`.
2. Apply a Keystone CR using the basic-deployment fixture, with `database.clusterRef` and `cache.clusterRef` pointing at the existing `openstack-db` / `openstack-memcached` (which already live in the `openstack` namespace and are unaffected by the new namespace's PSS label).
3. Trigger one Fernet rotation manually (`kubectl create job --from=cronjob/<name>-fernet-rotation manual-rotation -n <test-ns>`) so the rotation Pod is exercised under the restricted policy, not just the API Deployment.

Assertion:

[...]

PR #324, opened 2026-05-01
feat(CC-0104): Add e2e test for Keystone CR in PSS-restricted namespace

berendt:
# GitHub Issue #317: test(e2e): Reconcile Keystone CR in PSS-restricted namespace (#277 Phase 1)

## Context

Parent: #277, Phase 1 → *Security / compliance baseline*.

The operator already builds restricted PodSecurityContexts everywhere it spawns a Pod —
`operators/keystone/internal/controller/security_context.go:21` (`restrictedSecurityContext`)
is referenced from:

- `reconcile_deployment.go:174` (Keystone API Deployment)
- `reconcile_bootstrap.go:157` (bootstrap Job)
- `reconcile_database.go:806` (db-sync, db-expand, db-migrate, schema-check Jobs)
- `reconcile_fernet.go:315/328` (rotation init/main containers)
- `reconcile_credential.go:318/331` (rotation init/main containers)
- `reconcile_trustflush.go:124` (trust flush CronJob)
- `reconcile_policyvalidation.go:174` (policy validation Job)

The chart's operator pod itself sets pod- and container-level restricted contexts in
`helm/keystone-operator/templates/deployment.yaml` (lines 22-27 and 67-78). Unit tests
(`security_context_test.go`, `reconcile_bootstrap_test.go:319`,
`reconcile_fernet_test.go:558`, `reconcile_credential_test.go:565`) verify each call site.

What's missing: an end-to-end gate that proves the **whole** workload — operator + Keystone API
+ every Job and CronJob — admits cleanly into a namespace labelled
`pod-security.kubernetes.io/enforce=restricted`. A regression where any future reconciler
forgets `restrictedSecurityContext()` would be caught by the API server's PSS admission, but
only if exercised at this enforcement level, which today's tests do not do.

## Scope — new test

A single Chainsaw test directory: `tests/e2e/keystone/pod-security-restricted/`.

Scenario:

1. Create a fresh test namespace labelled `pod-security.kubernetes.io/enforce=restricted, pod-security.kubernetes.io/enforce-version=latest`.
2. Apply a Keystone CR using the basic-deployment fixture, with `database.clusterRef` and `cache.clusterRef` pointing at the existing `openstack-db` / `openstack-memcached` (which already live in the `openstack` namespace and are unaffected by the new namespace's PSS label).
3. Trigger one Fernet rotation manually (`kubectl create job --from=cronjob/<name>-fernet-rotation manual-rotation -n <test-ns>`) so the rotation Pod is exercised under the restricted policy, not just the API Deployment.

Assertion:

- `(conditions[?type=='Ready']).status == \"True\"`, reason `AllReady`, within 5 min
- `kubectl get events -n <test-ns> --field-selector reason=FailedCreate` returns zero entries with a message containing `violates PodSecurity \"restricted:latest\"`
- The manual rotation Job reaches `succeeded > 0`

Out of scope: PSS at the **audit** or **warn** levels (Kubernetes default-on for any unlabelled
namespace already issues warn-level audit lines). Audit-level coverage is automatic; the
restricted enforce gate is the regression net.

## Out of scope (prerequisites, separate tickets)

None — every Pod spec the reconciler creates already sets `restrictedSecurityContext()`. The
test exercises the existing surface; if a future Pod is added without the helper, this test
catches it on the first failed admission.

## Test design

```
tests/e2e/keystone/pod-security-restricted/
├── 00-namespace.yaml         (Namespace with PSS labels)
├── 01-keystone-cr.yaml       (CR in the new namespace, basic-deployment fixture)
└── chainsaw-test.yaml        (3 steps)
```

Steps:

1. `apply: 00-namespace.yaml` — Chainsaw's per-test namespace plumbing must be **disabled** for this test (`spec.namespace: \"\"` plus an explicit namespace in the manifest), because Chainsaw creates the implicit per-test namespace without PSS labels. Mirror the pattern from `tests/e2e/keystone/namespace-scoped-rbac/` (which also pins its namespace).
2. `apply: 01-keystone-cr.yaml` — assert `Ready=True`.
3. `script` step: trigger the manual rotation Job, wait for `kubectl wait --for=condition=complete job/manual-rotation -n <test-ns> --timeout=2m`, then `kubectl get events -n <test-ns>` and `grep -v \"violates PodSecurity\"` and assert no matching events.

Catch block: `kubectl get pods -n <test-ns> -o yaml | grep -A3 securityContext`,
`kubectl get events -n <test-ns> --sort-by='.lastTimestamp'`. PSS admission rejections produce
a `FailedCreate` event on the parent (Deployment, Job, CronJob) with the violation listed
verbatim, so the catch block surfaces the cause.

The test fits the `tests/e2e/chainsaw-config.yaml` budget (5 min assert).

## Definition of Done

- [ ] `chainsaw test tests/e2e/keystone/pod-security-restricted` green locally
- [ ] `make e2e` green; two consecutive `parallel: 4` CI runs without new flakes
- [ ] Sanity check: temporarily strip `restrictedSecurityContext()` from one reconciler in review, observe this test failing with the expected `violates PodSecurity` event
- [ ] No new infra in `deploy/kind/`; no chart changes; no operator code changes
- [ ] Parent #277: the one covered checkbox ticked
- [ ] Commit prefix `test(CC-xxxx)` once a feature code is assigned

## Risks / open questions

- **Cross-namespace MariaDB/Memcached references**: the CR points its `database.clusterRef` at `openstack/openstack-db`. The MariaDB operator places the resulting DB CRs in the cluster's namespace (the operator's own choice), so the PSS label on the test namespace does not affect them. If a future operator change moves DB CR creation into the consumer's namespace, this test would also need that namespace to admit MariaDB/Memcached operator pods — handle when it happens.
- **Pre-existing operator pod**: the `keystone-operator` Deployment is in `openstack`, not in the test namespace, so PSS does not gate it. This test is about *workload* Pods. A separate follow-up could add a PSS test for the operator chart by installing a release into a restricted namespace; out of scope here.

---
Source: https://github.com/C5C3/forge/issues/317


berendt:
> Reviewed by planwerk-review 9afcbb4 with Claude Code



## BLOCKING (1)

### B-001: Specification Violation: ExternalSecrets used instead of plain Secrets (REQ-003)
**File**: `tests/e2e/keystone/pod-security-restricted/00-namespace.yaml:35-75` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: REQ-003 (SHALL) requires plain Kubernetes Secrets in the test namespace, but the implementation uses ExternalSecrets. The spec's literal requirement is unimplementable: reconcileSecrets calls WaitForExternalSecret (reconcile_secrets.go:86), and that helper (internal/common/secrets/secrets.go:44-58) does a Get on ExternalSecret and returns (false, nil) on NotFound — a plain Secret would never satisfy the gate. The deviation is documented in the file header (lines 14-25) but the spec was not updated to match.

**Code**:
```
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: keystone-admin
  namespace: keystone-pss-restricted-test
...
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: keystone-db
```

**Action Required**: Update the spec (REQ-003 + Implementation Tasks 1.1, 1.2) to recognize that ExternalSecrets are required, or document this as an accepted resolved deviation in the planwerk record. The implementation is functionally correct; the spec rationale is wrong.

**Suggested Fix**:
```
Amend REQ-003 and Implementation Tasks 1.1/1.2 in the planwerk spec to require ExternalSecrets (matching the operator's WaitForExternalSecret gate) rather than plain Secrets with a 32-char hex password.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update REQ-003 and Implementation Tasks 1.1/1.2 in the spec to require ExternalSecrets sourced from openstack-namespace ExternalSecrets | Spec matches reality; future readers see correct requirements; no implementation churn | Requires spec amendment / stakeholder review | LOW | Spec remains misleading; future readers may try to 'fix' the implementation back to plain Secrets and break the test gate |
| B | Leave REQ-003 as-is and record a resolved deviation note in the planwerk record explaining why ExternalSecrets are required | Lighter-weight than full spec rewrite; preserves traceability of the original intent | Spec body still says SHALL plain Secrets; readers must follow the deviation pointer to learn what really shipped | LOW | Same as option A but slightly diluted because most teams check deviations |

**Recommended**: A — The SHALL clause is physically unimplementable given reconcileSecrets/WaitForExternalSecret behavior, so the spec is wrong rather than 'deviated from'. A resolved-deviation note implies the spec was correct but circumstances changed — that framing doesn't fit here.

**Related**: W-002, W-003

---

## CRITICAL (1)

### C-001: Acceptance Criterion Gap: policy-validation Job is not exercised (Story 1)
**File**: `tests/e2e/keystone/pod-security-restricted/01-keystone-cr.yaml:29-51` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Story 1 acceptance criterion explicitly names the policy-validation Job among the workloads the test must exercise, and REQ-004 scenario 2 references PolicyValidReady. The CR fixture has no policyOverrides field, so reconcilePolicyValidation short-circuits (reconcile_policyvalidation.go:46-58) — no Job, no Pod, no PSS admission check. The test claims to be 'the executable regression net for any future reconciler that forgets restrictedSecurityContext()', but the policy-validation call site at reconcile_policyvalidation.go:174 is silently uncovered. The implementation's design notes do not justify this omission (unlike credential-rotation/trust-flush, where the 'shared helper' rationale is documented).

**Code**:
```
spec:
  # 01-keystone-cr.yaml:29-51 — no policyOverrides field present
```

**Action Required**: Add a 00-policy-cm.yaml fixture (mirroring tests/e2e/keystone/policy-overrides/00-policy-cm.yaml) and set spec.policyOverrides.configMapRef in 01-keystone-cr.yaml. Apply the ConfigMap before the CR.

**Suggested Fix**:
```
Create tests/e2e/keystone/pod-security-restricted/00-policy-cm.yaml mirroring tests/e2e/keystone/policy-overrides/00-policy-cm.yaml, and add spec.policyOverrides.configMapRef to 01-keystone-cr.yaml so reconcilePolicyValidation runs buildPolicyValidationJob and the policy-validation Pod is admitted under PSS.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Add 00-policy-cm.yaml fixture and policyOverrides to the CR to exercise the policy-validation Job under PSS | Closes the explicit Story 1 acceptance criterion gap; makes the test live up to its 'regression net' claim; covers the reconcile_policyvalidation.go:174 restrictedSecurityContext() call site | Adds a small fixture; slightly increases test runtime | LOW | A future regression in reconcile_policyvalidation.go that forgets restrictedSecurityContext() will pass this test silently — exactly what the test was supposed to prevent |
| B | Update Story 1 acceptance criteria to remove the policy-validation Job and document the rationale in the test file header | No fixture work; test stays minimal | Reduces test coverage scope; weakens the regression-net claim; requires spec amendment that contradicts the original intent | LOW | Spec remains inaccurate about what the test actually covers |

**Recommended**: A — The user story explicitly names the policy-validation Job, and a known mirror fixture already exists in tests/e2e/keystone/policy-overrides/00-policy-cm.yaml. Closing the gap is cheap and preserves the test's stated purpose as a regression net for every restrictedSecurityContext() call site.

---

## WARNING (6)

### W-001: Test claims overstate coverage of policy-validation Job admission
**File**: `tests/e2e/keystone/pod-security-restricted/chainsaw-test.yaml:149-153` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Cross-check test assertions against their documentation claims

**Problem**: The chainsaw-test.yaml Step 3 comment and keystone-e2e-tests.md both state that the policy-validation Job is one of the Pods admitted under PSS. In reality, reconcilePolicyValidation (operators/keystone/internal/controller/reconcile_policyvalidation.go:46-58) only creates the Job when keystone.Spec.PolicyOverrides != nil. The test's CR (01-keystone-cr.yaml) does not set policyOverrides, so the Job is never created and PSS admission for it is never tested. PolicyValidReady=True/PolicyValidationNotRequired flips True via the no-op path, which is not evidence of PSS admission. The same overstatement could mask a future regression where restrictedSecurityContext() is removed from the policy-validation Job builder: this test would still pass.

**Code**:
```
  # ── Step 3: Apply the Keystone CR, assert Ready=True/AllReady ──────────
  # REQ-002, REQ-004. The Ready=True signal is the primary end-to-end
  # gate — every Pod the reconciler creates (API Deployment, bootstrap Job,
  # db-sync Job, policy-validation Job) must have been admitted under
  # restricted PSS for the sub-conditions to flip True.
```

**Action Required**: Either exercise the policy-validation Job by adding policyOverrides to the test CR plus a minimal policy ConfigMap fixture, or remove the policy-validation Job claim from the chainsaw-test.yaml Step 3 comment and the keystone-e2e-tests.md inventory and Purpose sections.

**Suggested Fix**:
```
Add policyOverrides.configMapRef to 01-keystone-cr.yaml together with a minimal policy ConfigMap fixture so reconcilePolicyValidation actually creates the Job under restricted PSS. This keeps the comment and docs honest by making them describe what the test actually exercises.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Drop the 'policy-validation Job' claim from the Step 3 comment and from keystone-e2e-tests.md (inventory table at :151 and Purpose at :638-646), noting it is not exercised here and leaning on the same 'all Pod-creating reconcilers share restrictedSecurityContext()' rationale already used for credential-rotation and trust-flush. | Smallest diff; aligned with rationale already present in the test docstring. | The test still does not exercise that Pod under PSS. | LOW | An undetected regression in restrictedSecurityContext() for the policy-validation Job ships with a green test that claims to cover it. |
| B | Add policyOverrides.configMapRef to 01-keystone-cr.yaml and bundle a minimal policy ConfigMap fixture so the Job is actually created and admitted under PSS. | Comments and docs become accurate; gives genuine PSS coverage of the policy-validation Job. | Requires a new fixture file and adds a few seconds to the test run. | MED | Same as A — the policy-validation Pod's PSS admission is not actually verified. |
| C | Keep the comments as-is and rely on review-by-review correction. | No diff. | The inaccuracy persists indefinitely. | LOW | Loss of credibility for the test inventory; future readers trust a claim the test does not back. |

**Recommended**: B — The PR's stated goal is that every Pod the reconciler creates admits cleanly under PSS; opting the CR into a tiny policy override actually exercises that path and keeps the assertion language honest. Per the cross-check-test-claims pattern, a doc claim without a matching assertion is a finding regardless of how plausible the underlying behaviour is.

**Related**: W-003, I-001

---

### W-002: Asymmetric cleanupPolicy: Skip with brownfield-database test
**File**: `tests/e2e/keystone/pod-security-restricted/00-brownfield-db-setup.yaml:47-75` — **Fix**: ASK — **Confidence**: uncertain — **Pattern**: Apply bug fixes to all instances of the same pattern

**Problem**: The new test's User and Grant CRs carry cleanupPolicy: Skip, but the existing brownfield-database test's equivalent CRs (tests/e2e/keystone/brownfield-database/00-brownfield-db-setup.yaml:20-46) default to Delete. Both User CRs manage the same canonical MariaDB user 'keystone'. Under parallel: 4 the suites can run concurrently. The PSS test author explicitly protects against this test's teardown affecting a sibling, but the inverse is unprotected: if brownfield-database finishes first and Chainsaw deletes its User CR with default cleanupPolicy: Delete, the MariaDB operator may issue DROP USER 'keystone', surfacing as authentication failures inside the still-running PSS test until its User CR's reconciler restores the user. The exact behaviour depends on MariaDB-operator semantics around two User CRs targeting the same MariaDB user.

**Code**:
```
# `cleanupPolicy: Skip`
# keeps the user alive when this test's User CR is deleted, so a sibling
# test still managing the same user (e.g. brownfield-database) is not
# stranded mid-flight (REQ-008).
```

**Action Required**: Add cleanupPolicy: Skip to the User and Grant in tests/e2e/keystone/brownfield-database/00-brownfield-db-setup.yaml so symmetric protection is in place in both directions.

**Suggested Fix**:
```
Add cleanupPolicy: Skip to the User and Grant resources in tests/e2e/keystone/brownfield-database/00-brownfield-db-setup.yaml (lines 20-46) so the brownfield-database test does not drop the canonical 'keystone' MariaDB user out from under any concurrent suite — including the new PSS test — when it finishes first.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Add cleanupPolicy: Skip to the User and Grant in tests/e2e/keystone/brownfield-database/00-brownfield-db-setup.yaml. | Symmetric protection in both directions; one-line change per CR; future tests sharing the canonical user inherit the safety. | Modifies a test outside this PR's stated scope. | LOW | brownfield-database cleanup can drop the user out from under any concurrent test using it, producing flaky auth failures whose root cause is hard to attribute. |
| B | Give the new PSS test its own MariaDB user (e.g., keystone_pss_restricted_user) instead of sharing 'keystone'. | Complete isolation; eliminates cleanup ordering as a concern entirely. | Requires a separate OpenBao path with that username or a Secret transformation; significant scope creep. | HIGH | Same as A — concurrent runs can race on the shared user. |
| C | Document the residual race in the test's docstring as an accepted risk and keep the asymmetry. | No scope creep; ships fastest. | Future flakes remain possible and will be hard to attribute. | LOW | Future on-call eats the diagnosis time when concurrent runs flake. |

**Recommended**: A — It is a one-line change per CR scoped to a sibling test that already shares the same fragile assumption, and it makes brownfield-database more robust against any future suite that joins the canonical user. The PSS author already chose this pattern for their own CRs; backporting it is the natural completion under the 'apply fixes to all instances' pattern.

---

### W-003: Brittle line-number references in comments
**File**: `tests/e2e/keystone/pod-security-restricted/chainsaw-test.yaml:26` — **Fix**: ASK — **Confidence**: verified — **Pattern**: Avoid brittle line-number references in comments

**Problem**: Six places in the diff reference Go source lines by absolute number: chainsaw-test.yaml:26 and :35 (reconcile_database.go:70, reconcile_secrets.go:86), 00-namespace.yaml:16 (reconcile_secrets.go:86), 00-brownfield-db-setup.yaml:11 (reconcile_database.go:70), 01-keystone-cr.yaml:10 (reconcile_database.go:70), and docs/reference/testing/keystone-e2e-tests.md:705 (reconcile_secrets.go:86). All are correct today, but the project pattern explicitly flags absolute :70/:86-style references as drift risk: the next unrelated edit to either Go file will silently invalidate them.

**Code**:
```
#   - Brownfield mode (database.host + cache.servers, no clusterRef):
#     `Database.ClusterRef` / `Cache.ClusterRef` are namespace-local
#     (operators/keystone/internal/controller/reconcile_database.go:70).
```

**Action Required**: Replace each line-number reference with a stable anchor (function name such as mariaDBResourceKey, or the call site like reconcileSecrets's WaitForExternalSecret call) across all six locations.

**Suggested Fix**:
```
In all six locations, replace the file:line reference with a function-name or call-site anchor — e.g. 'see mariaDBResourceKey in reconcile_database.go' instead of 'reconcile_database.go:70', and 'see the WaitForExternalSecret call in reconcileSecrets' instead of 'reconcile_secrets.go:86'. Function names survive every refactor that does not rename them.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Replace each line-number reference with a stable anchor — function name (e.g. mariaDBResourceKey) or the call site (reconcileSecrets's WaitForExternalSecret call). | Survives every refactor that does not rename the function; preserves the contextual value of the reference. | Requires touching six lines across five files. | LOW | References rot on the next unrelated edit and silently mislead future readers. |
| B | Drop the references entirely and rely on git grep for readers who want to chase them. | Zero maintenance cost going forward. | Loses context that is genuinely useful for a first reader of the test. | LOW | Same as A — the references rot if left. |

**Recommended**: A — The references are valuable for first-read context; replacing line numbers with function names preserves the value without the rot. The project pattern names this exact :70/:86 shape as a finding example.

**Related**: W-001

---

### W-001: Manual rotation Job creation not idempotent under retry
**File**: `tests/e2e/keystone/pod-security-restricted/chainsaw-test.yaml:223-237` — **Fix**: AUTO-FIX — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: REQ-005 requires Job creation to be idempotent under retry. The Job name is deterministic, but kubectl create job is unconditional — there is no precheck and no delete-first guard. A re-run within the same chainsaw invocation would fail with AlreadyExists and the leading 'set -e' would abort.

**Code**:
```
set -euo pipefail
NS=keystone-pss-restricted-test
JOB=keystone-pss-restricted-manual-rotate
CRONJOB=keystone-pss-restricted-fernet-rotate

kubectl create job "$JOB" --from="cronjob/$CRONJOB" -n "$NS"
kubectl wait --for=condition=complete "job/$JOB" -n "$NS" --timeout=2m
```

**Action Required**: Prepend a kubectl delete --ignore-not-found=true --wait=true step before the create command so reruns succeed.

**Suggested Fix**:
```
set -euo pipefail
NS=keystone-pss-restricted-test
JOB=keystone-pss-restricted-manual-rotate
CRONJOB=keystone-pss-restricted-fernet-rotate

kubectl delete job "$JOB" -n "$NS" --ignore-not-found=true --wait=true
kubectl create job "$JOB" --from="cronjob/$CRONJOB" -n "$NS"
kubectl wait --for=condition=complete "job/$JOB" -n "$NS" --timeout=2m
```

---

### W-002: Task Description Mismatch: Task 1.1 says "plain Secrets" but implementation creates ExternalSecrets
**File**: `tests/e2e/keystone/pod-security-restricted/00-namespace.yaml:35-75` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Task 1.1 is marked done, but the artifact (two ExternalSecret objects, no plain Secret objects, no 32-char hex test values) does not match the description ('plain Secrets keystone-admin and keystone-db with TEST-ONLY values (32-char hex)'). Same root cause as the BLOCKING REQ-003 finding — the task description inherits the spec's incorrect assumption.

**Code**:
```
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: keystone-admin
  namespace: keystone-pss-restricted-test
...
apiVersion: external-secrets.io/v1
kind: ExternalSecret
metadata:
  name: keystone-db
```

**Action Required**: Update the task description (or the spec) to reflect that ExternalSecrets are required. Don't change the implementation.

**Suggested Fix**:
```
Rewrite Task 1.1 in the implementation tasks document to specify ExternalSecrets sourced from existing openstack-namespace ExternalSecrets (no 32-char hex test values), matching what was actually shipped.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update Task 1.1 to specify ExternalSecrets sourced from existing openstack-namespace ExternalSecrets | Task accurately describes what shipped; aligns with the operator constraint surfaced in the BLOCKING finding | Requires task list amendment | LOW | A 'done' task whose description doesn't match the artifact erodes trust in the task list |
| B | Add a deviation note alongside Task 1.1 explaining why the artifact differs | Less invasive than rewriting; preserves history of original intent | Task description body still misleads; readers must follow the deviation pointer | LOW | Same as option A but slightly diluted |

**Recommended**: A — Same reasoning as the BLOCKING REQ-003 finding: the original task description was based on an unimplementable spec clause, so the cleanest fix is to update the task to match what the operator's gating actually allows.

**Related**: B-001

---

### W-003: Task Description Mismatch: Task 1.2 says to create `keystone-pss-restricted-db-pw` Secret
**File**: `tests/e2e/keystone/pod-security-restricted/00-brownfield-db-setup.yaml` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: Task 1.2 is marked done but no keystone-pss-restricted-db-pw Secret manifest is included. The User CR's passwordSecretKeyRef points to the existing canonical keystone-db Secret in the openstack namespace. Functionally correct (the canonical Secret has the right password), but the artifact does not match the task description.

**Code**:
```
passwordSecretKeyRef:
  name: keystone-db
  key: password
```

**Action Required**: Either update Task 1.2 to record that the implementation reuses the canonical openstack/keystone-db Secret, or add the dedicated keystone-pss-restricted-db-pw Secret if isolation from the canonical user is desired.

**Suggested Fix**:
```
Decide between reusing the canonical Secret (then update Task 1.2 to document the reuse) or adding the dedicated keystone-pss-restricted-db-pw Secret (then keep Task 1.2 as written).
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Update Task 1.2 to record that the implementation reuses the canonical openstack/keystone-db Secret | Matches the existing brownfield-database test pattern; eliminates an extra Secret; reuses the OpenBao-synced password; aligns with the User CR's cleanupPolicy: Skip | Tighter coupling to canonical resources; latent parallel-run race noted in INFO finding | LOW | Task description remains inaccurate |
| B | Add a dedicated keystone-pss-restricted-db-pw Secret to fully isolate the test | Isolation from canonical user; safer under parallel: 4; matches Task 1.2 as written | Extra Secret to maintain; diverges from the existing brownfield-database pattern | LOW | Possible parallel-run flakes if canonical user teardown collides |

**Recommended**: A — The current implementation matches the established brownfield-database test pattern and uses cleanupPolicy: Skip to preserve the canonical user. Reuse is the lower-risk path and updating Task 1.2 to document it is sufficient.

**Related**: I-001

---

## INFO (2)

### I-001: Steps 1-2 lack catch blocks for diagnostic dumps on failure
**File**: `docs/reference/testing/keystone-e2e-tests.md:660-667` — **Fix**: ASK — **Confidence**: verified

**Problem**: The keystone-e2e-tests.md catch-block summary describes catch coverage for Steps 3-5. Steps 1 and 2 do not have catch blocks, so a failure there (e.g., keystone-admin ExternalSecret stuck syncing, or one of the brownfield MariaDB CRs stuck Not-Ready) would surface only as a Chainsaw timeout with no diagnostic dump. This is consistent with sibling tests, but adding a minimal catch (last 30 events, ExternalSecret status) on Steps 1-2 would be cheap insurance for future iterations.

**Code**:
```
docs/reference/testing/keystone-e2e-tests.md:660-667 — catch-block summary describes Steps 3-5 only; Steps 1 and 2 have no catch blocks in chainsaw-test.yaml.
```

**Action Required**: Add minimal catch blocks to Steps 1-2 of chainsaw-test.yaml capturing last-30 events and ExternalSecret/User/Grant status, so timeouts in those steps produce actionable diagnostics.

**Suggested Fix**:
```
Add minimal Chainsaw catch blocks to Steps 1 and 2 in chainsaw-test.yaml that dump the last 30 events in the test namespace plus the .status of ExternalSecret/User/Grant resources, mirroring the diagnostic style already used in Steps 3-5.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Add minimal catch blocks to Steps 1-2 (last-30 events plus ExternalSecret/User/Grant status). | Future timeouts in setup steps produce actionable diagnostics; cheap insurance. | Slightly more YAML to maintain; very minor. | LOW | Future failures in Steps 1-2 surface as opaque Chainsaw timeouts and waste on-call time. |
| B | Leave as-is; consistent with sibling tests. | No diff; matches existing convention across the suite. | Setup-step failures stay opaque; future debugging is slower. | LOW | Same as the problem itself — opaque failures in setup steps. |

**Recommended**: B — The reviewer explicitly flagged this as a non-blocking observation consistent with sibling tests; addressing it is best left to a follow-up that can update the suite uniformly rather than only this one test.

**Related**: W-001

---

### I-001: Positive Deviation: User CR reuses the canonical `keystone` MariaDB user
**File**: `tests/e2e/keystone/pod-security-restricted/00-brownfield-db-setup.yaml:48-59` — **Fix**: ASK — **Confidence**: verified — **Pattern**: feature-compliance

**Problem**: REQ-002 scenario 3 and Implementation Task 1.2 imply a dedicated keystone-pss-restricted-user MariaDB user. The User CR resource name is unique (preserving REQ-008 collision-avoidance), but spec.name is the shared canonical 'keystone' user — mirroring tests/e2e/keystone/brownfield-database/00-brownfield-db-setup.yaml:23-31. cleanupPolicy: Skip preserves the canonical user on teardown. Note: brownfield-database does NOT set cleanupPolicy: Skip on its User CR, so a parallel-run race remains possible if its teardown drops the canonical user mid-flight here. Pre-existing concern, not introduced by this PR.

**Code**:
```
kind: User
metadata:
  name: keystone-pss-restricted-user
spec:
  name: keystone
  passwordSecretKeyRef:
    name: keystone-db
```

**Action Required**: Track this as a sensible simplification matching the brownfield-database pattern. If parallel-run flakes appear under parallel: 4, revisit by either setting cleanupPolicy: Skip on the brownfield-database User CR or moving this test to a dedicated MariaDB user.

**Suggested Fix**:
```
No code change required for this PR. If flakes appear under parallel: 4, set cleanupPolicy: Skip on the brownfield-database test's User CR as the cheapest mitigation, or migrate pod-security-restricted to a dedicated MariaDB user for full isolation.
```

**Fix Options**:

| Option | Approach | Pros | Cons | Effort | Risk if skipped |
|--------|----------|------|------|--------|-----------------|
| A | Accept as-is; track parallel-run behavior under parallel: 4 and revisit only if flakes appear | No churn; aligns with the existing brownfield-database pattern; eliminates an extra Secret; preserves the OpenBao-synced password | Latent parallel-run race remains if brownfield-database teardown drops the canonical user | LOW | If flakes appear later, harder to attribute root cause without prior monitoring |
| B | Pre-emptively set cleanupPolicy: Skip on the brownfield-database test's User CR to eliminate the latent race | Removes the race up-front; cheap one-line change | Touches a test outside this PR's scope; minor risk of unrelated regressions | LOW | Race remains latent |

**Recommended**: A — The reviewer explicitly classifies this as a pre-existing concern not introduced by this PR. Accepting the simplification matches the existing pattern, and the issue can be addressed if and when flakes actually appear under parallel: 4.

**Related**: W-003

---

## Summary

The PR adds a thoughtfully designed Chainsaw E2E test that closes a real coverage gap (no existing test proves the Keystone reconciler's full workload admits under PSS Restricted), with brownfield mode, ESO ExternalSecrets, a manual fernet-rotation Job, and solid catch-block diagnostics for Steps 3-5. Scope is clean and the docs update touches all the right places. Three issues stand out: the comments and docs overclaim that the policy-validation Job is exercised when the CR never sets policyOverrides; cleanupPolicy: Skip is asymmetric versus the sibling brownfield-database test sharing the same canonical MariaDB user; and six absolute Go-line references will rot on any unrelated edit. Overall quality is high — the test is well-engineered and the findings are about precision rather than design. (includes adversarial review pass)

| Severity | Count |
|----------|-------|
| BLOCKING | 1 |
| CRITICAL | 1 |
| WARNING  | 6 |
| INFO     | 2 |

> [!IMPORTANT]
> **Recommendation**: Not blocking. Address Findings 1 and 2 before merge — both are accuracy/robustness issues with low-effort fixes. Finding 3 (line-number references) is mechanical and can land in this PR or as a quick follow-up. The INFO observation on Steps 1-2 catch blocks is optional and can wait.

gndrmnn (approved):

Issue #383, opened 2026-05-31
Keystone admin password rotation — operator-scheduled CronJob (Model B, Part 2 of #381)


berendt:
**Blocked on / follow-up to:** #381 (Part 1, Model A — re-bootstrap on admin Secret change). This issue is **Part 2 (Model B)** of that work and was explicitly carved out as a separate change by #381's concrete boundary 10 and its first Non-Goal. It cannot land before #381 because Model B's entire value depends on Part 1's `ESO → Secret → operator → bootstrap Job` cutover path being in place — Model B only adds the *upstream* half (operator-scheduled generation of a new password and its delivery into OpenBao).

**Description:**

Part 1 (#381) closes the *application* half of admin-credential rotation: when the `keystone-admin` Secret's `password` byte changes, the controller re-runs the idempotent `keystone-manage bootstrap` Job and Keystone begins accepting the new password. What Part 1 deliberately does **not** do is *produce* a new password — it assumes some external actor (an operator running `bao kv put kv-v2/bootstrap/keystone-admin password=<new>`, or a future scheduler) writes the new value into OpenBao, from where ESO's existing `keystone-admin` ExternalSecret (`deploy/eso/externalsecrets/keystone-admin.yaml`, `refreshInterval: 1h`, `creationPolicy: Owner`) syncs it into the cluster Secret.

Model B is that scheduler. It adds an **opt-in, feature-gated, operator-managed CronJob** that periodically generates a fresh strong admin password and delivers it into OpenBao via a `PushSecret`, so rotation happens on a schedule instead of depending on a human remembering to run `bao kv put`. The new password then reaches Keystone via *exactly* the Part 1 path — OpenBao → ESO → `keystone-admin` Secret → re-bootstrap — which is precisely why Model B can only land cleanly on top of Part 1.

This mirrors the split-compute-write rotation architecture already shipped for Fernet keys (`reconcile_fernet.go`) and credential keys (`reconcile_credential.go`) under CC-0081, with one structural difference: the production source of truth for the admin password is **OpenBao**, not an in-cluster Secret. So the apply side of the pipeline terminates in a `PushSecret` to `bootstrap/keystone-admin` rather than in an in-cluster production Secret update.

Concrete boundaries between what exists and what this issue adds:

[...]

Issue #381, opened 2026-05-31
Keystone admin credential rotation (re-bootstrap on password change)


berendt:
**Description:**

This issue closes the rotation gap on the Keystone admin credential. The pieces ESO needs to deliver a rotated password to the cluster — the OpenBao KV path `kv-v2/data/bootstrap/keystone-admin`, the `ExternalSecret` definition, the controller's Secret watch, and the idempotent `keystone-manage bootstrap` invocation — are all in place. The missing step is causing the existing bootstrap `Job` to re-run after the admin Secret's `password` data byte changes; today the Job's PodSpec is byte-identical before and after rotation because the password is injected by `valueFrom.secretKeyRef`, so the SHA-256 PodSpec hash in `internal/common/job/job.go` decides the completed Job is current and never deletes it.

Concrete boundaries between what exists and what this issue adds:

1. **ESO sync from OpenBao already runs.** `deploy/eso/externalsecrets/keystone-admin.yaml:11-29` defines an `ExternalSecret` with `refreshInterval: 1h` pulling `bootstrap/keystone-admin` (property `password`) from the `openbao-cluster-store` `ClusterSecretStore` into a Kubernetes Secret named `keystone-admin` in namespace `openstack`, `creationPolicy: Owner`. Nothing in this issue touches that file — the issue assumes ESO is the source of truth.

2. **The Secret watch already enqueues a reconcile.** `operators/keystone/internal/controller/keystone_controller.go:752-826` defines `secretToKeystoneMapper`, registered on the controller-runtime builder at line 719 via `Watches(&corev1.Secret{}, handler.EnqueueRequestsFromMapFunc(secretToKeystoneMapper(mgr.GetClient())))`. The mapper resolves a Secret event to the referencing Keystone CRs using the `KeystoneSecretNameIndexKey` field index (line 61). The indexer's extractor (`keystoneSecretNameExtractor` at line 69) already returns both `spec.database.secretRef.name` and `spec.bootstrap.adminPasswordSecretRef.name`, so a write to the `keystone-admin` Secret already triggers reconcile for the right CRs. Nothing in this issue extends the watch.

[...]

PR #385, opened 2026-06-01
feat(CC-0109): Schedule Keystone admin password rotation via operator Cr...

berendt:
# GitHub Issue #383: Keystone admin password rotation — operator-scheduled CronJob (Model B, Part 2 of #381)

**Blocked on / follow-up to:** #381 (Part 1, Model A — re-bootstrap on admin Secret change). This issue is **Part 2 (Model B)** of that work and was explicitly carved out as a separate change by #381's concrete boundary 10 and its first Non-Goal. It cannot land before #381 because Model B's entire value depends on Part 1's `ESO → Secret → operator → bootstrap Job` cutover path being in place — Model B only adds the *upstream* half (operator-scheduled generation of a new password and its delivery into OpenBao).

**Description:**

Part 1 (#381) closes the *application* half of admin-credential rotation: when the `keystone-admin` Secret's `password` byte changes, the controller re-runs the idempotent `keystone-manage bootstrap` Job and Keystone begins accepting the new password. What Part 1 deliberately does **not** do is *produce* a new password — it assumes some external actor (an operator running `bao kv put kv-v2/bootstrap/keystone-admin password=<new>`, or a future scheduler) writes the new value into OpenBao, from where ESO's existing `keystone-admin` ExternalSecret (`deploy/eso/externalsecrets/keystone-admin.yaml`, `refreshInterval: 1h`, `creationPolicy: Owner`) syncs it into the cluster Secret.

Model B is that scheduler. It adds an **opt-in, feature-gated, operator-managed CronJob** that periodically generates a fresh strong admin password and delivers it into OpenBao via a `PushSecret`, so rotation happens on a schedule instead of depending on a human remembering to run `bao kv put`. The new password then reaches Keystone via *exactly* the Part 1 path — OpenBao → ESO → `keystone-admin` Secret → re-bootstrap — which is precisely why Model B can only land cleanly on top of Part 1.

This mirrors the split-compute-write rotation architecture already shipped for Fernet keys (`reconcile_fernet.go`) and credential keys (`reconcile_credential.go`) under CC-0081, with one structural difference: the production source of truth for the admin password is **OpenBao**, not an in-cluster Secret. So the apply side of the pipeline terminates in a `PushSecret` to `bootstrap/keystone-admin` rather than in an in-cluster production Secret update.

Concrete boundaries between what exists and what this issue adds:

1. **The Fernet/credential CronJob pattern is the template, and it is fully built.** `reconcileFernetKeys` (`operators/keystone/internal/controller/reconcile_fernet.go:45-151`) and `reconcileCredentialKeys` (`reconcile_credential.go:52-159`) already implement the complete shape Model B copies: (a) an operator-owned staging Secret the CronJob PATCHes into (`ensureStagingSecret`, `rotation_staging.go:121-147`), (b) a split RBAC Role — read-only on the production Secret, `get`+`patch` on the staging Secret only (`ensureFernetRotationRBAC`, `reconcile_fernet.go:159-217`), (c) a CronJob whose script computes the new material and PATCHes it onto the staging Secret with a `forge.c5c3.io/rotation-completed-at` commit annotation (`fernetRotationCronJob`, `reconcile_fernet.go:286-421`), (d) an operator-side apply step that validates the staging output and commits it (`applyRotationOutput`, `rotation_staging.go:182-269`), and (e) a `PushSecret` backing the result up to OpenBao (`fernetKeysPushSecret`, `reconcile_fernet.go:435-461`, ensured via `secrets.EnsurePushSecret` at `reconcile_fernet.go:137`). Model B reuses all five primitives; only the leaf details (what the script generates, where the apply step writes) differ.

2. **A new feature-gate field on `BootstrapSpec` is required.** `BootstrapSpec` (`operators/keystone/api/v1alpha1/keystone_types.go:388-407`) currently holds only `AdminUser`, `AdminPasswordSecretRef`, `Region`, `PublicEndpoint`. Model B adds an optional `PasswordRotation *PasswordRotationSpec` field (`spec.bootstrap.passwordRotation`), **default disabled**, mirroring the leaf-default convention of `FernetSpec` (`keystone_types.go:332-342`), `CredentialKeysSpec` (`:344-354`), and `TrustFlushSpec` (`:356-380`). Proposed fields: `enabled bool` (default `false`), `schedule string` (cron, e.g. default `"0 0 1 * *"` monthly), and `suspend bool` (pause the CronJob without deleting it, exactly as `TrustFlushSpec.Suspend` does). An optional `passwordLength int32` with a sane minimum and default may be added if generation length needs to be tunable. Because this is a real CRD schema change (unlike Part 1, which added zero spec fields), `zz_generated.deepcopy.go`, the controller-gen CRD bases under `operators/keystone/config/crd/bases/`, and the Helm chart copy under `operators/keystone/helm/keystone-operator/crds/` **must** be regenerated and kept in lockstep (see the `check-crd-drift` skill).

3. **Defaulting and validation webhooks need to learn the new field.** Following the `TrustFlushSpec` precedent (`keystone_types.go:356-380` notes the defaulting webhook materializes the parent struct so leaf defaults fire deterministically), the defaulting webhook should materialize a default `schedule` when `passwordRotation.enabled` is true and `schedule` is empty, and the validating webhook should reject a malformed cron expression. The validating webhook is also the right place to enforce any Part-1 prerequisite invariants (e.g. that `adminPasswordSecretRef` resolves to the ESO-Owned `keystone-admin` Secret, since pushing a new password is pointless if ESO is not the source of truth for that Secret).

4. **The apply side terminates in OpenBao, not an in-cluster Secret — this is the one real divergence from the Fernet/credential template.** For Fernet/credential, `applyRotationOutput` (`rotation_staging.go:182-269`) commits the staging payload onto the *in-cluster production keys Secret*, and a separate `PushSecret` backs that Secret up to OpenBao. For the admin password, the production source of truth is OpenBao itself (ESO `creationPolicy: Owner` makes the in-cluster `keystone-admin` Secret a read-only mirror — see #381's "Migrating away from Owner" Non-Goal). Therefore the apply step must commit the new password to a **dedicated push-source Secret** (e.g. `<keystone>-admin-password-next`) that a `PushSecret` syncs to `kv-v2/data/bootstrap/keystone-admin`. The push-source Secret MUST be distinct from the ESO-Owned `keystone-admin` Secret, otherwise the `PushSecret` writer and ESO's `creationPolicy: Owner` reconciler fight over the same object every refresh. Once the new password lands in OpenBao, the existing `keystone-admin` ExternalSecret pulls it back into the `keystone-admin` Secret and Part 1 takes over.

5. **A new OpenBao write policy for the bootstrap admin path is required, and it is the most security-sensitive piece.** Today `kv-v2/data/bootstrap/*` is **read-only**: `deploy/openbao/policies/eso-management.hcl` grants only `read`, and the existing write policy `deploy/openbao/policies/push-keystone-keys.hcl` is *deliberately* scoped to the `…/keystone/+/fernet-keys` and `…/credential-keys` leaves and explicitly does **not** cover `bootstrap/*`. Model B needs a new, narrowly-scoped policy granting `create`/`update`/`read`/`delete` on **both** `kv-v2/data/bootstrap/keystone-admin` **and** `kv-v2/metadata/bootstrap/keystone-admin` (ESO's KV-v2 provider writes the metadata path too — see the rationale block in `push-keystone-keys.hcl`), bound alongside on the same auth role that runs the admin-password `PushSecret`. `eso-management.hcl` must stay read-only so the CC-0083 audit invariant ("a leaked management-cluster ESO token on `eso-management` alone cannot write to OpenBao") is preserved. Granting write to the *admin bootstrap path specifically* is a higher-trust grant than the key-backup paths and should be called out in review.

6. **A new status sub-condition and its metrics label binding are required.** Fernet/credential set `FernetKeysReady`/`CredentialKeysReady`; Model B should add a `PasswordRotationReady` condition wired into the aggregate `subConditionTypes` list and into the `subReconcilerConditionTypes` metrics-label map so the Prometheus `condition_type` label resolves instead of showing `UNKNOWN` (see the `check-condition-coverage` skill — this is a hard gate). When `passwordRotation.enabled` is false, the sub-reconciler should be a clean no-op (and tear down any previously-created CronJob/staging Secret/RBAC/PushSecret) so disabling the feature returns to the Part-1-only steady state.

7. **The generation script follows the extracted-script convention (CC-0073).** Both rotation CronJobs embed their script from `scripts/` via `//go:embed` (`reconcile_fernet.go:40` → `scripts/fernet_rotate.sh`; `reconcile_credential.go:40` → `scripts/credential_rotate.sh`) so the shell is independently lintable and testable, and materialize it through an immutable ConfigMap (`config.CreateImmutableConfigMap`, `reconcile_fernet.go:122`). Model B's generator (e.g. `scripts/admin_password_rotate.sh`) is *simpler* than the key-rotation scripts — it generates a strong random password (stdlib only, CC-0013, e.g. `python -c 'import secrets…'`) and PATCHes it onto the staging Secret with the `forge.c5c3.io/rotation-completed-at` annotation. There is **no** `keystone-manage` invocation and **no** database migration step (contrast `credential_rotate.sh`, whose `credential_migrate` step is load-bearing) — the password is opaque to Keystone until `keystone-manage bootstrap` consumes it via Part 1.

8. **The single shared `bootstrap/keystone-admin` path is a known multi-CR / parallel-e2e hazard.** Unlike the per-CR Fernet/credential paths (`openstack/keystone/{name}/…`, made per-CR by CC-0093 specifically to stop two CRs sharing one backing object), the admin password ExternalSecret is hardcoded to the single flat path `bootstrap/keystone-admin` in a fixed `openstack` namespace. Two Keystone CRs with Model B enabled would both push to the same OpenBao object and clobber each other, and parallel e2e suites would race on it (this matches the existing shared-OpenBao-KV-path flake class). This issue must make an explicit design decision: either (a) accept the single-admin / single-CR assumption and document it as a precondition for enabling Model B, or (b) make the push path per-CR (mirroring CC-0093) and reconcile that with the hardcoded ExternalSecret. Option (a) matches today's reality (one cluster admin, one hardcoded ExternalSecret) and is the recommended scope; option (b) is a larger change and likely out of scope here.

**Motivation:**

Part 1 shrinks the blast radius of an admin-credential compromise from "until cluster rebuild" to "until next ESO refresh + reconcile + bootstrap-Job lifetime" — but only *if someone rotates the password*. A rotation mechanism that still depends on a human running `bao kv put` on a remembered cadence is a process control, not a technical control: it degrades silently the moment the runbook is forgotten, and it leaves the single most privileged cloud credential static for as long as nobody acts. Model B converts that process control into an enforced schedule. With `spec.bootstrap.passwordRotation.enabled: true`, the admin password rotates automatically on the configured cadence, the rotation is observable through the same condition/metrics surface as Fernet and credential rotation, and "the admin password has not been rotated in N months" stops being a recurring audit finding (OWASP A07 — long-lived authentication credential).

Crucially this reuses machinery that already exists and is battle-tested in production: the split-compute-write CronJob, the operator-owned staging Secret, the least-privilege RBAC split, and the `PushSecret`-to-OpenBao path are all shipped for Fernet/credential keys under CC-0081/CC-0083/CC-0093. The marginal cost of Model B is one new (simpler) generation script, one feature-gated sub-reconciler that wires those primitives together, one CRD field, and one narrowly-scoped OpenBao write policy.

**Affected Areas:**

- `operators/keystone/api/v1alpha1/keystone_types.go` — add `PasswordRotationSpec` and the `PasswordRotation *PasswordRotationSpec` field on `BootstrapSpec` (`:388-407`); mirror the leaf-default convention of `FernetSpec`/`CredentialKeysSpec`/`TrustFlushSpec`.
- `operators/keystone/api/v1alpha1/zz_generated.deepcopy.go`, `operators/keystone/config/crd/bases/`, `operators/keystone/helm/keystone-operator/crds/` — regenerate from the new spec field; keep the three CRD representations in lockstep (`check-crd-drift`).
- `operators/keystone/internal/webhook/` (defaulting + validating webhooks) — default `schedule` when `passwordRotation.enabled` and `schedule` empty; reject malformed cron; enforce Part-1 prerequisite invariants.
- `operators/keystone/internal/controller/reconcile_passwordrotation.go` (new) — feature-gated sub-reconciler mirroring `reconcile_fernet.go`: ensure push-source Secret, staging Secret, split RBAC, generation CronJob, and the `bootstrap/keystone-admin` `PushSecret`; set `PasswordRotationReady`; tear everything down when disabled.
- `operators/keystone/internal/controller/scripts/admin_password_rotate.sh` (new) — stdlib-only strong-password generator that PATCHes the staging Secret with the `forge.c5c3.io/rotation-completed-at` annotation (CC-0073, CC-0013); add to the script lint/test harness.
- `operators/keystone/internal/controller/keystone_controller.go` — register the new sub-reconciler in the step chain and add `PasswordRotationReady` to `subConditionTypes`.
- `operators/keystone/internal/metrics/` — add `PasswordRotationReady` to the `subReconcilerConditionTypes` label map (`check-condition-coverage`); optionally extend the `keystone_operator_key_rotation_age_seconds` gauge with a `keyType="admin-password"` series fed from the push-source Secret's `forge.c5c3.io/rotation-completed-at` annotation.
- `deploy/openbao/policies/push-keystone-admin.hcl` (new) — narrowly-scoped `create`/`update`/`read`/`delete` on `kv-v2/data/bootstrap/keystone-admin` and `kv-v2/metadata/bootstrap/keystone-admin`; bound alongside the admin-password `PushSecret`'s role. `eso-management.hcl` stays read-only.
- `operators/keystone/internal/controller/reconcile_passwordrotation_test.go` (new) — unit coverage: resources built when enabled, absent/torn-down when disabled, `suspend` pauses the CronJob, `PasswordRotationReady` transitions, generated CronJob shape (split RBAC, staging-Secret env, PushSecret remote key).
- `operators/keystone/internal/controller/integration_test.go` — envtest coverage of enable → CronJob created → (simulated) staging PATCH → apply → push-source Secret updated.
- `tests/e2e/keystone/admin-password-scheduled-rotation/` (new Chainsaw suite; the `tests/e2e/keystone/fernet-rotation/` directory — `00-keystone-cr.yaml` + `chainsaw-test.yaml` — is the structural template) — enable rotation, drive/trigger a CronJob run, observe the new password reach OpenBao and round-trip through ESO into the `keystone-admin` Secret, observe Part 1 re-bootstrap, assert the new password authenticates (HTTP 201) and the old one fails (HTTP 401). Must account for boundary 8's single-shared-path hazard in test isolation.
- `docs/reference/keystone/keystone-reconciler.md` — document the `reconcilePasswordRotation` step, the `passwordRotation` spec, the CronJob/staging/RBAC/PushSecret topology, and the `PasswordRotationReady` condition.
- `docs/guides/keystone-key-rotation.md` (or a sibling guide) — operator-facing how-to for enabling scheduled admin-password rotation and the observable evidence (CronJob runs, OpenBao value change, ESO sync, re-bootstrap, post-rotation token check).
- `architecture/docs/09-implementation/05-keystone-dependencies.md` and `architecture/docs/05-deployment/02-secret-management.md` (in the `architecture/` submodule) — mark Model B as implemented under "Admin Credential Rotation" / Credential-Rotation mechanism 3; bump the submodule pointer.

**Acceptance Criteria:**

- [ ] `spec.bootstrap.passwordRotation` exists with `enabled` (default `false`), `schedule` (cron), and `suspend` (default `false`). The CRD schema change is regenerated across all three representations (`zz_generated.deepcopy.go`, `config/crd/bases/`, Helm `crds/`) with no drift.
- [ ] With `passwordRotation.enabled: false` (the default), **no** Model B resources exist and behaviour is byte-for-byte the Part-1 (#381) steady state. Toggling from `true` back to `false` tears down the CronJob, staging Secret, push-source Secret, RBAC, and `PushSecret`.
- [ ] With `passwordRotation.enabled: true`, the operator creates: a CronJob on the configured `schedule`, an operator-owned staging Secret (CronJob owns only `.data` via `get`+`patch`; no `create`/`delete`), a read-only RBAC grant on the push-source Secret, and a `PushSecret` targeting `bootstrap/keystone-admin`. The split-RBAC shape matches `ensureFernetRotationRBAC` (`reconcile_fernet.go:159-217`).
- [ ] The defaulting webhook materializes a default `schedule` when `enabled` is true and `schedule` is empty; the validating webhook rejects a malformed cron expression.
- [ ] `suspend: true` pauses the CronJob (`spec.suspend`) without deleting it or any sibling resource, matching `TrustFlushSpec.Suspend` semantics.
- [ ] End-to-end: enabling rotation and triggering a CronJob run results in a new password written to `kv-v2/bootstrap/keystone-admin`, synced by the existing `keystone-admin` ExternalSecret into the `keystone-admin` Secret, and applied by Part 1's re-bootstrap such that the `admin` user authenticates against `POST /v3/auth/tokens` with the new password (HTTP 201) and the old password fails (HTTP 401).
- [ ] The push-source Secret is a **distinct** object from the ESO-Owned `keystone-admin` Secret, so the `PushSecret` writer and ESO's `creationPolicy: Owner` reconciler never fight over the same object.
- [ ] A `PasswordRotationReady` status condition is set by the sub-reconciler, included in the aggregate `subConditionTypes`, and present in the `subReconcilerConditionTypes` metrics-label map (Prometheus `condition_type` resolves, never `UNKNOWN`). Verified by the `check-condition-coverage` gate.
- [ ] A new OpenBao policy grants `create`/`update`/`read`/`delete` on both `kv-v2/data/bootstrap/keystone-admin` and `kv-v2/metadata/bootstrap/keystone-admin`, scoped to that exact path. `eso-management.hcl` remains read-only; the CC-0083 audit invariant is preserved and the higher-trust grant is documented.
- [ ] The generation script is extracted under `scripts/`, embedded via `//go:embed`, materialized through an immutable ConfigMap, uses stdlib only (CC-0013/CC-0073), performs **no** `keystone-manage` call and **no** DB migration, and stamps the `forge.c5c3.io/rotation-completed-at` commit annotation atomically with its staging PATCH.
- [ ] Unit tests cover enabled/disabled/suspend, the generated CronJob/staging/RBAC/PushSecret shape, and `PasswordRotationReady` transitions. Envtest covers the enable → CronJob → staging-PATCH → apply → push-source-Secret-updated chain.
- [ ] A Chainsaw E2E suite under `tests/e2e/keystone/admin-password-scheduled-rotation/` exercises the full scheduled-rotation path and respects the single-shared-`bootstrap/keystone-admin`-path isolation constraint (boundary 8).
- [ ] The multi-CR / single-shared-path decision (boundary 8) is made explicitly and documented: either Model B is documented as requiring a single Keystone CR per `bootstrap/keystone-admin` path, or the push path is made per-CR.
- [ ] Reconciler reference and operator guide document the feature; the `architecture/` submodule chapters are updated and the submodule pointer is bumped.

**Non-Goals:**

- Part 1 itself (#381). Model B depends on it but does not re-implement the re-bootstrap-on-Secret-change path, the password-content hash, or the `PodSpecHash` change.
- Rotating service users (nova, neutron, cinder, …) or application credentials. Those use their own service users / application credentials, not the admin password, and require a different mechanism.
- Revoking already-issued Fernet tokens on rotation. A password change does not invalidate existing tokens; immediate session invalidation is a separate operation (`openstack token revoke`, or a Fernet primary-key rotation via the existing CronJob).
- Changing ESO's `refreshInterval` from 1h. Operators who need a faster cutover after a scheduled rotation can force-sync the ExternalSecret (`kubectl annotate externalsecret keystone-admin force-sync=…`); an SLA-driven interval change is a deployment-wide decision out of scope here.
- Migrating the `keystone-admin` Secret away from `creationPolicy: Owner`. The correct invariant remains "write to OpenBao first, let ESO own the in-cluster Secret"; Model B writes to OpenBao via a `PushSecret` and does not relax `Owner`.
- Adding a controller-side OpenBao client. Delivery into OpenBao is done declaratively via a `PushSecret` (the established CC-0083 pattern), not via direct operator→OpenBao API calls.
- Making the `bootstrap/keystone-admin` OpenBao path per-CR as a hard requirement. The recommended scope (boundary 8, option a) is the single-admin / single-CR assumption; a per-CR path layout is a larger follow-up if multi-CR admin rotation is ever needed.
- Changing the password generation primitive beyond "strong random from the language stdlib". No HSM, no external password vault, no complexity-policy engine.

**References:**

- Source issue / Part 1 — re-bootstrap on admin Secret change (#381)
- Fernet split-compute-write sub-reconciler — primary template (`operators/keystone/internal/controller/reconcile_fernet.go:45-461`)
- Credential split-compute-write sibling (`operators/keystone/internal/controller/reconcile_credential.go:52-465`)
- Staging-Secret + apply primitives: `ensureStagingSecret`, `applyRotationOutput`, `StagingSecretLabelKey`, `RotationCompletedAnnotation` (`operators/keystone/internal/controller/rotation_staging.go:29-269`)
- Split-RBAC pattern (read-only production, get+patch staging) — `ensureFernetRotationRBAC` (`operators/keystone/internal/controller/reconcile_fernet.go:159-217`)
- `PushSecret` to OpenBao via `secrets.EnsurePushSecret` (`operators/keystone/internal/controller/reconcile_fernet.go:135-139`, `fernetKeysPushSecret` `:435-461`)
- `BootstrapSpec` — target for the new `passwordRotation` field (`operators/keystone/api/v1alpha1/keystone_types.go:388-407`)
- Leaf-default + defaulting-webhook convention — `FernetSpec`/`CredentialKeysSpec`/`TrustFlushSpec` (`operators/keystone/api/v1alpha1/keystone_types.go:332-380`)
- `keystone-admin` ExternalSecret (1h refresh, `creationPolicy: Owner`, path `bootstrap/keystone-admin`) (`deploy/eso/externalsecrets/keystone-admin.yaml`)
- OpenBao read-only bootstrap grant — must stay read-only (`deploy/openbao/policies/eso-management.hcl`)
- OpenBao write-policy template + KV-v2 data/metadata rationale (`deploy/openbao/policies/push-keystone-keys.hcl`)
- Extracted-script + immutable-ConfigMap convention CC-0073 (`reconcile_fernet.go:40,122`, `scripts/fernet_rotate.sh`)
- Bootstrap reconciler (Part 1 consumer of the new password) (`operators/keystone/internal/controller/reconcile_bootstrap.go`)
- Chainsaw rotation template (`tests/e2e/keystone/fernet-rotation/chainsaw-test.yaml`, `00-keystone-cr.yaml`)
- Relevant feature IDs: CC-0081 (split-compute-write), CC-0083 (PushSecret write policy + audit invariant), CC-0093 (per-CR OpenBao paths), CC-0073 (extracted scripts), CC-0013 (stdlib-only). A new CC-NNNN is assigned at implementation time.
- OWASP A07 (Identification and Authentication Failures) — scheduled rotation of the longest-lived privileged credential

**Labels:** enhancement

---
Source: https://github.com/C5C3/forge/issues/383


tests/unit/deploy/test_push_keystone_admin_policy.sh
gndrmnn, line 1: The test file `tests/unit/deploy/test_push_keystone_admin_policy.sh` serves no relevant purpose and can be deleted entirely

gndrmnn (approved):

Issue #457, opened 2026-06-11
chore(repo): ignore key material and scratch CRs; stop publishing rendered tempest.conf with admin password as CI artifact


berendt:
## Problems

1. **OpenBao mTLS client bundle not gitignored.** `openbao-client.p12` (cert + private key) sits untracked at the repo root and is **not** covered by `.gitignore` — there are no `*.p12`/`*.pem`/`*.key` patterns, only a specific `keystone-ca.crt` entry. A routine `git add .` would stage and commit the private client credential for the OpenBao secret backend. (Verified via `git log --all -- openbao-client.p12` that it is not in history yet — no scrub needed if fixed now.)
2. **Cleartext admin password in a CI artifact.** `hack/ci-run-tempest.sh:117` renders `${KEYSTONE_ADMIN_PASSWORD}` into `_output/tempest/config/tempest.conf`, and `.github/workflows/ci.yaml:1119-1123` uploads the whole `_output/tempest/` tree as an artifact with `retention-days: 14`. The rendered config containing `admin_password = <cleartext>` is downloadable for 14 days by anyone with artifact access.
3. Minor: `hack/run-tempest.sh:247` substitutes the password into `sed` without escaping (the CI variant escapes at `ci-run-tempest.sh:110-114`; the local script does not) — passwords containing `|`, `&`, `\` break the substitution.
4. Minor: scratch CR manifests `controlplane.yaml`, `cp2.yml`, `demo/` are untracked and not ignored (`keystone.yaml` is).

## Proposed fix

- Add bare patterns to `.gitignore` (repo style: no leading slash): `*.p12`, `*.pem`, `*.key`, `*.pfx`, plus the scratch CR names.
- Exclude the rendered config from the artifact upload (e.g. upload only result XML, or `rm -f _output/tempest/config/tempest.conf` after the run), or inject the password via env/volume instead of baking it into a persisted file.
- Mirror the sed escaping from `ci-run-tempest.sh` into `run-tempest.sh`.

_Found by the 2026-06 pre-production audit (verified at file:line on `main` @ 0a2f19ba)._

PR #483, opened 2026-06-11
fix(common): default zap logging to production mode

### LINKED ISSUE #458: fix(common): bootstrap manager hardcodes zap Development mode for production binaries
## Problem

`internal/common/bootstrap/manager.go:118` hardcodes `opts := zap.Options{Development: true}`. The only override is the `--zap-devel` CLI flag, and neither operator Helm chart passes any zap flags (`operators/keystone/helm/keystone-operator/templates/deployment.yaml:30-36` passes only `--leader-elect`/`--namespace`; same for c5c3).

## Impact

Both production operator binaries run with DebugLevel logging, console encoder (not JSON), stacktraces on warn, and `DPanic` that actually panics.

## Proposed fix

- Default `Development: false` in the shared manager config.
- Expose log level/encoder as Helm chart values so dev setups can opt back in explicitly.

berendt:
Closes #458

## Problem

`internal/common/bootstrap/manager.go` hardcoded `zap.Options{Development: true}`, and neither operator Helm chart passed any zap flags. Both production operator binaries (keystone, c5c3) therefore ran with the controller-runtime **development** logging profile: DebugLevel verbosity, a console encoder instead of JSON, stacktraces on warnings, and a `DPanic` that actually panics.

## Change set (commit order)

1. **`fix(common): default zap logging to production mode`**
   Extract a `zapOptions()` helper (mirroring the existing `cacheOptions()` helper) returning `zap.Options{Development: false}`, so both binaries default to the production profile — JSON encoder, info level, error-level stacktraces. Runtime opt-in to development logging stays available through the `--zap-devel` / `--zap-log-level` / `--zap-encoder` flags that `BindFlags` already registers. Adds `TestZapOptions_productionDefault`.

2. **`feat(helm): expose operator log level and encoder as chart values`**
   Adds an opt-in `logging` block (`development`, `level`, `encoder`) to both the keystone-operator and c5c3-operator charts so dev setups can restore verbose/console logging explicitly:
   - `values.yaml`: the three values default to `false`/`""`.
   - `deployment.yaml`: each `--zap-*` flag is rendered only when its value is set, so the default args list (and the existing exact-match deployment test) is unchanged and production stays on the JSON/info profile.
   - `values.schema.json`: validates the block — boolean `development`, a `level` pattern (`debug|info|error|panic|<positive int>`), and a `json|console` `encoder` enum — matching the controller-runtime flag grammar so typos fail at `helm template` time.
   - helm-unittest: deployment + schema tests for the default (no zap flags), the `development` toggle, `level`/`encoder` rendering, and rejection of invalid values.
   - `Chart.yaml`: keystone-operator `0.3.0 → 0.4.0`, c5c3-operator `0.1.0 → 0.2.0`, with `artifacthub.io/changes` entries.

3. **`docs(backend): document operator logging chart values`**
   Adds a `logging` section (and test-coverage rows) to the Helm values schema reference.

## Verification

| Command | Result |
| --- | --- |
| `go test ./internal/common/...` | pass |
| `go vet ./internal/common/...` | pass |
| `make format-check` (gofumpt v0.9.2) | pass |
| `golangci-lint run ./...` (internal/common) | 0 issues |
| `go build ./...` (keystone, c5c3) | pass |
| `helm lint` (both charts) | 0 failed |
| `helm unittest` keystone-operator | 182 passed |
| `helm unittest` c5c3-operator | 139 passed |
| `make check-docs-ids` | clean |

`helm template` confirmed: default values render no `--zap-*` flags; `--set logging.development=true --set logging.level=debug --set logging.encoder=console` renders all three flags.

## Notes / out of scope

- No internal feature ID is assigned to this change (none in the issue title; following the precedent of the issue-#457 implementation, which used descriptive conventional-commit scopes rather than a CC-ID).
- The c5c3-operator chart is updated symmetrically and its helm-unittest suite passes locally, but CI's `helm-validate` job only runs helm-unittest against the keystone-operator chart — a pre-existing gap, left untouched.
- The schema reference's "Chart version" line still reads `0.2.0` (already stale against `main`'s `0.3.0`, independent of this change); left as-is to keep the change surgical.

### DIFF
--- a/docs/reference/backend/helm-values-schema.md
+++ b/docs/reference/backend/helm-values-schema.md
@@ -105,6 +105,21 @@ for the privilege-escalation path this closes. The default stays `false` because
 | --- | --- | --- | --- |
 | `metrics.port` | `integer` | minimum: `1`, maximum: `65535` | `8080` |
 
+### logging
+
+The operator runs the controller-runtime zap logger in the production profile by
+default (`development: false`): JSON encoder, info-level verbosity, and stack traces
+only at error level. Override these for human-readable console output during local
+development. Each value maps to a controller-runtime `--zap-*` flag and is omitted
+from the operator args when left at its default, so the production profile is the
+behaviour unless a field is set explicitly.
+
+| Field | Type | Constraint | Default |
+| --- | --- | --- | --- |
+| `logging.development` | `boolean` | — | `false` |
+| `logging.level` | `string` | pattern: `debug`, `info`, `error`, `panic`, or a positive integer (empty allowed) | `""` |
+| `logging.encoder` | `string` | enum: `json`, `console` (empty allowed) | `""` |
+
 ### serviceAccount
 
 | Field | Type | Constraint | Default |
@@ -166,6 +181,7 @@ Schema validation is tested with helm-unittest in
 | Invalid quantities | `resources.limits.cpu: "not-valid"` |
 | Exponent+suffix | `cpu: "1e3m"`, `memory: "1e3Ki"` |
 | Conditional constraint | `rbac.namespaceScoped=true` with `webhook.enabled=true` |
+| Logging constraints | `logging.development: "yes"`, `logging.encoder: "xml"`, `logging.level: "verbose"` |
 
 ### Positive Tests (acceptance)
 
@@ -177,3 +193,4 @@ Schema validation is tested with helm-unittest in
 | Numeric resource quantities | `cpu: 0.5` |
 | Exponent-only quantities | `cpu: "1e3"` |
 | Conditional constraint | `rbac.namespaceScoped=true` with `webhook.enabled=false` |
+| Logging overrides | `development: true`, `level: debug`, `encoder: console` |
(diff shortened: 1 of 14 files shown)


internal/common/bootstrap/manager_test.go
```diff
@@ -141,3 +141,14 @@ func TestCacheOptions_singleNamespaceEntry(t *testing.T) {
...
+// TestZapOptions_productionDefault verifies that the shared logger options
+// default to the controller-runtime production profile (Development=false):
+// JSON encoder, info-level verbosity, and stacktraces only at error level. A
+// regression to Development=true would ship debug-level console logging and a
+// DPanic that panics in both production operator binaries.
+func TestZapOptions_productionDefault(t *testing.T) {
+	if zapOptions().Development {
+		t.Fatal("expected zap Development to be false for production operator binaries")
+	}
+}
```
gndrmnn: Useless test, remove lines 144-154

gndrmnn (commented):
otherwise ok

gndrmnn (approved):

PR #485, opened 2026-06-11
fix(operators): drop webhook DELETE interception, add operator PDB/spread, uncached admission checks

### LINKED ISSUE #459: fix(operators): webhook DELETE interception can wedge namespace deletion; no PDB for 2-replica operator deployments; cached-client admission checks
## Problems

1. **DELETE verb + `failurePolicy: Fail` + no-op ValidateDelete.** Both validating webhooks register `verbs=create;update;delete` with `failurePolicy=fail` (`operators/keystone/api/v1alpha1/keystone_webhook.go:100`, `operators/c5c3/api/v1alpha1/controlplane_webhook.go:91`; rendered in both Helm charts), but both `ValidateDelete` implementations unconditionally return nil. The webhook is served in-process by the operator — if the operator is down, **no Keystone/ControlPlane CR can be deleted**: namespace deletion wedges in `Terminating`, and disaster cleanup ("operator broken, remove the CR") is blocked by the broken operator itself. The DELETE interception buys nothing today.
2. **No PodDisruptionBudget / anti-affinity for the operators.** Both charts default `replicas: 2` (`values.yaml:10`) but ship no PDB and no spread constraint — a single node drain can evict both replicas at once, taking admission down cluster-wide (compounds problem 1).
3. **One-ControlPlane-per-namespace check is cache-stale and racy.** `controlplane_webhook.go:299-320` lists via `mgr.GetClient()` (informer cache, injected at `operators/c5c3/main.go:85`). Two concurrent CREATEs both see an empty list; even sequential creates within the cache-sync window both admit. The reconciler has no defense-in-depth (unlike `CredentialRotation`, which handles `AmbiguousControlPlane`).
4. **PriorityClass existence check via cached client** (`keystone_webhook.go:789-804`): a stale cache can reject a just-created PriorityClass, and the lazy informer start happens inside the webhook timeout.

## Proposed fix

- Drop `delete` from the webhook verbs (marker + both Helm templates + `config/webhook/manifests.yaml`).
- Add PDB `minAvailable: 1` + topology spread/anti-affinity to both operator charts.
- Use `mgr.GetAPIReader()` for the one-per-namespace and PriorityClass checks; add a reconciler-side duplicate check (`Ready=False/DuplicateControlPlane` for all but the oldest CP).

berendt:
Closes #459.

Remediates the four operator-availability findings from the 2026-06 pre-production audit, in commit order:

1. **fix(operators): stop intercepting DELETE in validating webhooks** — drops the `delete` verb from both validating-webhook markers, the regenerated `config/webhook/manifests.yaml`, and both chart webhook templates. Both `ValidateDelete` implementations were unconditional no-ops, and because the webhook is served in-process with `failurePolicy: Fail`, a down operator made CR — and thereby namespace — deletion impossible. The helm-unittest suites now assert `notContains: DELETE` so a reintroduction fails the chart tests. The no-op `ValidateDelete` methods remain (required by the `admission.Validator` interface) with comments stating they are never invoked.

2. **feat(charts): add PodDisruptionBudget and node spread to operators** — both charts default `replicas: 2` but shipped no PDB and no spread constraint, so a single node drain could evict both replicas at once and take admission down cluster-wide. Adds a `policy/v1` PDB with `minAvailable: 1`, rendered only when `replicas > 1` (a single-replica install must never wedge node drains), plus a soft hostname `topologySpreadConstraint` (`ScheduleAnyway`, so single-node kind clusters stay schedulable). Chart versions bumped (keystone-operator 0.4.0, c5c3-operator 0.2.0) with `artifacthub.io/changes` entries; the FluxCD HelmReleases pin `>=0.1.0 <1.0.0`, so both bumps stay in range. No new values keys, hence no `values.schema.json` change.

3. **fix(operators): read webhook admission lookups through the API reader** — injects `mgr.GetAPIReader()` instead of `mgr.GetClient()` into both webhook structs (they already accept `client.Reader`). The one-ControlPlane-per-namespace List no longer races the informer cache (two concurrent CREATEs both seeing an empty cache), and the Keystone PriorityClass `Get` neither rejects a just-created class off a stale cache nor lazily starts an informer inside the webhook timeout. The envtest wiring mirrors the production wiring.

4. **fix(c5c3): park duplicate ControlPlanes behind the namespace's oldest** — reconciler-side defense-in-depth the issue called out as missing (unlike `CredentialRotation`'s `AmbiguousControlPlane` handling): before the sub-reconciler chain, every ControlPlane except the namespace's oldest (`creationTimestamp`, lexically smallest name breaking ties) is parked with `Ready=False` / `DuplicateControlPlane` naming the incumbent, runs no sub-reconcilers, and requeues every 30s — deleting the incumbent fires no watch event for the duplicate, so the periodic requeue is what lets it take over. The parked status write deliberately bypasses `updateStatus`, which would otherwise overwrite the reason via `setReadyCondition`; rationale is in the method comment.

### Verification

- `make test` — all unit tests pass (internal/common, keystone, c5c3)
- `go test -tags=integration` (envtest, k8s 1.35) for `operators/keystone/api/v1alpha1` and all of `operators/c5c3` — pass, including the multi-ControlPlane suites
- `golangci-lint run` in all three modules — 0 issues; `gofumpt -l` clean
- `helm unittest` both charts (178 + 135 tests) and `helm lint --strict` — pass; all five CI `helm template` scenarios render
- `make manifests` is idempotent (controller-gen v0.20.1); `make verify-crd-sync` and `make check-docs-ids` pass
- `go test -race -count=1` on the changed controller package — pass

### Deviations / notes

- No e2e (Chainsaw) test was added: the four changes are admission wiring, chart resources, and a reconciler pre-flight guard, all covered by unit, helm-unittest, and envtest layers; no existing e2e directory convention covers operator-chart resources.
- Pre-existing doc drift noticed but left untouched (out of scope): `docs/reference/c5c3/controlplane-crd.md` introduces the status-conditions section with "five sub-reconcilers" and a five-element gate diagram while `subConditionTypes` has seven entries. Where my edit touched the adjacent Ready sentence I sidestepped the wrong count ("all sub-conditions") rather than propagating it.

### DIFF
--- a/docs/reference/backend/keystone-operator-packaging.md
+++ b/docs/reference/backend/keystone-operator-packaging.md
@@ -149,7 +149,7 @@ docker run --rm keystone-operator:dev --help
 | `name` | `keystone-operator` |
 | `description` | `A Helm chart for deploying the Keystone OpenStack operator` |
 | `type` | `application` |
-| `version` | `0.1.0` |
+| `version` | `0.5.0` |
 | `appVersion` | `0.1.0` |
 
 ### Shared Library Subchart
@@ -250,6 +250,7 @@ The chart renders the following Kubernetes resources with default values:
 | ClusterRoleBinding | `rbac.authorization.k8s.io/v1/ClusterRoleBinding` | `{fullname}` | Always |
 | Deployment | `apps/v1/Deployment` | `{fullname}` | Always |
 | Service | `v1/Service` | `{fullname}` | Always |
+| PodDisruptionBudget | `policy/v1/PodDisruptionBudget` | `{fullname}` | `replicas > 1` |
 | MutatingWebhookConfiguration | `admissionregistration.k8s.io/v1` | `{fullname}-mutating` | `webhook.enabled` |
 | ValidatingWebhookConfiguration | `admissionregistration.k8s.io/v1` | `{fullname}-validating` | `webhook.enabled` |
 
@@ -262,7 +263,7 @@ All resources include standard Helm labels via the `operator-library.labels` hel
 
 | Label | Value |
 | --- | --- |
-| `helm.sh/chart` | `keystone-operator-0.1.0` |
+| `helm.sh/chart` | `keystone-operator-0.5.0` |
 | `app.kubernetes.io/name` | `keystone-operator` |
 | `app.kubernetes.io/instance` | `{release-name}` |
 | `app.kubernetes.io/version` | `0.1.0` |
@@ -323,6 +324,16 @@ configurable via Helm values.
 | `capabilities.drop` | `[ALL]` |
 | `readOnlyRootFilesystem` | `true` |
 | `seccompProfile.type` | `RuntimeDefault` |
+
+**Scheduling and disruption:**
+
+The pod template carries a best-effort topology spread constraint (`maxSkew: 1`,
+`topologyKey: kubernetes.io/hostname`, `whenUnsatisfiable: ScheduleAnyway`) so
+replicas land on different nodes where possible while single-node clusters
+(kind) stay schedulable. Together with the `PodDisruptionBudget`
+(`minAvailable: 1`, rendered only when `replicas > 1`), a voluntary disruption
+such as a node drain can never evict every replica — and with it the in-process
+admission webhook — at once.
 
 ### Service Configuration
 
--- a/docs/reference/backend/keystone-operator-packaging.md
+++ b/docs/reference/backend/keystone-operator-packaging.md
@@ -390,13 +390,17 @@ Two webhook configurations are rendered when `webhook.enabled=true`:
 | --- | --- |
 | Webhook name | `vkeystone.kb.io` |
 | Path | `/validate-keystone-openstack-c5c3-io-v1alpha1-keystone` |
-| Operations | `CREATE`, `UPDATE`, `DELETE` |
+| Operations | `CREATE`, `UPDATE` |
 | API group | `keystone.openstack.c5c3.io` |
 | API version | `v1alpha1` |
 | Resource | `keystones` |
 | Failure policy | `Fail` |
 | Side effects | `None` |
 | Admission review versions | `v1` |
+
+Neither configuration intercepts `DELETE`: the webhook is served in-process by
+the operator, so with `failurePolicy: Fail` a `DELETE` rule would let a down
+operator block CR — and thereby namespace — deletion.
 
 Both configurations include the annotation:
 
--- a/docs/reference/c5c3/controlplane-crd.md
+++ b/docs/reference/c5c3/controlplane-crd.md
@@ -692,9 +692,12 @@ admission webhooks for the `ControlPlane` CRD via the typed-generic
 interfaces from controller-runtime. `CredentialRotation` and `SecretAggregate`
 have no webhook at this level.
 
-The struct carries a `Client client.Reader`, injected at startup for any future
-cluster-scoped lookups; it is currently unused by `validate()` but kept to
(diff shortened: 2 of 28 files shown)


berendt:
# Rebase analysis: C5C3/forge#485

> Rebased onto `main`  
> Analyzed by planwerk-review 94b6755 with Claude Code



## 751f00c fix(operators): stop intercepting DELETE in validating webhooks

## fab452a feat(charts): add PodDisruptionBudget and node spread to operators

- **semantic-change** in `operators/keystone/helm/keystone-operator/Chart.yaml` — This commit's message states it bumps keystone-operator to 0.4.0 and c5c3-operator to 0.2.0, but after the rebase its Chart.yaml diff contains NO `version:` change — only artifacthub.io/changes entries. Upstream commit 2cbb413d already bumped keystone-operator 0.3.0->0.4.0 and c5c3-operator 0.1.0->0.2.0 for the unrelated log-level/encoder feature, so git's three-way merge saw the version already at the target and dropped this commit's identical bump hunk. Net effect: the new PDB template (pdb.yaml) + topologySpreadConstraints — a chart-surface-growing feature — plus the webhook-DELETE template change consolidated into this commit now ship under chart versions 0.4.0/0.2.0 that upstream already published for the log-level feature alone. Two different chart contents now exist under one version number, a semver violation against the project's bump-on-feature convention (CC-0090).
  - Action: Restore the version bumps the rebase absorbed: set operators/keystone/helm/keystone-operator/Chart.yaml version to 0.5.0 and operators/c5c3/helm/c5c3-operator/Chart.yaml version to 0.3.0. Update the packaging reference (docs/reference/backend/keystone-operator-packaging.md lines 152 and 266 currently show 0.4.0 / keystone-operator-0.4.0) and correct the commit message, which still claims the now-absent 0.4.0/0.2.0 bump.
  - Upstream: 2cbb413d75aa96b086efea69b96358de41714f15 feat(helm): expose operator log level and encoder as chart values
  - Confidence: likely

## b602ae6 fix(operators): read webhook admission lookups through the API reader

## 8db09e4 fix(c5c3): park duplicate ControlPlanes behind the namespace's oldest

## Summary

The rebase landed cleanly in functional terms: both operators build against the upstream-bumped k8s v0.36.2 / controller-runtime v0.24.1 / ESO apis digest, and the API-reader swap (GetClient->GetAPIReader), the DELETE-verb removal, and the c5c3 duplicate-parking guard all rely on stable APIs untouched by the upstream range. The one real casualty is fab452a0's chart-version bump: upstream's log-level commit (2cbb413d) independently bumped keystone-operator to 0.4.0 and c5c3-operator to 0.2.0, so the rebase silently dropped fab452a0's identical bump, leaving the new PDB/node-spread feature shipping under a chart version upstream already published for a different feature. The _deployment.tpl and deployment_test.yaml overlaps between the PDB commit and the upstream log-level commit merged correctly (both topologySpread and zap-flag content present, no duplicate test names).

> [!IMPORTANT]
> **Recommendation**: Safe to push from a build/behavior standpoint — nothing is broken at runtime and all production code compiles against the new dependencies. Before pushing, re-apply the chart-version bump the rebase absorbed in fab452a0 (keystone-operator 0.4.0->0.5.0, c5c3-operator 0.2.0->0.3.0, plus the packaging-doc and commit-message updates) so the PDB feature does not collide with upstream's log-level release under the same chart version. The other three commits need no adjustment.

---

_Rebase analysis generated by [planwerk-review](https://github.com/planwerk/planwerk-review) rebase with Claude Code_

gndrmnn (approved):

PR #495, opened 2026-06-12
feat(keystone-operator): allow aborting an in-flight upgrade by reverting spec.image.tag

### LINKED ISSUE #468: feat(keystone-operator): allow aborting an in-flight upgrade by reverting spec.image.tag
## Problem

During an active upgrade, **any** `spec.image.tag != status.targetRelease` — including a revert to `status.installedRelease` — returns a hard `UpgradeTargetChanged` error on every reconcile (`reconcile_database.go:313-331`). Nothing clears `status.upgradePhase`/`targetRelease` except contract completion (`reconcile_database.go:657-659`), and a permanently failed expand/migrate Job cannot be retried either (see the failed-Job rerun issue). The only escape from a stuck upgrade is manual status-subresource surgery.

## Proposed fix

Special-case `spec.image.tag == status.installedRelease` as an abort path:
- clear `upgradePhase`/`targetRelease`,
- delete the `db-expand`/`db-migrate`/`db-contract` Jobs,
- emit an `UpgradeAborted` event,
- `return ctrl.Result{Requeue: true}, nil`.

Document the semantics (abort is only safe before contract has run — expand/migrate are additive by design). Add unit + e2e coverage for abort-during-expand.

berendt:
Closes #468.

## Problem

During an active upgrade, **any** `spec.image.tag != status.targetRelease` — including a revert to `status.installedRelease` — returned a hard `UpgradeTargetChanged` error on every reconcile (`reconcile_database.go`). Nothing cleared `status.upgradePhase`/`targetRelease` except contract completion, so an upgrade wedged on a stuck or permanently failed expand/migrate Job could only be escaped through manual status-subresource surgery.

## Change (in commit order)

1. **`feat(keystone-operator): abort in-flight upgrade by reverting tag`**
   - In `reconcileDatabase`, special-case `spec.image.tag == status.installedRelease` while `upgradePhase != ""` as an abort path. It is checked **before** the `UpgradeTargetChanged` guard, because a revert also satisfies `tag != targetRelease`.
   - `abortUpgrade` deletes the `db-expand`/`db-migrate`/`db-contract` Jobs (via `deleteUpgradeJobs`, `DeletePropagationBackground` so owned Pods go too), clears `upgradePhase`/`targetRelease`, emits an `UpgradeAborted` event, and returns `ctrl.Result{Requeue: true}, nil` so the steady-state `db_sync` path re-derives `DatabaseReady`.
   - Jobs are deleted **before** status is cleared, so a transient delete error retains the upgrade state and the next reconcile retries the abort.
   - Unit tests: abort during expand (state cleared, phase Jobs deleted, event, requeue); abort from every active phase (also proves no `UpgradeTargetChanged` hard error on revert); delete-error retains state for retry.

2. **`test(keystone-operator): cover upgrade abort during expand (e2e)`**
   - New `tests/e2e/keystone/upgrade-abort/` Chainsaw suite. Installs at `2026.1`, then upgrades to `2026.2` — a valid sequential target whose image is never pre-loaded into kind, so the `db-expand` Pod stays in `ImagePullBackOff` and the upgrade wedges **deterministically** in `Expanding`. Reverting to `2026.1` must clear the upgrade, delete the phase Jobs, emit `UpgradeAborted`, and return to `Ready`. Unique CR + database name for the parallel runner.

3. **`docs(keystone): document aborting an in-flight upgrade`**
   - New "Aborting an Upgrade" section in the upgrade-flow reference, plus updates to the `targetRelease` note, the decision tree, the `UpgradeTargetChanged` troubleshooting resolution, and the "No automatic rollback" limitation. New suite added to the e2e-tests reference overview + directory tree.

## Semantics / safety note (documented)

Abort is only safe **before** the contract phase has run: expand and migrate are additive, so the pre-contract schema is a superset both releases run against. Contract drops the now-unused columns. The controller does **not** gate the abort on the current phase (matching the issue's proposed fix); the caveat is documented for operators.

## Local verification

- `gofmt -l` (changed Go files) — clean
- `go vet ./internal/controller/` — clean
- `go test ./internal/controller/ -count=1` — pass
- `golangci-lint run ./internal/controller/...` — 0 issues
- `make check-docs-ids` — pass
- New e2e suite YAML parses; full run is CI's job (needs a kind cluster + loaded images).

## Notes for the reviewer

- No CC-ID is assigned to #468 in `.planwerk/`, so code comments reference the GitHub issue (`#468`), mirroring the existing `(CC-NNNN, #issue)` convention; docs stay free of CC/REQ/issue IDs.
- The summary table and per-suite detail sections in `keystone-e2e-tests.md` are a curated subset (the sibling `upgrade-flow` suite is absent from both), so `upgrade-abort` was added only to the completeness enumerations (overview grid + directory tree) for parity.

### DIFF
--- a/operators/keystone/internal/controller/reconcile_database.go
+++ b/operators/keystone/internal/controller/reconcile_database.go
@@ -97,7 +97,14 @@ const (
 	conditionReasonExpandComplete        = "ExpandComplete"
 	conditionReasonMigrateComplete       = "MigrateComplete"
 	conditionReasonUpgradeComplete       = "UpgradeComplete"
+	conditionReasonUpgradeAborted        = "UpgradeAborted"
 )
+
+// upgradeJobSuffixes lists the buildDBJob nameSuffixes for the Jobs created
+// during the expand-migrate-contract upgrade flow. abortUpgrade deletes exactly
+// these (and not db-sync or schema-check, which belong to the steady-state path)
+// so reverting an in-flight upgrade leaves no stale phase Jobs behind (#468).
+var upgradeJobSuffixes = []string{"db-expand", "db-migrate", "db-contract"}
 
 // recordDBJobTerminalState observes the named DB-related Job's terminal
 // condition and emits keystone_operator_db_sync_total /
@@ -311,6 +318,14 @@ func (r *KeystoneReconciler) reconcileDatabase(ctx context.Context, keystone *ke
 
 	// Active upgrade: delegate to phase-specific handling (CC-0056).
 	if keystone.Status.UpgradePhase != "" {
+		// Abort path: reverting spec.image.tag back to the installed release
+		// cancels the in-flight upgrade and unblocks a stuck or permanently
+		// failed expand/migrate phase (#468). Checked before the
+		// target-changed guard below because a revert also satisfies
+		// tag != targetRelease and would otherwise return a hard error.
+		if keystone.Spec.Image.Tag == keystone.Status.InstalledRelease {
+			return r.abortUpgrade(ctx, keystone)
+		}
 		// Detect tag change during an active upgrade (CC-0056, REQ-009).
 		if keystone.Spec.Image.Tag != keystone.Status.TargetRelease {
 			logger.Info("Image tag changed during active upgrade",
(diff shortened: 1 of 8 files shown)


operators/keystone/internal/controller/reconcile_database.go
```diff
@@ -505,6 +520,73 @@ func (r *KeystoneReconciler) initiateUpgrade(ctx context.Context, keystone *keys
...
+// permanently failed expand/migrate Job (#468).
+//
+// Aborting is only safe before the contract phase has run. Expand and migrate
+// are additive by design: they create the new columns/tables and backfill data
+// without dropping anything the installed release still reads, so the
+// pre-contract schema is a superset both releases run against. Contract drops
+// the now-unused columns; once it has started, the installed release would be
+// pointed at a schema missing fields it expects, so a post-contract revert is
+// not a safe rollback. The reverted release is the operator's responsibility to
+// validate \u2014 the controller does not gate the abort on the current phase.
```
gndrmnn:
```suggestion
// not a safe rollback. The reverted release is the operator's responsibility to
// validate the controller does not gate the abort on the current phase.
```

gndrmnn (approved):

PR #499, opened 2026-06-13
fix(c5c3-operator): mode-flip/rename orphans, unreachable missing-CRD handling, missing ESO watches, misc gaps

### LINKED ISSUE #476: fix(c5c3-operator): mode-flip/rename orphans, unreachable missing-CRD handling, missing ESO watches, misc gaps
## Problems

1. **Mode-flip and rename orphans.** Flipping to brownfield or renaming `clusterRef.Name` leaves the old MariaDB/Memcached running and owned (`reconcile_infrastructure.go:89-126`) — an orphaned 3-replica Galera keeps consuming 100Gi×3 until CP deletion. Brownfield early-exit never deletes a previously projected ExternalSecret (`reconcile_dbcredentials.go:88-98`); renaming `cloudCredentialsRef.secretName` leaks the old ES (`reconcile_korc.go:1205-1213`). → each sub-reconciler deletes the child it would have projected under the previous mode/name, or `ValidateUpdate` restricts the change.
2. **Missing-CRD safety is mostly unreachable.** `Owns()` on all ORC kinds + unstructured Memcached (`controlplane_controller.go:277-291`) makes manager start fail if any CRD is absent — the elaborate in-reconcile `KORCCRDNotInstalled` branches can only fire if a CRD is deleted *after* start; `ensureMemcached` has no no-match handling at all (`reconcile_infrastructure.go:234-250`). → probe the RESTMapper at startup and conditionally register watches (keystone does this for HTTPRoute, `keystone_controller.go:699-730`), or document the hard dependency and drop the dead branches.
3. **No ESO watches.** No `Owns`/`Watches` on ExternalSecret/PushSecret/ClusterSecretStore — `AdminCredentialReady`/`DBCredentialsReady` stay stale-True between 10-min resyncs during an OpenBao/ESO outage. Keystone watches these for exactly this reason (`keystone_controller.go:752,765-767`). → add the watches + the `IsClusterSecretStoreReady` gate.
4. Misc: invalid rotation interval returns `ctrl.Result{}, nil` without requeue while the chain continues past the failed sub-reconciler (`reconcile_keystone.go:91-104` + `controlplane_controller.go:117-121`); `status.services`/`updatePhase` declared but never written (`controlplane_types.go:348-356`); `reconcileCatalog` misses the empty-`SecretName` fallback used everywhere else (`reconcile_korc.go:1263`); `gateway.hostname` unvalidated → `https:///v3` endpoint; `keystone.Spec.Database` shallow-copies pointers from `cp.Spec` (`reconcile_keystone.go:113`) — use `DeepCopy()` like the Gateway projection.

berendt:
Closes #476

Pre-production hardening pass over the c5c3 (ControlPlane) operator, addressing the four defect classes from the 2026-06 audit. Each problem is its own small, reviewable commit; the destructive Problem-2 deletion is isolated so it can be dropped if the team prefers the RESTMapper-probe alternative.

## Problem 1 — mode-flip / rename orphans

Flipping the database/cache mode (managed `clusterRef` ↔ brownfield `host`/`servers`), renaming a managed `clusterRef.Name`, or renaming `cloudCredentialsRef.secretName` left the previously-projected MariaDB/Memcached child (and its credentials) or the K-ORC clouds.yaml ExternalSecret orphaned and owned until CP deletion.

- **`fix(c5c3): make infra mode, clusterRef and cloud secretName immutable`** — `ValidateUpdate` now rejects changes to those create-only fields. The webhook (not CEL) is load-bearing because the affected leaves live in the shared `commonv1.DatabaseSpec`/`CacheSpec` reused by the keystone operator. `validate()` was refactored to return a `field.ErrorList` so spec + immutability errors fold into one `Invalid` response.

## Problem 2 — unreachable missing-CRD handling

`SetupWithManager` `Owns`/`Watches` every K-ORC, Memcached and ESO kind, so the manager fails fast at startup if any CRD is absent — the in-reconcile `KORCCRDNotInstalled`/`catalogCRDMissing` branches could only fire if a CRD were deleted *after* start.

- **`refactor(c5c3): drop unreachable missing-CRD reconcile branches`** — treat K-ORC (like Memcached/ESO/MariaDB) as a hard dependency and remove the dead branches; no-match errors fall through to the generic error returns. Drops the `catalogCRDMissing` helper and the orphaned `meta`/`logr` imports. **Isolated commit** — drop it to substitute the RESTMapper-probe approach instead.

## Problem 3 — missing ESO watches

`AdminCredentialReady`/`DBCredentialsReady`/`AdminPasswordReady` could stay stale-True between resyncs during an OpenBao/ESO outage.

- **`feat(c5c3): watch ESO objects and gate on ClusterSecretStore`** — `Owns(ExternalSecret)`/`Owns(PushSecret)` + a `ClusterSecretStore` watch (cluster-wide mapper), plus an `IsClusterSecretStoreReady` gate in the three credential sub-reconcilers, mirroring the keystone operator's `reconcile_secrets.go`. RBAC was already granted.

## Problem 4 — misc

- **`fix(c5c3): validate gateway.hostname on the ControlPlane webhook`** — reject a gateway without a hostname (would derive an empty `https:///v3` endpoint).
- **`fix(c5c3): deep-copy projected Keystone database and cache specs`** — `DeepCopy()` the projected `Database`/`Cache` so the child never aliases `cp.Spec`, matching the Gateway projection.
- **`fix(c5c3): error on invalid Keystone rotation interval`** — return the error so the chain stops and requeues instead of `ctrl.Result{}, nil` letting it continue.
- **`fix(c5c3): apply clouds.yaml secretName fallback in reconcileCatalog`** — empty `SecretName` falls back to `korcCloudsYamlSecretName` like the other call sites.
- **`feat(c5c3): populate status.services and status.updatePhase`** — both fields were declared but never written; `updatePhase` is fixed at `Idle`, `services["keystone"]` mirrors `KeystoneReady` + release.

- **`docs(c5c3): refresh ControlPlane reference for #476 changes`** — Watches table, `SecretStoreNotReady` rows, immutability/gateway contract, removed `KORCCRDNotInstalled` rows + hard-dependency note, status-field wiring (reconciler + CRD + infrastructure-manifests docs).

## Tests & verification

- New unit tests: webhook immutability + gateway, catalog secretName fallback, the three store-not-ready gates, the new ClusterSecretStore mapper; updated the rotation-interval and missing-CRD tests to the new behavior.
- Integration tests seed a Ready `ClusterSecretStore` before driving the chain; the full envtest suite passes.
- `go build` / `go vet` / `golangci-lint` (0 issues) / `gofumpt` clean / `make manifests` (no CRD drift) / `make check-docs-ids` clean.

## Notes for the reviewer

- **Problem 1 fork:** implemented as webhook immutability (smallest, single-file). The alternative the issue also sanctions — delete-orphans in each sub-reconciler — would keep post-create mode/name edits allowed at materially more cost.
- **Problem 2 fork:** implemented as drop-dead-branches (commit isolated). The RESTMapper-probe + conditional `Owns()` alternative preserves graceful degradation but yields a split dependency model.
- **DeepCopy (Problem 4):** behavior-preserving and untestable through the fake client (it deep-copies on store), so it relies on the existing projection-value test rather than a vacuous aliasing assertion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,16 @@ CONTROLLER_GEN ?= controller-gen
 # Default resolves via GOPATH so local runs work without manually exporting GOBIN.
 SETUP_ENVTEST ?= $(shell go env GOPATH)/bin/setup-envtest
 
+# Local tool binaries are installed under bin/ (gitignored) so local dev pins
+# the same versions as CI without depending on whatever sits on $PATH.
+LOCALBIN ?= $(CURDIR)/bin
+
 # Pin gofumpt version to match CI (single source of truth for local dev).
 # Must be kept in sync with GOFUMPT_VERSION in .github/workflows/ci.yaml.
+# GOFUMPT resolves to the bin/ copy managed by install-gofumpt at the pinned
+# version, so a stale $PATH gofumpt can never silently format differently.
 GOFUMPT_VERSION ?= v0.10.0
-GOFUMPT ?= gofumpt
+GOFUMPT ?= $(LOCALBIN)/gofumpt
 
 # Kubernetes version for envtest binary downloads.
 # Pin to a specific version for reproducible integration tests across runs.
@@ -186,14 +192,14 @@ test-shell:
 .PHONY: fmt
 # fmt formats all tracked Go files with gofumpt.
 # Only formats git-tracked files to skip generated, vendored, or tooling code.
-fmt:
+fmt: install-gofumpt
 	@echo "Formatting Go files with gofumpt..."
 	@git ls-files '*.go' | xargs $(GOFUMPT) -w
 
 .PHONY: format-check
 # format-check verifies all tracked Go files conform to gofumpt formatting.
 # Mirrors the CI format-check job for local pre-commit validation.
-format-check:
+format-check: install-gofumpt
 	@unformatted=$$(git ls-files '*.go' | xargs $(GOFUMPT) -l); \
 	if [ -n "$$unformatted" ]; then \
 		echo "The following files are not formatted with gofumpt:"; \
@@ -204,10 +210,14 @@ format-check:
 	echo "Format check passed."
 
 .PHONY: install-gofumpt
-# install-gofumpt installs gofumpt at the pinned version.
-# Ensures local development uses the same version as CI.
+# install-gofumpt installs gofumpt at the pinned version into bin/.
+# Re-installs when the binary is missing or its version drifts from the pin so
+# fmt/format-check always run the exact version the CI format-check job uses.
 install-gofumpt:
-	go install mvdan.cc/gofumpt@$(GOFUMPT_VERSION)
+	@if ! test -x '$(GOFUMPT)' || ! '$(GOFUMPT)' --version | grep -qF '$(GOFUMPT_VERSION)'; then \
+		echo "Installing gofumpt $(GOFUMPT_VERSION) into $(LOCALBIN)..."; \
+		GOBIN='$(LOCALBIN)' go install mvdan.cc/gofumpt@$(GOFUMPT_VERSION); \
+	fi
 
 # ============================================================================
 # Code Generation Targets
--- a/docs/reference/c5c3/controlplane-crd.md
+++ b/docs/reference/c5c3/controlplane-crd.md
@@ -299,8 +299,8 @@ Keystone service.
 | `replicas` | `*int32` | No | `nil` (Keystone operator default, 3) | Overrides the number of Keystone API replicas. When `nil`, the reconciler leaves `replicas` unset on the projected Keystone CR, so the Keystone operator applies its own default. Minimum: 1. |
 | `image` | [`*commonv1.ImageSpec`](../keystone/keystone-crd.md#imagespec) | No | `nil` | Overrides the Keystone container image. When `nil`, the reconciler derives the image as `ghcr.io/c5c3/keystone:{spec.openStackRelease}`. When set, the whole image reference is used verbatim. |
 | `policyOverrides` | [`*commonv1.PolicySpec`](../keystone/keystone-crd.md#policyspec) | No | `nil` | Per-service oslo.policy overrides for Keystone. When set, these take precedence over `spec.global` for the Keystone service. |
-| `rotationInterval` | `*metav1.Duration` | No | `nil` | Overrides the Fernet / credential-key rotation interval the reconciler derives for the projected Keystone CR. When `nil`, the reconciler derives a default schedule. When set, the duration is converted to a cron expression and applied to both `fernet.rotationSchedule` and `credentialKeys.rotationSchedule` on the projected Keystone CR; an unconvertible interval surfaces `KeystoneReady=False` with reason `InvalidRotationInterval`. |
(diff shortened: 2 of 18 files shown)


gndrmnn (approved):

berendt:
# Rebase analysis: C5C3/forge#499

> Rebased onto `main`  
> Analyzed by [planwerk-review](https://github.com/planwerk/planwerk-review) 3c413ef with Claude



## 9526ae8 fix(c5c3): make infra mode, clusterRef and cloud secretName immutable

## ae036b0 fix(c5c3): validate gateway.hostname on the ControlPlane webhook

- **lint-rule** in `operators/c5c3/api/v1alpha1/controlplane_webhook.go` — This commit was authored before the base scrubbed internal feature/requirement IDs. While it forked, upstream added an always-on, non-path-filtered `feature-ids` CI job (.github/workflows/ci.yaml:250, `if: github.event_name == 'pull_request'`) that runs `make check-feature-ids` -> scripts/check-no-feature-ids.sh, which fails on any `CC-NNNN`/`REQ-NNN` token in git-tracked files (case-insensitive, non-letter boundary). The gateway-hostname check this commit adds carries a `// REQ-006:` comment at controlplane_webhook.go:297 — the only remaining feature ID in the c5c3 tree. The chore commits (874b4a27/f1104fee/c20ab7f6) cleaned the base, so this line is a fresh violation that will fail the gate on PR #499.
  - Action: Drop the `REQ-006:` token from the comment (e.g. start the comment at `// when a gateway is configured, its hostname must be set. Mirrors the ...`). The explanatory text and the kubebuilder MinLength reference can stay; only the internal requirement ID must go.
  - Upstream: f80fadc7f03c8e6a08c26596bc0c3f867e47b26c ci: add repo-wide feature-ID gate, fold in docs-only check
  - Confidence: verified

## b0c9a59 fix(c5c3): deep-copy projected Keystone database and cache specs

## 527110a fix(c5c3): error on invalid Keystone rotation interval

## e7ec8d7 fix(c5c3): apply clouds.yaml secretName fallback in reconcileCatalog

## 4b50e66 feat(c5c3): populate status.services and status.updatePhase

## 712301a feat(c5c3): watch ESO objects and gate on ClusterSecretStore

## e9e51bd refactor(c5c3): drop unreachable missing-CRD reconcile branches

## c4a24ac docs(c5c3): refresh ControlPlane reference for #476 changes

## Summary

The rebase landed cleanly: `go build ./...` for the c5c3 module passes, and every upstream interaction was verified compatible. The webhook commits (9526ae80, ae036b01) sit correctly on top of the upstream DELETE-verb drop (7f80ad2f) and the GetAPIReader wiring (249258c4), both of which survive in the current tree; the ESO watch commit (712301a3) uses the exact import paths the keystone operator uses at the bumped ESO digest (05dd6620), and its helpers/constants all resolve; the status and refactor commits (4b50e665, e9e51bd1) integrate consistently with the upstream duplicate-parking guard (de795115); and the docs commit (c4a24ac6) did not clobber the upstream-added DuplicateControlPlane/uncached-reader/one-per-namespace content. The single concrete defect is that ae036b01 reintroduces a `REQ-006` internal requirement ID in controlplane_webhook.go:297, which the new always-on `feature-ids` CI gate (f80fadc7) — absent when the PR forked — now rejects.

> [!IMPORTANT]
> **Recommendation**: Not safe to push as-is: the `feature-ids` CI job will fail on the `REQ-006` token in ae036b01. Scrub that single ID from the comment (a one-line edit, ideally amended into ae036b01) and the branch is otherwise clean and ready — no other adjustments are required.

---

_Rebase analysis generated by [planwerk-review](https://github.com/planwerk/planwerk-review) 3c413ef rebase with Claude_

PR #500, opened 2026-06-13
fix(keystone-operator): rotation staging/commit races, ESO finalizer assumptions, DB-TLS Certificate orphan on disable

### LINKED ISSUE #475: fix(keystone-operator): rotation staging/commit races, ESO finalizer assumptions, DB-TLS Certificate orphan on disable
## Problems

1. **Staging commit can lose or mix rotation runs.** (a) `rotation_staging.go:258` deletes the staging Secret after the production update with no preconditions — a CronJob PATCHing new rotation output between the Get (`:192`) and the Delete loses that output uncommitted. → use `client.Preconditions{UID, ResourceVersion}` and tolerate the conflict. (b) The rotation scripts PATCH with strategic-merge semantics (`scripts/fernet_rotate.sh:20-27`): after a `RotationRejected` (staging deliberately retained, `rotation_staging.go:228-232`), the next run merges over the stale payload — leftover key indices can survive and `validateRotationOutput` may accept a key set keystone-manage never produced together. → operator clears `staging.Data` on rejection.
2. **Deletion can deadlock on a third-party finalizer contract.** `reconcile_secrets.go:40-55` defines two contradictory "ESO cleanup finalizer" constants (`pushsecret.externalsecrets.io/finalizer` vs `external-secrets.io/cleanup`); pass-0 of `finalizeOpenBaoSecrets` (`:215-239`) blocks CR deletion on observing the former. Only one chainsaw test validates the literal against real ESO — an ESO upgrade that renames its finalizer turns into "CR deletion hangs at WaitingForESOAdoption forever". → fix the comments, assert the literal against the pinned ESO version in e2e, add a bounded adoption-wait.
3. **Disabling `spec.database.tls` orphans the managed Certificate.** `reconcile_databasetls.go:62-90` only sets the condition; the `<name>-db-client` Certificate keeps being renewed by cert-manager until CR deletion (HPA/NetworkPolicy/HTTPRoute all delete on disable). Also no `Owns(Certificate)` watch — `DatabaseTLSReady` never reflects later issuance failures. → delete on the non-managed paths; consider a CRD-gated watch.
4. Minor: `buildUpgradeJob` doc comment (`reconcile_database.go:850-853`) contradicts actual behavior (all three phases use the new release); misleads on-call during a stuck upgrade. `applyAdminPasswordRotation` stamps `rotation-completed-at` when committing to the push source (`reconcile_passwordrotation.go:532-539`) although the password goes live only after the full ESO round-trip + bootstrap re-run (~1h+) — document or stamp on bootstrap completion.

berendt:
## Summary

Four independent pre-production hardening fixes in the keystone-operator,
each a small, surgical change at the file:line the audit cited. No API-type,
CRD, or RBAC changes were required (the cert-manager `certificates` verbs are
already granted in the marker and Helm `_helpers.tpl`).

Closes #475

## Changes (in commit order)

1. **`fix(keystone-operator): guard rotation staging delete with preconditions`**
   — Problem 1a. The split-compute-write commit deleted the staging Secret with
   an unconditional `Delete`; a rotation CronJob PATCHing fresh output between
   the operator's read and the delete had that output removed uncommitted. Both
   `applyRotationOutput` (fernet/credential) and `applyAdminPasswordRotation`
   now carry `client.Preconditions{UID, ResourceVersion}` on the delete and
   tolerate the resulting 409 Conflict — the newer payload survives and commits
   next reconcile.

2. **`fix(keystone-operator): clear rejected rotation staging payload`**
   — Problem 1b. The rotation scripts PATCH with strategic-merge semantics over
   a multi-key `.data` map, so after a `RotationRejected` the next run merged
   over the stale payload and `validateRotationOutput` could accept a key set
   keystone-manage never produced together. The operator now clears
   `staging.Data` and the completion annotation on rejection (the Secret object
   is kept; the Warning event is the diagnostic). The single-key admin-password
   path is intentionally left as-is.

3. **`fix(keystone-operator): bound the OpenBao finalizer adoption wait`**
   — Problem 2 (mechanism). Pass-0 of `finalizeOpenBaoSecrets` blocked CR
   deletion forever if a backup PushSecret never gained ESO's adoption
   finalizer (e.g. an ESO finalizer rename, or ESO down). A new
   `OpenBaoAdoptionWaitTimeout` (10m) bounds the wait: past it the handler emits
   an `ESOAdoptionTimedOut` Warning and force-deletes via Pass-1. Safe in the
   rename case (ESO still purges via its renamed finalizer); orphans the kv-v2
   path only if ESO is genuinely down — an explicit, event-surfaced trade-off
   over hanging.

4. **`fix(keystone-operator): correct ESO finalizer comments, assert in e2e`**
   — Problem 2 (contract). The two finalizer constants both called themselves
   "the cleanup finalizer". Their comments now distinguish the
   *adoption* finalizer (`pushsecret.externalsecrets.io/finalizer`, the only one
   production branches on) from the test-only *remote-purge* finalizer. The
   `deletion-cleanup` chainsaw suite now asserts the adoption literal against
   the deployed ESO version before deletion, so an upstream rename fails CI
   loudly instead of hanging.

5. **`fix(keystone-operator): delete managed DB-client cert on TLS disable`**
   — Problem 3. Disabling `spec.database.tls` (or switching to brownfield) left
   the managed `<name>-db-client` Certificate being renewed by cert-manager
   until CR deletion. `reconcileDatabaseTLS` now deletes it on both non-managed
   paths (cert-manager GCs the issued Secret via the owner-ref cascade), and a
   cert-manager-CRD-gated `Owns(Certificate)` watch reflects later issuance
   failures in `DatabaseTLSReady`. Both are gated on a `RESTMapper` probe
   mirroring the Gateway-API/HTTPRoute pattern, so the default no-TLS config
   never errors with "no matches for kind Certificate".

6. **`docs(keystone-operator): fix upgrade-job comment and rotation-age note`**
   — Problem 4. `buildUpgradeJob`'s comment claimed expand/migrate use the old
   release; all three phases actually run the new image — corrected so it no
   longer misleads on-call. Documented that `rotation-completed-at` for
   `key_type="admin-password"` marks commit-to-push-source (not time-since-live,
   which lags by the full ESO round-trip + bootstrap re-run, ~1h+), in the
   reconciler comments and the metrics reference.

## Tests

- New unit tests for: precondition-guarded staging delete tolerating Conflict
  (fernet + admin-password); rejected-staging clearing; bounded adoption wait
  (proceeds after timeout / blocks within window); DB-client Certificate delete
  on disable + brownfield + cert-manager-absent skip; `isCertManagerAvailable`.
- Existing rejection tests strengthened to assert the cleared staging payload.
- New e2e assertion pins the ESO adoption-finalizer literal in
  `tests/e2e/keystone/deletion-cleanup/`.

## Local verification

- `make test-operator OPERATOR=keystone` — pass (controller 88.3%)
- `go test -race ./operators/keystone/internal/controller/` — pass
- `golangci-lint run ./...` (keystone module) — 0 issues; `gofmt -l` clean
- `make check-docs-ids` — pass
- `go vet -tags=integration` — compiles. envtest/kind suites (integration, e2e)
  deferred to CI; the bounded-wait change only alters behavior after 10m, which
  no integration test triggers.

## Deviations from the issue

- **Admin-password sibling (1a):** the issue cites only `rotation_staging.go`,
  but `applyAdminPasswordRotation` has the identical delete race, so the
  precondition fix is applied there too (project "apply bug fixes to all
  instances" pattern).
- **Problem 4b** takes the lighter "document" option (proportionate to a
  "Minor" item); stamping `rotation-completed-at` on bootstrap completion would
  require threading the timestamp through the bootstrap re-run path and is left
  for a separate issue.
- Folded the planned dedicated `RejectionClearsStagingData` test into the two
  existing rejection tests (which now assert clearing) to avoid a near-duplicate
  test.

### DIFF
--- a/docs/guides/enable-keystone-database-tls.md
+++ b/docs/guides/enable-keystone-database-tls.md
@@ -255,10 +255,11 @@ kubectl -n openstack patch keystone keystone --type json --patch '[
 ```
 
 The `DatabaseTLSReady` condition transitions to `True` with `reason=NotRequired`,
-the `<name>-db-client` Secret is garbage-collected via the cert-manager
-`Certificate` owner reference cascade, and subsequent connections fall back to
-plaintext TCP — but only succeed if MariaDB's `spec.tls.required` is also
-turned off (see warning above).
+the operator deletes the managed `<name>-db-client` `Certificate` (so
+cert-manager stops renewing it), cert-manager then garbage-collects the issued
+Secret via the `Certificate` owner-reference cascade, and subsequent
+connections fall back to plaintext TCP — but only succeed if MariaDB's
+`spec.tls.required` is also turned off (see warning above).
 
 ---
 
--- a/docs/guides/keystone-key-rotation.md
+++ b/docs/guides/keystone-key-rotation.md
@@ -167,8 +167,12 @@ subsequent rotations.
 ## 5. Recover from `RotationRejected`
 
 The operator validates every staged rotation before copying it onto the
-production Secret. On failure it emits a Warning event and **keeps the
-staging Secret in place** so you can inspect what the CronJob produced.
+production Secret. On failure it emits a Warning event and **clears the
+staged payload** — the staging Secret object is kept, but its `.data` and
+the `forge.c5c3.io/rotation-completed-at` annotation are removed so the
+next CronJob run starts from an empty base rather than merging over a
+rejected payload. The Warning event message, not the staging Secret
+contents, is the record of what was rejected.
 
 ### Symptoms
 
@@ -189,13 +193,11 @@ The companion Warning reason `RotationAnnotationInvalid` indicates the
 annotation was present but did not parse as RFC3339; the remediation path
 is the same.
 
-### Inspect the staging Secret
+### Match the rejection reason
 
-```bash
-kubectl -n <ns> get secret <ks>-fernet-keys-rotation -o yaml
-```
-
-Match the event message against the operator's validation contract
+Because the staged `.data` is cleared on rejection, the `RotationRejected`
+event message above — not the staging Secret — is what tells you which
+rule failed. Match it against the operator's validation contract
 (see [Operator validation rules](../reference/keystone/keystone-reconciler.md#key-rotation-rbac-split)):
 
 | Event message contains | Likely cause |
@@ -212,16 +214,18 @@ The recovery sequence is always:
 
 1. Fix the cause (CronJob image, script, or CR spec) so the next rotation
    will produce valid output.
-2. Force a fresh rotation:
+2. Force a fresh rotation. The operator already cleared the staged data on
+   rejection, so deleting the staging Secret by hand is optional
+   belt-and-suspenders:
 
    ```bash
-   kubectl -n <ns> delete secret <ks>-fernet-keys-rotation
+   kubectl -n <ns> delete secret <ks>-fernet-keys-rotation   # optional
    kubectl -n <ns> create job --from=cronjob/<ks>-fernet-rotate \
      <ks>-fernet-rotate-recover-$(date +%s)
    ```
 
-   The operator re-creates an empty staging Secret on the next reconcile;
-   the new Job PATCHes it; the operator validates and applies.
+   The new Job PATCHes the (already-empty) staging Secret; the operator
+   validates and applies.
 3. Confirm apply by repeating steps 2-4 above.
 
 > **Production safety note.** The production Secret is never modified
--- a/docs/reference/keystone-operator-metrics.md
+++ b/docs/reference/keystone-operator-metrics.md
@@ -27,7 +27,7 @@ for cluster-side wiring.
 | --- | --- | --- | --- |
 | `keystone_operator_reconcile_duration_seconds` | Histogram | `sub_reconciler` | Latency of each sub-reconciler invocation |
 | `keystone_operator_reconcile_errors_total` | Counter | `sub_reconciler`, `condition_type` | Count of sub-reconciler failures per Ready sub-condition |
-| `keystone_operator_key_rotation_age_seconds` | Gauge | `keystone`, `namespace`, `key_type` | Age of the most recent successful Fernet/credential rotation |
+| `keystone_operator_key_rotation_age_seconds` | Gauge | `keystone`, `namespace`, `key_type` | Age of the most recent successful Fernet/credential/admin-password rotation |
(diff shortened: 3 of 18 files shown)


gndrmnn (approved):

berendt:
# Rebase analysis: C5C3/forge#500

> Rebased onto `main`  
> Analyzed by planwerk-review e1efd0d with Claude



## 7bd111b fix(keystone-operator): guard rotation staging delete with preconditions

## c76e199 fix(keystone-operator): clear rejected rotation staging payload

## f806219 fix(keystone-operator): bound the OpenBao finalizer adoption wait

## de70d63 fix(keystone-operator): correct ESO finalizer comments, assert in e2e

## 2a05a86 fix(keystone-operator): delete managed DB-client cert on TLS disable

## fa428ef docs(keystone-operator): fix upgrade-job comment and rotation-age note

## Summary

The rebase landed cleanly: all six commits sit on top of the 27 upstream commits with no surviving semantic breakage. I verified every cross-file dependency the commits rely on still exists with the assumed shape — the watch-mapper refactor (9c9f3ad1) kept SetupWithManager in keystone_controller.go where commit 5 edits it; the table-driven chain (84cde019) and rotation-requeue (7e9e4df4) refactors touch different files than commits 1-2; the upgrade-abort commit (125509f7) leaves commit 6's buildUpgradeJob comment accurate (all three phases still pass Spec.Image.Tag, abort deletes jobs rather than rebuilding with the old tag); and the external-secrets/gomega bumps are go.mod-only with unchanged module hashes. The one item I could not execute in this sandbox (go/gofumpt are blocked) is the dev-tooling bump in 97322039, which moved the CI pins to gofumpt v0.10.0 / golangci-lint v2.12.2 / controller-gen v0.21.0; that upstream commit reformatted only two spots monorepo-wide and neither matches the patterns in the new code (which mirror untouched adjacent Eventf/logger.Info/RESTMapper code), and no rebased commit touches go.mod/CRDs/_types.go/webhooks, so controller-gen regen and the webhook commits have zero surface area.

> [!IMPORTANT]
> **Recommendation**: Safe to push as-is — no grounded code adjustments are required. As the only pre-push verification, run `make format-check` and the linter locally with the newly pinned gofumpt v0.10.0 / golangci-lint v2.12.2 (this could not be executed here because the Go toolchain is unavailable in the sandbox); the new code mirrors adjacent patterns the tooling bump left untouched, so a clean run is expected.

---

_Rebase analysis generated by [planwerk-review](https://github.com/planwerk/planwerk-review) rebase with Claude_

PR #503, opened 2026-06-14
refactor(common): shared instrumentation/metrics, policy reuse + validation gap, dead-code sweep

### LINKED ISSUE #479: refactor(common): shared instrumentation/metrics, policy merge/validate reuse (closes a validation gap), dead-code sweep
## Problems

1. **Byte-identical instrumentation duplicated across operators.** `operators/{keystone,c5c3}/internal/controller/instrumentation.go` contain identical `instrumentSubReconciler`, test-rebind vars and UNKNOWN fallback — only the condition-type map differs. Same for the metrics collector skeletons (`internal/metrics/collectors.go` both sides). → `internal/common/instrumentation` taking `(metricPrefix, conditionTypes)`; keystone keeps its extra collectors locally.
2. **Policy logic re-implemented per operator while the common version rots unused** — with a real validation gap: `policy.MergePolicies`/`policy.ValidatePolicyRules` have zero importers; c5c3 re-implements merge semantics (`helpers.go:62-114`); the keystone webhook hand-rolls only the empty-*name* check (`keystone_webhook.go:470-477`) and **misses the empty-value check** — `rules: {"identity:get": ""}` passes admission and reaches oslo.policy. The c5c3 webhook validates rules not at all (invalid rules wedge the CP indirectly via the keystone webhook). → extend `MergePolicies` to the pointer signature, use `ValidatePolicyRules` in both webhooks, delete the local copies.
3. **Confirmed-dead exports** (zero importers outside their own packages): `testutil/assertions` (158 LOC) and `testutil/builders` (entire packages), `config.InjectSecrets`, `types.MessagingSpec`, `tls.GetTLSSecret`, `database.RunDBSyncJob` (pure alias, 1 call site). Plus over-exported internals (`IsJobComplete/IsJobFailed`, `Is*Ready`, …) that widen the compat surface.
4. Smaller items: shared SHA-256 admin-password digest is implemented twice (`reconcile_bootstrap.go:88-89` vs `reconcile_korc.go:897-900`) although it is a cross-operator **contract** — one drift breaks the bootstrap re-run gate; `WaitForExternalSecret` collapses "doesn't exist" into "not ready" (`secrets/secrets.go:44-59`) — return a tri-state; `PruneImmutableConfigMaps` positional-string run + non-deterministic 1s-granularity tie-break (`config/config.go:227,257-259`); `bootstrap.Run` parses global flags internally (un-callable twice, untestable); the keystone/c5c3 `envtest_setup.go` files duplicate ~200 LOC of manager-hosting skeleton; the XOR invariants on `types.DatabaseSpec`/`CacheSpec` are documented but carry no CEL markers, so each operator re-implements them.

berendt:
## Summary

Implements the 2026-06 pre-production audit follow-up #479: consolidate
byte-identical cross-operator code into `internal/common`, make the unused
`policy` helpers the single source of truth (closing a real admission
validation gap), sweep confirmed-dead exports, and land the smaller
contract/quality fixes.

Closes #479

## Change set (in commit order)

**Problem 1 — shared instrumentation / metrics**
- `refactor(common): add shared sub-reconciler instrumentation package` —
  new `internal/common/instrumentation` with a prefix-parameterized
  `Metrics` (the two shared `*_reconcile_duration_seconds` /
  `*_reconcile_errors_total` vectors) and an `Instrumenter` that resolves
  the `condition_type` label from a per-operator map, with an `UNKNOWN`
  fallback.
- `refactor(keystone): adopt shared instrumentation, slim metrics` — keeps
  the keystone-only collectors (key-rotation age, db-sync) locally; the
  shared vectors and the `instrumentSubReconciler` wrapper now delegate to
  the common package. AST drift-guard tests retained.
- `refactor(c5c3): adopt shared instrumentation, drop metrics package` —
  the c5c3 `metrics` package had nothing but the two shared vectors, so it
  is removed and the controller constructs its `Metrics` inline.

**Problem 2 — policy reuse + closed validation gap**
- `refactor(policy): pointer-based MergePolicies with nil-copy semantics` —
  `MergePolicies` now takes `*PolicySpec`, absorbing c5c3's local
  `projectPolicyOverrides`/`copyPolicySpec` (deleted) as the single merge.
- `fix(common): reject empty policy rule names and values via CEL` /
  `fix(keystone): validate policy rule values in the admission webhook` —
  the keystone webhook now calls `policy.ValidatePolicyRules`, which adds
  the previously-missing empty-**value** check (`rules: {"identity:get": ""}`
  no longer reaches oslo.policy), with a matching CRD CEL rule.
- `fix(c5c3): validate policy rules at admission, reuse MergePolicies` — the
  ControlPlane webhook now validates `spec.global.rules` and
  `spec.services.keystone.policyOverrides.rules` (previously validated not
  at all).

**Problem 3 — dead-code sweep + over-export trim**
- `refactor(common): delete dead exports and unused test packages` —
  removes `testutil/assertions`, `testutil/builders`, `config.InjectSecrets`,
  `types.MessagingSpec`, `tls.GetTLSSecret`, and the `database.RunDBSyncJob`
  alias (its single caller now uses `job.RunJob`); regenerates deepcopy.
- `refactor(common): unexport package-internal readiness helpers` —
  `IsJobComplete/IsJobFailed`, `IsDatabaseReady/IsUserReady/IsGrantReady`,
  `IsCertificateReady` had zero cross-package importers. (`IsDeploymentReady`
  is deliberately kept exported — `testutil/simulators` consumes it.)

**Problem 4 — smaller contract / quality fixes**
- `feat(common): enforce DatabaseSpec/CacheSpec XOR via CEL markers` — the
  XOR invariants on the shared types are now type-level CEL, removing the
  per-operator field markers while keeping the webhook checks as
  defense-in-depth.
- `refactor(common): share admin-password digest across operators` —
  `secrets.AdminPasswordDigest` is the single source of truth for the
  cross-operator digest contract (keystone bootstrap re-run gate / c5c3 AC
  annotation).
- `refactor(common): make WaitForExternalSecret report a tri-state` —
  `(exists, ready, err)` so callers can tell "ExternalSecret absent" from
  "not yet synced".
- `refactor(common): give PruneImmutableConfigMaps a params struct` —
  `PruneOptions` + a stable, name-tie-broken sort so same-second ConfigMaps
  prune deterministically.
- `refactor(common): make bootstrap.Run flag parsing reentrant` — flags
  parse on a fresh `flag.FlagSet` so `Run` is callable/testable.
- `refactor(common): extract shared envtest manager-hosting helper` —
  `StartManagedEnvTest` removes ~300 LOC of duplicated, timing-sensitive
  manager-hosting skeleton across the two operators' `envtest_setup.go`.

## Verification

- `go build ./...`, `go test ./...` (unit) green in all three modules.
- `go test -tags=integration` green for `internal/common` and both
  operators (envtest, k8s 1.35).
- `gofumpt -l` clean on every changed file; docs free of CC-/REQ- IDs.

## Notes / deviations

- The issue bundles ~15 changes; the plan suggested separate per-Problem
  PRs, but this is delivered as one branch (one PR, per the closing-keyword
  linkage requirement). Commits stay small and grouped by Problem.
- `bootstrap.Run` uses a `ContinueOnError` FlagSet (returns the parse error
  to `main.go`) rather than `ExitOnError`, so the parser is testable.
- For `WaitForExternalSecret`, only the keystone caller (externally-provided
  ExternalSecrets) surfaces the new "not found yet" distinction in its
  status message; the c5c3 callers create their own ExternalSecret and log
  the flag instead.
- The envtest extraction covers the three webhook-hosting setup functions;
  the keystone-only no-webhook / minimal variants are left untouched (not
  cross-operator duplication).


internal/common/policy/policy.go
```diff
@@ -81,35 +81,68 @@ func RenderPolicyYAML(rules map[string]string) (string, error) {
...
+		return copyPolicySpec(override)
+	}
+	if override == nil {
+		return copyPolicySpec(base)
+	}
+
+	merged := &types.PolicySpec{}
 
-	// Merge rules: copy base first, then overlay override.
+	// Merge Rules: base is the base, override overwrites on key conflict.
```
gndrmnn, line 103: This sentence hurts to read. Drop the line or use the previous one.

gndrmnn (approved):

PR #511, opened 2026-06-19
fix(keystone-operator): NetworkPolicy egress breaks rotation CronJobs and brownfield/custom-port databases; operator health probe not whitelisted

### LINKED ISSUE #461: fix(keystone-operator): NetworkPolicy egress breaks rotation CronJobs and brownfield/custom-port databases; operator health probe not whitelisted
## Problems

All three only manifest under an **enforcing CNI** — kind/kindnet does not enforce NetworkPolicy, so the existing `network-policy` e2e cannot catch them.

1. **Rotation CronJobs lose kube-API access.** The policy's pod selector (`reconcile_networkpolicy.go:154-157`, `selectorLabels`) also matches the rotation CronJob pods, which carry `commonLabels` — a superset (`reconcile_fernet.go:303-306`, `reconcile_credential.go:306-309`, `reconcile_passwordrotation.go:402-405`). Egress allows only DNS:53, TCP 3306, TCP 11211 (`buildAutoEgressRules`, `reconcile_networkpolicy.go:170-219`), but the rotation scripts PATCH `kubernetes.default.svc:443` (`scripts/fernet_rotate.sh:18`, `credential_rotate.sh:19`, `admin_password_rotate.sh:31`). In production, **all key rotation stops at the first scheduled run**.
2. **Custom DB port / brownfield get no egress.** Port 3306 is hardcoded and only emitted when `Database.ClusterRef != nil` (`reconcile_networkpolicy.go:196-203`), but `dbPort()`/`resolveDatabaseHost()` honour `spec.database.port`, and brownfield (`spec.database.host`) gets no DB rule at all. The readiness probe is a TCP connect to exactly that host:port (`reconcile_deployment.go:75-78`) → probes fail → **all pods depooled**. Same for brownfield `cache.Servers`.
3. **Operator health check not whitelisted.** `reconcileHealthCheck` GETs the keystone Service from the operator pod (`reconcile_healthcheck.go:70-84`); ingress on 5000 only admits `spec.networkPolicy.ingress` sources + gateway namespace (`reconcile_networkpolicy.go:117-140`). Unless users list the operator namespace, `KeystoneAPIReady` flips False permanently → aggregate `Ready=False` for a healthy deployment.

## Proposed fix

- Give rotation CronJob pods a distinct label set (do not reuse Deployment selector labels), or auto-derive an apiserver egress rule (TCP 443/6443) when `spec.networkPolicy` is set.
- Derive the DB egress port from `dbPort(keystone)` and emit rules in brownfield mode too (DB and cache).
- Auto-append an ingress peer for the operator namespace (downward API/env), or at minimum webhook-warn.
- Consider an e2e leg on a NetworkPolicy-enforcing CNI (e.g. kind+Calico/Cilium) for the `network-policy` scenario.

berendt:
## Summary

Fixes the three NetworkPolicy defects from issue #461 that only manifest under
an enforcing CNI (kindnet does not enforce NetworkPolicy, so the existing
`network-policy` e2e cannot catch them). All production changes are in
`operators/keystone/internal/controller/`; no CRD/API types change, so no
schema regeneration is required.

Closes #461

## The three problems

1. **Rotation CronJobs lose kube-API access.** The fernet/credential/
   admin-password rotation CronJob pods carry `commonLabels` — a superset of
   the Deployment `selectorLabels` — so they are selected by the per-CR
   NetworkPolicy. Their scripts PATCH the rotated keys back to a Secret via
   `kubernetes.default.svc`, but egress allowed only DNS/3306/11211, so every
   scheduled rotation stopped at its first run.
2. **Custom DB port / brownfield get no egress.** Port 3306 was hardcoded and
   gated on `database.clusterRef`, while `dbPort()` honours
   `spec.database.port` and the readiness probe TCP-connects to that exact
   host:port. Brownfield (`spec.database.host`) and brownfield cache
   (`cache.servers`) got no egress at all.
3. **Operator health check not whitelisted.** `reconcileHealthCheck` GETs the
   Keystone Service on TCP 5000 from the operator pod, but ingress admitted
   only the user-declared sources and the gateway namespace, so
   `KeystoneAPIReady` flipped `False` permanently for a healthy deployment.

## Changes (commit by commit)

1. **`feat(keystone-operator): resolve operator namespace at startup`** — adds
   `DetectOperatorNamespace()` (`POD_NAMESPACE` env, falling back to the
   mounted ServiceAccount namespace file), a `KeystoneReconciler.OperatorNamespace`
   field, and the `main.go` wiring. Enabler; the field is consumed in commit 3.
2. **`fix(keystone-operator): add kube-apiserver egress to NetworkPolicy`** —
   `buildAutoEgressRules` now appends an always-on kube-apiserver rule
   (TCP 443+6443, covering the post-DNAT pod port on Calico/Cilium), derives
   the DB port from `dbPort()` and emits it in both managed and brownfield
   modes, and derives cache ports from the cache spec (`cacheEgressPorts`
   parses brownfield `host:port` strings) in both modes. Fixes problems 1 & 2.
3. **`fix(keystone-operator): whitelist operator namespace in NetworkPolicy`** —
   `buildKeystoneNetworkPolicy` takes the operator namespace and appends an
   ingress peer for it (mirroring the gateway-namespace peer);
   `reconcileNetworkPolicy` threads `r.OperatorNamespace` through. Fixes
   problem 3.
4. **`test(keystone-operator): assert new NetworkPolicy egress/ingress in e2e`** —
   extends the `network-policy` chainsaw suite with egress port-presence
   assertions (JMESPath, robust to rule ordering) and the operator-namespace
   ingress peer, and documents the kindnet non-enforcement scope caveat.
5. **`docs(keystone): document NetworkPolicy apiserver egress and op-ns ingress`** —
   refreshes the `NetworkPolicySpec` reference (auto-derived egress table,
   new auto-added ingress-peers table) and the `reconcileHealthCheck` section.

## Tests

- New unit tests in `reconcile_networkpolicy_test.go` assert by **port
  presence** (not index): apiserver 443+6443 in both modes, custom DB port
  (`33060`, not 3306), brownfield cache port (`11212`, not 11211), the
  no-cache-configured case (no cache rule), plus a table-driven `cacheEgressPorts`
  test covering bare-host, IPv6, dedup, and malformed-port edge cases.
- New tests for the operator-namespace ingress peer (present, empty-namespace
  no-op, gateway+operator coexistence, and reconcile wiring).
- New `operator_namespace_test.go` covers env-wins, file fallback,
  whitespace-trim, and the neither-available (`""`) edge case.

## Local verification

- `go test ./operators/keystone/...` — pass
- `golangci-lint run ./internal/controller/...` — 0 issues
- `make format-check` — pass
- `chainsaw lint test --file tests/e2e/keystone/network-policy/chainsaw-test.yaml` — valid

## Out of scope

Issue #461 item 4 ("Consider an e2e leg on a NetworkPolicy-enforcing CNI") is a
new CI infrastructure concern beyond this fix's blast radius. The existing
kindnet e2e can only assert the policy *spec*, not enforcement; this is noted in
a scope caveat in the chainsaw test.

---

_Implemented by [planwerk-review](https://github.com/planwerk/planwerk-review) e9105cf with Claude:claude-opus-4-8_

### DIFF
--- a/docs/reference/keystone/keystone-crd.md
+++ b/docs/reference/keystone/keystone-crd.md
@@ -646,15 +646,16 @@ Configures network isolation for the Keystone API pods. This is a
 pointer field (`*NetworkPolicySpec`) on `KeystoneSpec` — when `nil`, no
 NetworkPolicy is managed and the `NetworkPolicyReady` condition is set to
 `True` with reason `NetworkPolicyNotRequired`. When set, the operator creates
-a NetworkPolicy that restricts ingress on TCP 5000 to the declared sources and
-auto-derives egress rules for DNS, MariaDB (when `database.clusterRef` is set),
-and Memcached (when `cache.clusterRef` is set). Removing the field deletes the
-NetworkPolicy on the next reconcile.
+a NetworkPolicy that restricts ingress on TCP 5000 to the declared sources —
+plus the operator's own namespace (so the operator health check can reach the
+API) and, when `spec.gateway` is set, the gateway namespace — and auto-derives
+egress rules for DNS, the kube-apiserver, the database, and the cache. Removing
+the field deletes the NetworkPolicy on the next reconcile.
 
 | Field | Type | Required | Default | Description |
 | --- | --- | --- | --- | --- |
 | `ingress` | `[]NetworkPolicyIngressSource` | Yes | — | Sources allowed to reach Keystone API on TCP 5000. At least one entry required (enforced by CEL and webhook). |
-| `additionalEgress` | `[]networkingv1.NetworkPolicyEgressRule` | No | `nil` | Extra egress rules appended after the auto-derived rules. Use for brownfield backends or external integrations not covered by `ClusterRef` auto-derivation. |
+| `additionalEgress` | `[]networkingv1.NetworkPolicyEgressRule` | No | `nil` | Extra egress rules appended after the auto-derived rules. The auto-derived rules already cover the database and cache in both managed and brownfield modes, so reserve this for external integrations beyond those backends. |
 
 ### NetworkPolicyIngressSource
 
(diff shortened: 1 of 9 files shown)


berendt:
## Fix Report (iteration 1)

### Per check
- **e2e-operator (keystone)**
  - **Category:** test
  - **Root cause:** The `keystone-network-policy` chainsaw suite (step #2) asserted auto-derived egress with `"(egress[].ports[].port | contains(@, \`53\`))": true`. Chainsaw's JMESPath `contains()` compares by strict type; the rendered NetworkPolicy ports are integers (int64 from the unstructured object) while the `` `53` `` literals are JSON floats, so every check returned `false` and the assert timed out after 5m (09:49:52→09:54:52). The production reconciler is correct — the pods served `GET /v3 200` and the egress ports are present.
  - **Fix:** `tests/e2e/keystone/network-policy/chainsaw-test.yaml` — replaced the five `contains()` expressions with a declarative `egress:` block listing all four rules in the reconciler's deterministic order (DNS UDP/TCP 53; apiserver TCP 443/6443; DB TCP 3306; cache TCP 11211). This uses the same int-typed scalar form chainsaw already matches for the ingress `port: 5000` in the same file (which passed in the failing run). Comment rewritten to describe the actual idiom; commit message body updated to match.
  - **Local verification:** `go test ./internal/controller/ -run 'TestBuildKeystoneNetworkPolicy|TestReconcileNetworkPolicy' -count=1` → **ok** (proves the reconciler emits 53/443/6443/3306/11211 in order); `python3 yaml.safe_load_all` on the test → **YAML OK**. The full chainsaw e2e is **not reproducible in this environment** (needs a kind cluster + deployed operator); relying on the unit-test proof of reconciler output plus the empirical fact that the declarative `port: 5000` assert in this same file passed in the failing CI run.
  - **Regression test:** n/a — fix is inside test code (the chainsaw assertion); the reconciler's egress ports are already covered by the passing `TestBuildKeystoneNetworkPolicy_AutoDerivedEgress_*` unit tests.

### Diff summary
- **Files:** `tests/e2e/keystone/network-policy/chainsaw-test.yaml`
- **Approx lines added/removed:** +28 / -9 (vs. prior tip)
- **Commit strategy:** folded into `test(keystone-operator): assert new NetworkPolicy egress/ingress in e2e` (originally `5a1941e9`, now `3cb4472f`) via `git commit --fixup` + `--autosquash` rebase from the merge-base; that commit was then reworded so its body no longer claims a JMESPath/reorder-tolerant assertion. The docs commit was re-applied unchanged on top; the three base fix commits (`8eb58fb8`, `571653c8`, `7390615d`) were untouched. Pushed with `--force-with-lease`.

### Status
STATUS: DONE

---

_Fix report generated by [planwerk-review](https://github.com/planwerk/planwerk-review) 22cbb64 fix with Claude_

gndrmnn (approved):

Issue #480, opened 2026-06-11
test: pre-production gaps — watch-mapper coverage, operator helm-upgrade e2e, chaos CI gating, stuck-finalizer deletion, metrics lifecycle


berendt:
Overall test quality is high (keystone controller 89.7% incl. integration, real failure-path tests, zero skips). Five verified gaps remain:

1. **Watch mappers and `SetupWithManager` at 0% in unit AND integration profiles.** Integration tests build the controller inline, bypassing `SetupWithManager` — `mariaDBToKeystoneMapper` (`keystone_controller.go:851`), `clusterSecretStoreToKeystoneMapper` (`:878`) and both c5c3 setups are never executed by any test. A regression silently degrades reconcile latency to resync cadence or crashes the manager. → the mappers are plain `handler.MapFunc` closures: unit-test directly with a fake client (~50 lines each).
2. **No operator/CRD upgrade-in-place e2e.** All existing "upgrade" scenarios upgrade the *Keystone service* image; `grep -rl "helm upgrade" tests/` is empty. Given the cross-release bootstrap history (#420), this is the highest-risk untested path. → Chainsaw test: install last released chart from GHCR, CR to Ready, `helm upgrade` to the locally built chart+CRDs, assert Ready persists, no new bootstrap Job, `status.installedRelease` unchanged. (Related: #369.)
3. **Chaos suite never gates CI.** `.github/workflows/ci.yaml:853-904`: `continue-on-error: true` AND triggers only on `tests/e2e-chaos/**` changes or the `run-chaos` label — the operator-restart/partition scenarios effectively never run for operator code PRs and can never fail the build; the "revisit after 2-4 weeks" comment (line 849) is overdue. → flip `continue-on-error: false` for the pod-chaos leg, add `has-e2e-operators` to the trigger. (Coverage expansion itself is tracked in #453.)
4. **Deletion with a stuck finalizer untested.** `deletion-cleanup` only covers a healthy MariaDB operator. → chaos test: scale mariadb-operator to 0, delete the Keystone CR, assert a documented behavior (condition/event), scale up, assert completion.
5. **Keystone metrics global-collector path 0%** (`internal/metrics/collectors.go:125-195`): registration, recording and `DeleteForKeystone` label cleanup (stale-series leak risk) never executed in-process. → unit-test against a fresh `prometheus.NewRegistry()`.

[...]

Issue #462, opened 2026-06-11
fix(keystone-operator): reconcile fights the HPA over .spec.replicas when autoscaling is enabled


berendt:
## Problem

`buildKeystoneDeployment` always sets replicas from the spec (`reconcile_deployment.go:196,206`), and the shared update path applies the desired spec unconditionally (`internal/common/deployment/deployment.go:94-97`). With `spec.autoscaling` set, `reconcileHPA` creates an HPA targeting the same Deployment: every reconcile resets `.spec.replicas`, the HPA scales back, each Deployment write fires a watch event (`Owns(&appsv1.Deployment{})`) that re-runs Reconcile — a continuous scale-up/scale-down loop with real pod churn whenever the HPA's desired count differs from `spec.replicas`.

## Proposed fix

- In `buildKeystoneDeployment`, leave `Replicas` nil when `keystone.Spec.Autoscaling != nil`.
- In `EnsureDeployment`'s update path, preserve the live value when desired replicas is nil:
  ```go
  if deploy.Spec.Replicas == nil {
      deploy.Spec.Replicas = existing.Spec.Replicas
  }
  ```
- Keep an explicit value on the create path so the apiserver does not default to 1.
- Add a unit test: HPA enabled + live replicas≠spec.replicas ⇒ no Deployment update.

_Found by the 2026-06 pre-production audit (verified at file:line on `main` @ 0a2f19ba)._

berendt:
## Implementation Plan (issue #462)

### Summary
When `spec.autoscaling` is set, the operator currently pins the Deployment's `.spec.replicas` to `spec.replicas` on every reconcile, fighting the HPA and creating a scale-up/scale-down loop with pod churn. The smallest fix matches the issue's proposed approach exactly: make `buildKeystoneDeployment` leave `Replicas` nil when `spec.autoscaling != nil`, and teach the shared `EnsureDeployment` to (a) preserve the live replica value on the update path when desired replicas is nil and (b) write an explicit replica count on the create path so the apiserver does not implicitly default a nil value to 1. No API/CRD types change, so no code generation is required.

### Ground-Truth Notes
- `reconcile_deployment.go:196,206` (`buildKeystoneDeployment`) — confirmed. Line 196 `replicas := int32(keystone.Spec.Replicas)`, line 206 `Replicas: &replicas,`; replicas is always set unconditionally. File imports `k8s.io/utils/ptr` (line 17).
[...]

PR #512, opened 2026-06-19
refactor(common): adopt Server-Side Apply for the Ensure* family

### LINKED ISSUE #474: refactor(common): adopt Server-Side Apply for the Ensure* family — non-converging diffs and unconditional updates
## Problems

1. **Spec DeepEqual against server-defaulted objects never converges.** The guard `!apiequality.Semantic.DeepEqual(existing.Spec, desired.Spec)` compares a defaulted live spec against a non-defaulted desired spec. Verified instances: mariadb `User.spec.maxUserConnections` (`+kubebuilder:default=10`, never set by `buildUser`) → `ensureUser` issues Update + re-Get on **every** pass forever (`internal/common/database/database.go:46,103,135`); `PushSecret.spec.updatePolicy`/`refreshInterval` defaults → `EnsurePushSecret` updates every PushSecret every pass (`secrets/secrets.go:145`).
2. **Unconditional updates for Deployment/Service/CronJob** (`deployment/deployment.go:94-97`, `job/job.go:175-182`, documented as DECISION I-001): ~6 writes per reconcile pass even when fully converged; elevated Conflict exposure surfaces as transient condition noise. `EnsurePDB`/`EnsureHPA` in the same package already prove the change-detection pattern.
3. **Operational side effect:** `EnsureCronJob` reverts a manual `kubectl patch cronjob --suspend=true` within one reconcile — and the fernet/credential rotation CronJobs have no `spec.suspend` field, so an SRE **cannot pause key rotation during an incident**.
4. `ensureNetworkPolicy`/`ensureHTTPRoute` re-implement the snapshot/ownerref/DeepEqual-update pattern by hand (~130 lines) that a generic helper would cover.

## Proposed fix

Migrate the Ensure* family to Server-Side Apply with a fixed field manager (defaults stop participating in the diff; only owned fields are managed), or extend the PDB/HPA snapshot-diff pattern across the family. Provide a generic `EnsureObject[T client.Object]`. Add a suspend escape hatch for the rotation CronJobs. Treat `IsConflict` as benign requeue.

Related: #361 collects the broader reconcile-performance hotspots; this issue is the structural fix for its write-amplification class (e.g. B3).

berendt:
## Summary

Adopts Server-Side Apply (SSA) as the single create-or-update primitive for the
`Ensure*` family, replacing both the non-converging `DeepEqual`-then-`Update`
guards and the unconditional `Update` writes. A new generic
`apply.EnsureObject[T client.Object]` applies the desired object under a fixed
field manager (`forge-operator`) with `ForceOwnership`, wrapped in
`retry.RetryOnConflict`. Because the field manager owns only the fields the
builder sets, server-defaulted fields stop participating in the diff and
reconciliation converges; the hand-rolled snapshot/DeepEqual + sentinel-metadata
machinery in the keystone reconcilers collapses to a single call. Also adds a
declarative `suspend` escape hatch to the fernet/credential rotation CronJobs.

Closes #474

## What changed (in commit order)

1. **`refactor(common): add generic server-side apply EnsureObject`** — new
   `internal/common/apply` package. `EnsureObject` sets the owner reference,
   stamps the GVK, and applies via the non-deprecated `client.Client.Apply`
   using `ApplyConfigurationFromUnstructured`, wrapped in `RetryOnConflict`
   (the issue's "treat `IsConflict` as benign requeue"). Unit + envtest tests
   cover the apply shape, conflict retry, GVK-unknown error path, no-op
   convergence, and Update→Apply takeover.
2. **`refactor(common): apply Deployment/Service/PDB/HPA via SSA`** — removes the
   unconditional `Update` (problem #2) and the snapshot-diff in PDB/HPA; deletes
   DECISION I-001 and `normalizeMap`. Keeps the pre-apply `Get` only for the
   immutable-selector delete+recreate, the issue #462 nil-replicas handling, and
   the immutable-Service-field fail-fast.
3. **`refactor(common): apply CronJob via server-side apply`** — `EnsureCronJob`
   only; the immutable-Job `RunJob*` paths are unchanged.
4. **`refactor(common): apply PushSecret via server-side apply`** — fixes the
   `updatePolicy`/`refreshInterval` every-pass rewrite (problem #1) and updates
   the c5c3 `reconcile_korc.go` comment that described the removed guard.
5. **`refactor(common): apply MariaDB Database/User/Grant via SSA`** — fixes the
   `User.maxUserConnections` non-convergence (problem #1), proven by a new
   envtest test that the server default survives and a repeated apply does not
   rewrite.
6. **`refactor(keystone): apply NetworkPolicy/HTTPRoute via SSA helper`** —
   deletes the hand-rolled snapshot-diff and the entire HTTPRoute
   sentinel-annotation bookkeeping (~130 lines, problem #4); SSA field ownership
   removes dropped keys natively and preserves third-party keys.
7. **`feat(keystone): add suspend escape hatch to fernet/credential rotation`** —
   adds `spec.fernet.suspend` / `spec.credentialKeys.suspend` (default `false`,
   mirroring `trustFlush.suspend`) so an SRE can pause key rotation during an
   incident (problem #3). CRDs regenerated with controller-gen v0.21.0.
8. **`docs(keystone): document server-side apply and rotation suspend`** —
   shared-library + Keystone CRD/reconciler reference docs.
9. **`test(c5c3): assert PushSecret spec idempotency under SSA`** — adjusts the
   c5c3 no-churn unit test for fake-client SSA semantics (see Notes).

## Notes for reviewers

- **SSA API choice (deviation from the plan):** `client.Apply` (the patch var)
  is deprecated in controller-runtime v0.24.1 and would fail `staticcheck`
  (SA1019). `EnsureObject` therefore uses the non-deprecated
  `client.Client.Apply` with `ApplyConfigurationFromUnstructured`. Same SSA
  mechanism, non-deprecated API. `ToUnstructured` respects `omitempty`, so the
  problem #1 fix (omitted defaults not owned) holds — verified by envtest.
- **Fake-client SSA fidelity:** the controller-runtime fake client cannot apply
  an unstructured NetworkPolicy/HTTPRoute with its default typed converter and
  bumps `resourceVersion` on every apply (no no-op detection). So the keystone
  NP/HTTPRoute unit tests use the deduced type converter and assert
  *idempotency*; the *no-write convergence* and *cross-manager annotation
  preservation* guarantees are asserted against a real API server in envtest
  (`internal/common/apply`, `internal/common/deployment`, `internal/common/database`,
  `internal/common/secrets`, and a new
  `TestIntegration_HTTPRoute_PreservesThirdPartyAnnotation`).
- **envtest fake CRDs:** the minimal MariaDB fake CRDs now mirror the real
  `maxUserConnections` default and tolerate the non-omitempty struct fields a
  typed apply emits (`mariaDbRef`, `passwordPlugin`) via
  `x-kubernetes-preserve-unknown-fields`.
- **Out of scope:** `tls.EnsureCertificate` (no documented problem) and the
  `helm upgrade` Update→Apply migration e2e (sibling #480).
- **Architecture submodule:** `architecture/docs/09-implementation/02-shared-library.md`
  lives in a git submodule and is not updated here; the in-repo `docs/`
  reference is updated. A submodule doc update can follow separately.

## Verification

- `make lint` — clean (all three modules)
- gofumpt format-check — clean
- `go test ./...` unit — `internal/common`, `operators/keystone`,
  `operators/c5c3` all pass
- envtest integration — `internal/common` (apply/deployment/database/secrets/job)
  and the full `operators/keystone` controller suite pass; `operators/c5c3`
  controller suite pass
- `make verify-crd-sync` — passes
- The fernet/credential rotation Chainsaw suites validate with `chainsaw lint`
  (not run against a cluster in this branch)

---

_Implemented by [planwerk-review](https://github.com/planwerk/planwerk-review) e9105cf with Claude:claude-opus-4-8_

### DIFF
--- a/docs/reference/backend/kubernetes-packages.md
+++ b/docs/reference/backend/kubernetes-packages.md
@@ -12,11 +12,16 @@ core Kubernetes resources (Deployments, Services, Jobs, CronJobs, ConfigMaps).
 
 All packages share these conventions:
 
-- **Idempotent create-or-update** — `Ensure*` functions create a resource if it does not
-  exist or update its spec if it already exists. They use `Get` + `Create`/`Update` (not
-  `controllerutil.CreateOrUpdate`) for explicit control.
-- **Owner references** — `controllerutil.SetControllerReference` is called on create so
-  the resource is garbage-collected when the owning CR is deleted.
+- **Idempotent create-or-update via Server-Side Apply** — `Ensure*` functions apply the
+  desired object through the generic `apply.EnsureObject` helper, which uses Server-Side
+  Apply under the fixed field manager `forge-operator`. The field manager owns only the
+  fields the builder sets, so server-defaulted fields the builder omits are never claimed
+  or overwritten and a converged object is applied without a write. The apply is wrapped
+  in `retry.RetryOnConflict`, so a benign field-manager conflict is absorbed as an internal
+  retry rather than surfacing as condition noise.
+- **Owner references** — `controllerutil.SetControllerReference` is called on every apply
+  (not only on create) so the resource is garbage-collected when the owning CR is deleted
+  and the reference is re-enforced if it drifts.
 - **Readiness reporting** — Functions that create resources return `(bool, error)`, where
   `true` means the resource is ready and `false` means it exists but is not yet ready.
 - **Error wrapping** — All errors include context via `fmt.Errorf` with `%w` for
@@ -26,6 +31,7 @@ All packages share these conventions:
 
 | Package | Import Path |
 | --- | --- |
+| `apply` | `github.com/c5c3/forge/internal/common/apply` |
 | `config` | `github.com/c5c3/forge/internal/common/config` |
 | `database` | `github.com/c5c3/forge/internal/common/database` |
 | `deployment` | `github.com/c5c3/forge/internal/common/deployment` |
@@ -46,6 +52,44 @@ These packages import typed Go structs from external operator modules:
 
 > **Note:** The PushSecret API is `v1alpha1` (unstable). Its schema may change in future
 > ESO releases. Pin the ESO module version in `go.mod` to avoid breakage.
+
+---
+
+## Package: `apply`
+
+Provides the generic Server-Side Apply create-or-update primitive that backs the
+`Ensure*` family.
+
+### EnsureObject
+
+```go
+func EnsureObject[T client.Object](
+    ctx context.Context,
+    c client.Client,
+    scheme *runtime.Scheme,
+    owner client.Object,
+    obj T,
+    fieldManager string,
+) error
+```
+
+Creates or updates `obj` via Server-Side Apply under `fieldManager` and sets `owner` as
+the controller reference.
+
+**Behavior:**
+
+- Sets the controller owner reference and stamps the object's GVK (objects built in-code
+  carry an empty `TypeMeta`, which Server-Side Apply requires).
+- Applies the object with `client.FieldOwner(fieldManager)` and `client.ForceOwnership`,
+  wrapped in `retry.RetryOnConflict` so a benign field-manager conflict is retried
+  internally rather than surfaced.
+- The field manager owns only the fields the builder sets, so server-defaulted fields the
+  builder omits are never claimed and a converged object is applied without a write.
+- Decodes the server response back into `obj`, so callers may read fresh status (e.g.
+  readiness) without an extra `Get`.
+
+**Field manager:** the package exports `apply.FieldManager` (`"forge-operator"`), the
+stable field-manager name shared by all `Ensure*` helpers.
 
 ---
 
@@ -224,10 +268,11 @@ Creates or updates a MariaDB Database CR.
 
 **Behavior:**
 
-- On create: sets controller owner reference, creates the resource, returns `(false, nil)`.
-- On update: overwrites `existing.Spec` with the provided spec.
+- Applies the desired Database via `apply.EnsureObject` (Server-Side Apply); the field
+  manager owns only the fields the builder sets, so the server default on
+  `maxUserConnections` is preserved and a converged Database is not rewritten.
 - Readiness is determined by the package-internal `isDatabaseReady` helper on the
-  existing resource.
+  server-fresh apply response.
 
 ### EnsureDatabaseUser
 
@@ -279,9 +324,11 @@ Deployment exists but is not yet ready; `(false, error)` on failure.
 
 **Behavior:**
 
-- On create: sets controller owner reference, creates the resource, returns `(false, nil)`.
-- On update: overwrites `existing.Spec` with the provided spec.
-- Readiness is determined by `IsDeploymentReady` on the existing resource.
+- Applies the desired Deployment via `apply.EnsureObject` (Server-Side Apply); a converged
+  Deployment is not rewritten on every reconcile.
+- Retains a pre-apply `Get` only to delete-and-recreate on an immutable-selector change and
+  to preserve the HPA-owned replica count when `spec.replicas` is left nil.
+- Readiness is determined by `IsDeploymentReady` on the server-fresh apply response.
 
 ### EnsureService
 
@@ -299,10 +346,12 @@ Creates or updates a Service. Does not report readiness (Services are ready imme
 
 **Behavior:**
 
-- On create: sets controller owner reference, creates the resource.
-- On update: preserves the existing `ClusterIP` and `ClusterIPs` values assigned by the
-  API server before overwriting the spec. This prevents accidental ClusterIP reassignment,
-  which would break existing DNS-based service discovery.
+- Applies the desired Service via `apply.EnsureObject` (Server-Side Apply). The builder
+  leaves server-assigned fields (`ClusterIP`, `ClusterIPs`, `IPFamilies`, `NodePort`)
+  unset, so the field manager never owns them and the API server keeps the values it
+  assigned — no hand-rolled preservation is needed.
+- Retains a pre-apply `Get` only to fail fast when the caller explicitly sets one of those
+  immutable fields to a conflicting value.
 
 ### IsDeploymentReady
 
@@ -386,9 +435,8 @@ Creates or updates a CronJob with a controller owner reference.
 
 **Behavior:**
 
-- On create: sets controller owner reference, creates the resource.
-- On update: overwrites `existing.Spec` with the provided spec. CronJob spec updates
-  take effect on the next scheduled run.
+- Applies the desired CronJob via `apply.EnsureObject` (Server-Side Apply); a converged
+  CronJob is not rewritten, and spec changes take effect on the next scheduled run.
 
 The Job completion and permanent-failure checks (`isJobComplete` / `isJobFailed`)
 are package-internal helpers consumed by `RunJob`.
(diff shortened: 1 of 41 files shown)


berendt:
## Fix Report (iteration 1)

### Per check
- **e2e-operator (keystone)**
  - **Category:** test (chainsaw e2e — flaky/racy assertion; not a build/infra failure)
  - **Root cause:** `tests/e2e/keystone/upgrade-abort/chainsaw-test.yaml` step #4 asserts `upgradePhase: Expanding` / `DatabaseReady=ExpandInProgress`, then runs a **single-shot** `kubectl get job keystone-upgrade-abort-db-expand`. That status is written at upgrade *detection* (`reconcile_database.go` returns `Requeue: true` **before** the expand Job is created), so the status flip is observable while the Job does not yet exist. `reconcileExpand` creates the Job on the *next* reconcile pass. The single-shot check raced that create and failed with `ERROR: Job ... not found`. The catch-block diagnostics prove the operator is correct — the Job was present `Running 0/1, AGE 7s` (image `2026.2`, ImagePullBackOff, exactly as the test intends) ~7s later. The PR's SSA `Ensure*` refactor (plus pre-existing status-update 409 jitter) tightened reconcile timing enough to lose the race. `upgrade-flow` (passed) avoids this by checking jobs only *after* the upgrade completes.
  - **Fix:** `tests/e2e/keystone/upgrade-abort/chainsaw-test.yaml` — replaced the single-shot `kubectl get job` script with a chainsaw-native `assert` on `batch/v1 Job keystone-upgrade-abort-db-expand`, which polls up to the step's `assert` timeout (5m). This is condition-based (the project-endorsed pattern over single-shot/`sleep`) and matches the existing idiom in `policy-validation` and `schema-drift-detection`. The check is not weakened — it still requires the Job to exist; it just no longer races its creation.
  - **Local verification:** Not reproducible in this environment (requires a kind cluster + chainsaw + built operator images). Performed: `python3 yaml.safe_load_all` → `YAML OK`; verified the `assert: kind: Job` idiom is already used by two sibling suites. Relying on CI to confirm green.
  - **Regression test:** n/a — the change is inside test code itself (hardening a racy assertion); the operator behavior was already correct and unchanged.

### Diff summary
- **Files:** `tests/e2e/keystone/upgrade-abort/chainsaw-test.yaml`
- **Approx lines added/removed:** +13 / −11
- **Commit strategy:** New standalone commit `619840ab` — the upgrade-abort test is not touched by any branch commit (`origin/main..HEAD`); the race pre-exists on `main` and was exposed (not introduced) by the SSA timing changes, so it belongs to no existing commit. No autosquash/rebase was needed; the push was a fast-forward published with `--force-with-lease`.

### Status
STATUS: DONE_WITH_CONCERNS

Concern for a human reviewer: the failing check is an e2e chainsaw test the PR does not own, and the operator code (the PR's surface) is functionally correct — the only defect is the test's single-shot timing assumption. I made the minimal out-of-scope change to make the check deterministic, as the PR's own files cannot fix a test-timing race. The fix could not be verified locally (no kind/chainsaw here); CI confirmation is needed. I deliberately did **not** sweep the rest of the e2e suite for similar latent single-shot races (none others failed this run) to keep the change minimal.

---

_Fix report generated by [planwerk-review](https://github.com/planwerk/planwerk-review) 22cbb64 fix with Claude_

berendt:
## Fix Report (iteration 2)

### Per check
- **e2e-operator (keystone)**
  - Category: test (e2e assertion, Chainsaw)
  - Root cause: Commit `3a1dc6e6` deliberately removed the `keystone.openstack.c5c3.io/managed-annotations` sentinel-annotation bookkeeping from the HTTPRoute reconciler (Server-Side Apply field ownership replaces it, per issue #474 problem #4), but the `keystone-httproute` Chainsaw test Step 3 still asserted that sentinel annotation on the rendered HTTPRoute. The operator no longer stamps it, so the assert never matched and timed out at 5m (`--- FAIL: chainsaw/keystone-httproute (300.69s)`), while the other 45 suites passed.
  - Fix: `tests/e2e/keystone/httproute/chainsaw-test.yaml` — removed the stale `keystone.openstack.c5c3.io/managed-annotations: example.com/test-annotation` line from the Step 3 HTTPRoute assertion, leaving the still-valid `example.com/test-annotation: initial` (which the operator copies verbatim from `spec.gateway.annotations`).
  - Local verification: full Chainsaw e2e is not reproducible in this environment (needs a kind cluster, Gateway API CRD, and a built operator image) — relying on CI. Verified locally that the YAML parses (`python3 yaml.safe_load_all` → `YAML OK`) and confirmed by code + git-history analysis that `buildKeystoneHTTPRoute` produces only `example.com/test-annotation: initial`; `git grep managed-annotations` returns no code/test references (only historical `.planwerk/reviews/` records, intentionally untouched).
  - Regression test: n/a — the fix is inside e2e test/fixture code itself (the assertion is the test), so per the diagnosis workflow no separate regression test is added; the cross-manager annotation-preservation guarantee is already covered by the new `TestIntegration_HTTPRoute_PreservesThirdPartyAnnotation` envtest added in the same commit.

### Diff summary
- Files: `tests/e2e/keystone/httproute/chainsaw-test.yaml`
- Approx lines added/removed: +0/-1
- Commit strategy: folded into `a42fb4fa` (was `3a1dc6e6`) `refactor(keystone): apply NetworkPolicy/HTTPRoute via SSA helper` via `git commit --fixup` + `git rebase --autosquash` — the e2e assertion update belongs to the commit that removed the sentinel behavior it was asserting.

### Status
STATUS: DONE

This is a different root cause than a repeated approach: the SSA refactor changed observable HTTPRoute metadata, and the e2e assertion was the last remaining reference to the removed sentinel. The fix aligns the test with the PR's intended (and otherwise fully-tested) behavior rather than weakening any check.

---

_Fix report generated by [planwerk-review](https://github.com/planwerk/planwerk-review) 22cbb64 fix with Claude_

gndrmnn (approved):

berendt:
# Rebase analysis: C5C3/forge#512

> Rebased onto `main`  



## d0fdb06 refactor(common): add generic server-side apply EnsureObject

## e5e3427 refactor(common): apply Deployment/Service/PDB/HPA via SSA

## 0adad47 refactor(common): apply CronJob via server-side apply

## cc85ced refactor(common): apply PushSecret via server-side apply

## 5707c69 refactor(common): apply MariaDB Database/User/Grant via SSA

## d3599f4 refactor(keystone): apply NetworkPolicy/HTTPRoute via SSA helper

## c0911e7 feat(keystone): add suspend escape hatch to fernet/credential rotation

## 9123ddc docs(keystone): document server-side apply and rotation suspend

## 6e87720 test(c5c3): assert PushSecret spec idempotency under SSA

## 48e8eb2 test(keystone): poll for db-expand Job in upgrade-abort e2e

## Summary

The rebase landed cleanly despite heavy overlap: the SSA refactor rewrites exactly the internal/common files (deployment, job, secrets, database) and the keystone NetworkPolicy reconciler that upstream also churned (unexporting readiness helpers, deleting dead exports, the WaitForExternalSecret tri-state, the admin-password digest, and the NetworkPolicy egress/ingress hardening). All three Go modules build, no stale references to upstream-renamed/deleted symbols remain, and the NetworkPolicy SSA rewrite preserves the upstream kube-apiserver egress and operator-namespace ingress in both code and tests. The PushSecret SSA change coexists correctly with the tri-state WaitForExternalSecret and digest refactors, the K-ORC test resolves against the upstream per-area split, the suspend CRD regen is purely additive and keeps the upstream DatabaseSpec/CacheSpec XOR CEL markers, and the docs merge introduces no references to removed symbols.

> [!IMPORTANT]
> **Recommendation**: Safe to push as-is. No adjustments are needed for any commit. The only residual caveat is that `go test`/`go vet` were blocked by the local permission policy, so test-binary compilation was verified by manual symbol-resolution tracing rather than execution; every overlapping test symbol (validPushSecret, tri-state WaitForExternalSecret call sites, buildKeystoneNetworkPolicy's 2-arg signature with the operator-namespace assertions, adminAppCredentialPushSecretName/childNamespace, IsDeploymentReady, and the envtest helpers SetupEnvTest/SharedScheme/SkipIfEnvTestUnavailable/setupEnvTestWithController) was confirmed present. Running `go test ./...` across the three modules before merge is a cheap final confirmation.

---

Issue #466, opened 2026-06-11
feat(api): enforce field immutability via oldSelf CEL rules; reject openStackRelease downgrades


berendt:
## Problem

No field in either CRD is immutable. Both `ValidateUpdate` implementations discard `oldObj` (`keystone_webhook.go:236-238`, `controlplane_webhook.go:201-203`), and a repo-wide grep finds **zero** `oldSelf` XValidation rules.

Fields that must be immutable for correctness:

- `spec.database.database` and the `clusterRef`↔`host` mode (keystone + ControlPlane): renaming re-points db_sync at a fresh schema — Keystone silently "loses" all users/projects while the old data is orphaned. Data-loss class.
- `spec.bootstrap.adminUser`, `spec.bootstrap.region`: re-bootstrap against an existing deployment duplicates or strands catalog entries (known duplicate-admin failure class, cf. #420).
- ControlPlane `spec.openStackRelease`: the pattern admits **downgrades** (e.g. 2025.2 → 2024.1); Keystone DB migrations are not reversible. Needs a monotonic-upgrade check, which cannot exist while `oldObj` is dropped.

## Proposed fix

- Prefer CEL markers — `+kubebuilder:validation:XValidation:rule="self == oldSelf"` — on `database.database`, `bootstrap.adminUser`, `bootstrap.region` (these keep working even when the webhook is down).
- Implement release-downgrade rejection in `ValidateUpdate` (ordering logic; webhook is the right place).
- Regenerate CRDs + Helm copies; extend the invalid-CR fixture generator with update-rejection cases.

_Found by the 2026-06 pre-production audit (verified at file:line on `main` @ 0a2f19ba)._

berendt:
## Implementation Plan (issue #466)

### Summary
[...]

PR #514, opened 2026-06-19
fix(c5c3-operator): CredentialRotation reMint loop, K-ORC CEL immutability, lost-rotation race

### LINKED ISSUE #463: fix(c5c3-operator): CredentialRotation reMint loops forever; ApplicationCredential updates violate K-ORC CEL immutability; lost-rotation race
## Problems

1. **`reMint: true` re-rotates the admin credential forever.** `reconcile_credentialrotation.go:153-187` evaluates `nudge := cr.Spec.ReMint` on every reconcile with no completion latch. After a successful nudge, `reconcileKORC` re-mints and re-stamps the hash annotation — and the next reconcile of the same CredentialRotation (cache resync ~10 min via the shared `SyncPeriod` default, plus every operator restart) clears the annotation and triggers **another** delete+revoke+re-mint. A single `reMint: true` CR left in the cluster revokes the K-ORC admin application credential roughly every 10 minutes, indefinitely — each cycle opening the stale-credential auth window.
2. **Spec edits brick the KORC sub-reconciler.** The `CreateOrUpdate` mutate closure (`reconcile_korc.go:555-579`) unconditionally re-projects `ac.Spec.Resource.*` onto the existing ApplicationCredential, but K-ORC declares the whole `resource` block immutable via CEL (`self == oldSelf`). A legal, webhook-admitted change to `restricted` or `accessRules` is rejected by CEL on every pass → permanent `KORCReady=False/ApplicationCredentialError` loop; only manual AC deletion recovers.
3. **Lost-rotation race.** `reconcile_korc.go:573-576` re-stamps the hash annotation on every pass; interleaved with `reconcile_credentialrotation.go:261-269` clearing it, a concurrent nudge can be silently overwritten without a re-mint — while the rotation CR already reported `Ready=True/RotationTriggered`.

## Proposed fix

- Latch one-shot semantics: skip the nudge when `status.ObservedGeneration == metadata.generation` and the Ready reason is `RotationTriggered`/`NoRotationNeeded` (or record `status.lastTriggeredGeneration`); consider `GenerationChangedPredicate` on `For()`.
- On `Resource` drift, route through `remintAdminApplicationCredential` (delete+recreate is the only ORC-supported mutation path) instead of updating; optionally webhook-warn that these fields force a re-mint.
- Stamp the hash annotation only on create (`ac.CreationTimestamp.IsZero()`) or when absent — never overwrite an empty-string nudge marker; alternatively, CredentialRotation reports success only after observing a new AC UID.

berendt:
## Summary

Fixes the three K-ORC admin-credential rotation bugs from the 2026-06
pre-production audit (#463). All three share one root: the password-hash
annotation is used as an unguarded re-mint nudge.

Closes #463

## Problems & fixes (in commit order)

1. **`fix(c5c3-operator): latch CredentialRotation reMint to generation`**
   — `reMint: true` left in a `CredentialRotation` re-rotated the admin
   credential forever: `nudge := cr.Spec.ReMint` was evaluated with no
   completion latch, so every cache resync (~10 min via the shared
   `SyncPeriod`) and every operator restart cleared the hash annotation and
   triggered another delete+revoke+re-mint, re-opening the stale-credential
   window each cycle. An explicit `reMint` is now **latched** on a new
   `status.lastTriggeredGeneration` field — one nudge per spec edit. The
   auto-detect (password-hash change) path stays unlatched (it is
   self-limiting and relies on resync to observe an out-of-band rotation).

2. **`fix(c5c3-operator): re-mint AC on immutable resource-block drift`**
   — K-ORC declares the whole `spec.resource` block immutable via CEL
   (`self == oldSelf`, confirmed in `applicationcredential_types.go@v2.6.0`).
   The `CreateOrUpdate` mutate re-projected `ac.Spec.Resource.*`
   unconditionally, so a legal change to `restricted`/`accessRules` was
   rejected by CEL on every pass → permanent
   `KORCReady=False/ApplicationCredentialError` loop. Drift on the
   operator-managed fields (`Unrestricted`, `UserRef`, `SecretRef`,
   `AccessRules`) is now routed through the existing
   `remintAdminApplicationCredential` delete+recreate path. The comparison is
   restricted to managed fields (not whole-struct `DeepEqual`) so a
   K-ORC/CRD-defaulted sub-field can never read as permanent drift and become
   its own loop.

3. **`fix(c5c3-operator): preserve re-mint nudge marker on hash re-stamp`**
   — the mutate re-stamped the hash annotation unconditionally; interleaved
   with the rotation reconciler clearing it, a concurrent nudge could be
   silently overwritten (lost-rotation race) while the rotation CR already
   reported `RotationTriggered`. The hash is now stamped only on a fresh mint
   or when the annotation is absent, never overwriting a present-but-empty
   value (the nudge marker) — extracted as `shouldStampPasswordHash`.

4. **`docs(c5c3): document reMint latch and resource-drift re-mint`** —
   updates the ControlPlane reconciler and CRD references.

## Tests

- `TestRotation_ReMintLatchedToGeneration` — two passes, same generation:
  reMint fires once then reports `NoRotationNeeded` without re-clearing.
- `TestReconcileKORC_RestrictedDriftDeletesACForRemint` /
  `TestReconcileKORC_AccessRulesDriftDeletesACForRemint` — a matching hash
  but drifted resource block deletes the AC for re-mint.
- `TestShouldStampPasswordHash` (table) + `TestReconcileKORC_DoesNotOverwriteClearedNudgeMarker`
  (interceptor-staged race) — the empty nudge marker survives the re-stamp.
  The race test was confirmed to fail without the guard (negative control).

All new tests exercise the regression each fix targets and fail on the
pre-fix code. Existing re-mint / no-churn / `KORCReadyTrue` tests still pass.

## Verification

- `make test-operator OPERATOR=c5c3` — pass
- `make test-race OPERATOR=c5c3 RACE_FLAGS="-count=1"` — pass
- `go vet ./operators/c5c3/...` — pass
- `golangci-lint run` (c5c3 controller + api) — 0 issues
- `make generate && make manifests && make sync-crds && make verify-crd-sync` — no drift

## Notes / deviations

- **K-ORC CEL rule verified** (the plan flagged it as unverifiable from the
  sandbox): `ApplicationCredentialResourceSpec` carries
  `XValidation:rule="self == oldSelf"` at v2.6.0 — Problem 2's premise holds.
- **Nil `spec.resource` is not treated as drift.** The CEL immutability rule
  does not fire on the initial unset→set transition, so `CreateOrUpdate` is
  free to populate a nil block; treating nil as drift would wrongly re-mint
  the minimal never-minted fixture (`TestReconcileKORC_KORCReadyTrueWhenACAvailable`).
- **`GenerationChangedPredicate` deliberately not added** to `For()`: the
  reconciler relies on resync to run the auto-detect path, which that
  predicate would filter out. The generation latch alone fixes the loop.
- **Optional webhook-warn deferred** (it needs `oldObj` plumbing owned by
  #466); out of scope here.
- The `architecture/` submodule docs are in a separate repo and are not
  bumped here.

---

_Implemented by [planwerk-review](https://github.com/planwerk/planwerk-review) d0b3e45 with Claude:claude-opus-4-8_

### DIFF
--- a/docs/reference/c5c3/controlplane-crd.md
+++ b/docs/reference/c5c3/controlplane-crd.md
@@ -547,7 +547,7 @@ credential and reports progress via status conditions.
 | --- | --- | --- | --- | --- |
 | `target` | [`RotationTarget`](#rotationtarget) | Yes | — | Which credential to rotate. |
 | `bootstrap` | `bool` | No | `false` | When `true`, requests an initial **mint** of the credential rather than a rotation of an existing one. Idempotent: if the credential already exists it is a no-op. |
-| `reMint` | `bool` | No | `false` | When `true`, forces the reconciler to discard the current credential and mint a fresh one even if the existing credential is still valid. |
+| `reMint` | `bool` | No | `false` | When `true`, forces the reconciler to discard the current credential and mint a fresh one even if the existing credential is still valid. The nudge is **one-shot per spec generation** (latched on `status.lastTriggeredGeneration`), so a `reMint: true` left in the spec does not re-rotate on every resync. |
 | `intervalDays` | `*int32` | No | `nil` | **Deferred** — accepted by the schema but ignored by the L1 reconciler. Rotation cadence in days for scheduled rotation. Minimum: 1. |
 | `preRotationDays` | `*int32` | No | `nil` | **Deferred** — accepted but ignored. Days before expiry a replacement credential is minted (the overlap window). Minimum: 0. |
 | `gracePeriodDays` | `*int32` | No | `nil` | **Deferred** — accepted but ignored. Days the superseded credential remains valid after a rotation before it is revoked. Minimum: 0. |
@@ -575,6 +575,7 @@ credential and reports progress via status conditions.
 | --- | --- | --- |
 | `conditions` | `[]metav1.Condition` | Latest available observations of the rotation state. Upserted via the shared conditions helper. |
 | `observedGeneration` | `int64` | The `.metadata.generation` the controller last reconciled. |
+| `lastTriggeredGeneration` | `int64` | The most recent `.metadata.generation` for which an explicit `reMint` nudge was performed. Latches `reMint` to a single spec generation so a `reMint: true` left in the spec fires once per edit, not on every resync or restart. |
 
 ---
 
--- a/docs/reference/c5c3/controlplane-reconciler.md
+++ b/docs/reference/c5c3/controlplane-reconciler.md
@@ -635,7 +635,18 @@ that instructs K-ORC to mint the admin application credential, and drives re-min
   password-cloud — and **regenerate** the secret `value`, so the next pass recreates
   the AC for a fresh mint. The hash is computed by the package-level
   `computeAdminPasswordHash`, shared with the CredentialRotation reconciler so both
-  agree on one derivation.
+  agree on one derivation. The annotation is (re-)stamped **only on a fresh mint or
+  when it is absent**, never overwriting a present-but-empty value — that empty value
+  is the CredentialRotation reconciler's nudge marker, so preserving it keeps a
+  concurrently-cleared nudge from being silently lost (`shouldStampPasswordHash`).
+- **Re-mint on immutable resource-block drift.** K-ORC declares the AC's whole
+  `spec.resource` block immutable via CEL (`self == oldSelf`), so a legal, webhook-
+  admitted change to `restricted` or `accessRules` cannot be reconciled by an
+  in-place update — it would be rejected on every pass. `reconcileKORC` detects drift
+  on the operator-managed fields (`Unrestricted`, `UserRef`, `SecretRef`,
+  `AccessRules` — never the whole struct, so a K-ORC/CRD-defaulted sub-field can never
+  read as permanent drift) and routes it through the **same** delete+recreate re-mint
+  (`adminACResourceDrifted`).
 - **Re-mint progress / stall.** While the old AC is `Terminating` the condition is
   `KORCReady=False/ReMinting`; if it stays terminating longer than
   `remintStallTimeout` (**5m**) — a finalizer K-ORC cannot clear, e.g. it cannot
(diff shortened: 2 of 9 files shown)


berendt:
# Rebase analysis: C5C3/forge#514

> Rebased onto `main`  



## 738f1a5 fix(c5c3-operator): latch CredentialRotation reMint to generation

## 9aea721 fix(c5c3-operator): re-mint AC on immutable resource-block drift

## ed51396 fix(c5c3-operator): preserve re-mint nudge marker on hash re-stamp

## a5f28fe docs(c5c3): document reMint latch and resource-drift re-mint

## Summary

The rebase landed cleanly despite heavy overlap with upstream's c5c3-operator refactors. The most disruptive upstream commit, f8698b65 (split reconcile_korc.go into per-area files), was pure same-package code motion that explicitly retained the AC mint/re-mint state machine in reconcile_korc.go — exactly the code 9aea721e and ed51396a modify — so their insertions (adminACResourceDrifted/shouldStampPasswordHash) landed coherently alongside the upstream dedupe closure `fail` (55b422e0), terminal-error handling (71bcbf92), and clouds.yaml gating (ebfe27e7). `operators/c5c3 go build ./...` compiles with no errors, no conflict markers remain, the dropped internal/metrics package (7adf0d51) and deleted testutil/assertions+builders packages (846ceef8) are not referenced, and the admin-password digest rewire (3d3d6135, hashAdminPassword→secrets.AdminPasswordDigest) is byte-compatible with the annotation the commits compare. The new drift tests reuse a fixture set identical to the upstream-maintained TestReconcileKORC_HashMismatchDeletesACForRemint, and 738f1a55 touches reconcile_credentialrotation.go/credentialrotation_types.go which no upstream commit modified.

> [!IMPORTANT]
> **Recommendation**: Safe to push as-is. No adjustments are required from the upstream range; the production package builds and all symbols the rebased commits depend on still exist with unchanged signatures. The only residual caveat is that the test suite could not be executed in this environment (go test/go vet required approval), but symbol- and fixture-level analysis against the upstream-maintained reference tests shows no break — running `go test ./operators/c5c3/internal/controller/` once before merge would fully close the loop.

---

gndrmnn (approved):

PR #673, opened 2026-07-17
chore: add docs style guide

blut:


blut:
Replaced with https://github.com/C5C3/forge/pull/708

PR #674, opened 2026-07-17
chore: add docs skills

blut:


blut:
Replaced with https://github.com/C5C3/forge/pull/708

PR #739, opened 2026-07-24
fix: support Podman workflows and curl-based digest fallback

### LINKED ISSUE #695: make refresh-operator-digests: hard dependency on docker buildx blocks non-Docker setups. [[ Quick Start (Extended) ]]
## Problem

`make refresh-operator-digests` (which calls `hack/refresh-operator-image-digests.sh`) fails immediately on any machine without Docker Desktop installed.

The script's `preflight()` function checks for both `docker` and `docker buildx` before performing any work. If either is missing, it exits immediately:

```console
$ make refresh-operator-digests

Error: docker is required
# or
Error: docker buildx is required
```

This is a problem because the actual work performed by the script—resolving an image digest from the registry and writing a ConfigMap to the cluster—does not inherently depend on Docker.

Docker is only required for one implementation detail: resolving the digest behind the `:latest` tag.

## Affected Paths

- `make refresh-operator-digests`
- `hack/deploy-infra.sh` when running with `WITH_CONTROLPLANE=true CONTROLPLANE_OPERATORS=flux`, which invokes this script internally

## Affected Users

Any developer working with:

- Podman instead of Docker Desktop
- Lima, Rancher Desktop, or another OCI runtime
- Only `kubectl` installed (for example, when working against an existing cluster)

## Root Cause

`resolve_image_digest()` currently has a single implementation that depends on Docker Buildx:

```bash
docker buildx imagetools inspect "${image}" \
  --format '{{json .Manifest.Digest}}' 2>/dev/null | tr -d '"'
```

As a result, the script fails during preflight even though Docker is only required for digest resolution.

## Suggested Fix

Support multiple digest resolution backends, tried in the following order:

### 1. Docker Buildx (existing behaviour)

Use the current implementation when `docker buildx` is available.

```bash
docker buildx imagetools inspect "${image}" \
  --format '{{json .Manifest.Digest}}'
```

### 2. HTTP Registry API fallback

If Docker Buildx is unavailable, resolve the digest directly through the Docker Registry V2 API by:

- requesting

  ```text
  GET https://<registry>/v2/<repository>/manifests/<reference>
  ```

- handling the `401 Unauthorized` response and the `WWW-Authenticate: Bearer` challenge returned by GHCR
- requesting a bearer token
- repeating the request with the token
- reading the digest from the `Docker-Content-Digest` response header

This preserves the existing behaviour for Docker Desktop users while allowing the script to run on systems without Docker installed.

zaher-abb-vw:
`hack/refresh-operator-image-digests.sh` refused to run on any host without Docker: `preflight()` required `docker` plus a working `docker buildx` before doing any work, even though the only step that actually needs Docker is resolving the digest behind a tag. That blocked `make refresh-operator-digests` — and `hack/deploy-infra.sh` on the `WITH_CONTROLPLANE=true CONTROLPLANE_OPERATORS=flux` path, which calls it internally — for Podman, Lima, Rancher Desktop, and kubectl-only setups.

Closes #695

## What the reviewer walks through

**`hack/refresh-operator-image-digests.sh` — two digest backends instead of one**

`preflight()` now hard-requires only `kubectl`, and fails when *neither* `docker buildx` nor `curl` is present (`have_docker_buildx` / `have_curl`). `resolve_image_digest()` becomes a dispatcher: it tries `resolve_image_digest_via_docker` (the previous `docker buildx imagetools inspect` implementation, moved verbatim) and falls back to `resolve_image_digest_via_curl` — so a host that *has* Docker but cannot reach one particular image still gets the second attempt.

`resolve_image_digest_via_curl` speaks the Registry V2 API directly: it splits the ref into registry / repo / reference (a ref that already carries `@sha256:…` short-circuits, a bare name defaults to `registry-1.docker.io` + `library/`), requests `GET /v2/<repo>/manifests/<reference>` with the OCI and Docker manifest `Accept` types, and reads `Docker-Content-Digest` off a 200. On a 401 it parses the `WWW-Authenticate: Bearer` challenge, fetches a token from `realm` with `scope=repository:<repo>:pull`, and repeats the request — the ghcr.io path.

The per-image contract is unchanged: a resolve failure warns, skips that image, lets the others through, and still makes the overall exit code non-zero.

**`tests/unit/hack/refresh_operator_image_digests_test.sh` — a `curl` stub, and a fallback test**

The stub harness gains a recording `curl` alongside `docker` and `kubectl`. It is not optional: without it the new fallback fires a *real* network request at ghcr.io, which succeeds for the real published operator images and silently masks the failure scenarios these tests exercise. The stub keys on the same `DOCKER_FAIL_FOR` variable as the docker stub, so one env assignment fails *both* backends and Test 2 keeps testing "genuinely unresolvable" rather than "docker failed, curl quietly rescued it". No bearer challenge is simulated — the stub answers the first request, and fails closed if the token path is ever reached.

New Test 4 sets `DOCKER_BUILDX_FAIL=true`, asserts the digest still resolves, and asserts the curl log contains the `v2/c5c3/keystone-operator/manifests/latest` request. The later tests are renumbered.

**`docs/quick-start.md`, `docs/quick-start-extended.md`, `docs/quick-start-controlplane.md` — Podman as a kind provider**

Prerequisites now name a container runtime (Docker Desktop *or* Podman) rather than Docker Desktop, and the RAM row reads "available to the container runtime". `export KIND_EXPERIMENTAL_PROVIDER=podman` is set once before cluster creation and stays exported for the image-load step, the pull step gains a `podman pull` variant against the same provider, and the local build chain notes the `docker` → `podman` substitution. The refresh-digests note in the ControlPlane quick start records the new preference order (`docker buildx`, else `curl`).

## Test

```console
$ bash tests/unit/hack/refresh_operator_image_digests_test.sh
Results: 54 passed, 0 failed, 0 skipped
```

Hermetic — `docker`, `curl`, and `kubectl` are all stubbed, no network and no cluster.

### DIFF
--- a/docs/quick-start.md
+++ b/docs/quick-start.md
@@ -18,7 +18,10 @@ the local-build path, the production HelmRelease, E2E and Tempest, see
 
 ## Prerequisites
 
-- Docker Desktop running
+- A running container runtime:
+  - Docker Desktop, or
+  - Podman with its machine already running.
+
 - `make` on `PATH` for `install-test-deps`, `deploy-infra`, and `teardown-infra`
 - OpenStack CLI ([`python-openstackclient`](https://docs.openstack.org/python-openstackclient/latest/)) on `PATH` for the authenticated token check in Step 6
 - Pinned `kind`, `kubectl`, `Helm`, `jq` on `PATH`:
@@ -39,6 +42,13 @@ cd forge
 ```
 
 ## Step 2 — Cluster + infrastructure stack
+
+When using Podman, ensure its machine is already running and select Podman as
+kind's provider:
+
+```bash
+export KIND_EXPERIMENTAL_PROVIDER=podman
+```
 
 ```bash
 KIND_HOST_PORT=8443 make deploy-infra
@@ -81,6 +91,16 @@ kubectl wait helmrelease/keystone-operator -n keystone-system \
 ```bash
 RELEASE=2025.2
 docker pull ghcr.io/c5c3/keystone:${RELEASE}
+kind load docker-image ghcr.io/c5c3/keystone:${RELEASE} --name forge
+```
+
+With Podman, keep `KIND_EXPERIMENTAL_PROVIDER=podman` exported from the
+Prerequisites step, pull the image into Podman's image store, and load it into
+the same kind provider:
+
+```bash
+RELEASE=2025.2
+podman pull ghcr.io/c5c3/keystone:${RELEASE}
 kind load docker-image ghcr.io/c5c3/keystone:${RELEASE} --name forge
 ```
 
--- a/docs/quick-start-controlplane.md
+++ b/docs/quick-start-controlplane.md
@@ -26,6 +26,14 @@ Same toolchain as the [Quick Start](./quick-start.md), plus:
 - Roughly 8 GB RAM, 2 CPU cores, and 10 GB of free disk for a laptop-sized kind cluster
 - `yq` v4.x on `PATH` for the `KIND_HOST_PORT=8443` override path in Step 2
 
+Docker Desktop and Podman are both valid kind providers. When using Podman,
+ensure its machine is already running and select it explicitly before running
+Step 2:
+
+```bash
+export KIND_EXPERIMENTAL_PROVIDER=podman
+```
+
 - The bundled kind `ControlPlane` CR pins its backing services to a single
   instance (`spec.infrastructure.database.replicas: 1`, `cache.replicas: 1`) so
   the fresh-create chain fits a single-node kind cluster.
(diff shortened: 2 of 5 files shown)


berendt:
Resolve the merge conflicts and merge both commits into a single commit with a correct commit message (I will add a skill + docs for commit messages today).

docs/quick-start.md
```diff
@@ -18,7 +18,15 @@ the local-build path, the production HelmRelease, E2E and Tempest, see
...
 
-- Docker Desktop running
+- A running container runtime:
+  - Docker Desktop, or
+  - Podman with its machine started:
+
+    ```bash
+    podman machine start
+    export KIND_EXPERIMENTAL_PROVIDER=podman
+    ```
```
blut: Die Beschreibung wie man podman started wuerde ich hier nicht hinzfuegen.

docs/quick-start.md
```diff
@@ -84,6 +99,16 @@ docker pull ghcr.io/c5c3/keystone:${RELEASE}
...
+With Podman, pull the image into Podman's image store and keep kind on the same
+provider:
+
+```bash
+RELEASE=2025.2
+podman pull ghcr.io/c5c3/keystone:${RELEASE}
+KIND_EXPERIMENTAL_PROVIDER=podman \
+  kind load docker-image ghcr.io/c5c3/keystone:${RELEASE} --name forge
+```
+
```
blut: Ist aus meiner Sicht nicht notwendig. Alternativ alias docker=podman aufsetzen oder evtl. OCI Runtime Command offener formulieren

